PR #3, opened 2018-05-22
Apply lossless image compression to static images

acdha:
This saves ~1.4MB, roughly 30% per image on average, using jpegoptim and
zopflipng. This arguably should be automated, at least to the extent of having a Makefile or something in the img directory, since it has a substantial impact on the user's first experience.


rstorey:
This set of images is likely to be replaced or added to in the future. Would you please add links to the tools you used and instructions for the steps you followed to optimize these? I want to make sure the same steps are followed when other images are added to the project. Thanks.

Issue #9, opened 2018-05-23
Use lazy string formatting for logging


acdha:
There are a number of cases where we have logging calls which trigger immediate string formatting – for example:

https://github.com/LibraryOfCongress/concordia/blob/master/concordia/views.py#L43

Those should be changed to use [format strings but passing the arguments rather than forcing interpolation](https://docs.python.org/3/library/logging.html#logging.debug) – for example:

```python
logger.debug('Calling API path %s', abs_path)
logger.debug('Calling API path %(path)s', {'path': abs_path})
```

This avoids the performance impact of string formatting & memory allocation for messages which will never be displayed because they're below the current log output level but more importantly also provides the full context to the log handlers which allows event management tools such as Sentry to aggregate similar errors (see e.g. https://docs.sentry.io/clients/python/integrations/logging/#usage) because they can group on the format string rather than the formatted message.

PR #13, opened 2018-05-31
Refs #9, change string formatting of logger calls to be lazy

rstorey:
Refs #9

### DIFF
--- a/importer/importer/models.py
+++ b/importer/importer/models.py
@@ -15,8 +15,9 @@ sys.path.append(BASE_DIR)
 
 sys.path.append(os.path.join(BASE_DIR, 'config'))
 from config import Config
-# TODO: use util to import Config
 
+
+# TODO: use util to import Config
 
 
 class Importer:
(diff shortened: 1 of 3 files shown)


importer/importer/models.py
```diff
@@ -80,11 +81,12 @@ def check_total_item_count(self):
         actual_item_count = len(os.listdir(self.images_folder))
 
         if actual_item_count < int(self.item_count):
-            self.logger.error("Expected item count {0} but actual item count of {1} is {2}".format(
-                self.item_count,
-                self.images_folder,
-                actual_item_count
-            ))
+            self.logger.error("Expected item count %(item_count)d but actual item count of "
```
acdha: I like this in part because it'll play better with the translation framework if we ever enable localization support

importer/importer/models.py
```diff
@@ -260,11 +269,13 @@ def get_item_images(self, item_id):
...
-                self.collection_data[item_id]["size"],
-                item_id,
-                actual_item_count
-            ))
+            self.logger.error("Should have %(expected_count)d images for item %(item_id)s but "
+                              "instead have %(actual_count)d images",
+                              {'expected_count': self.collection_data[item_id]["size"],
+                               'item_id': item_id,
+                               'actual_count': actual_item_count
+                               }
```
acdha: Previously I'd ask about the linespace / indentation style consistency here but now I'm inclined to defer that to “just run Black”

provision_docker_server.sh
```diff
@@ -1,6 +1,6 @@
 sudo apt-get remove docker docker-engine docker.io
 sudo apt-get update
-sudo apt-get install \
+sudo apt-get install -y \
```
acdha: This seems unrelated to this PR but is a good change. I would consider either having these apt-get calls use `-qy` to be less noisy and maybe setting `DEBIAN_FRONTEND="noninteractive"` in the environment. If there's a chance that this will run on systems which have been upgraded from older versions of Docker it might also be a good idea to have the `install` calls use `-o Dpkg::Options::='--force-confnew'` to have APT overwrite any common config files even if there are local changes so you're always starting from the package's default config.

PR #16, opened 2018-05-31
WIP: Enable pipenv & common tools

acdha:


Pipfile
```diff
@@ -0,0 +1,39 @@
...
+verify_ssl = true
+name = "pypi"
+
+[packages]
+"psycopg2" = "*"
+"psycopg2-binary" = "*"
+gunicorn = "*"
+djangorestframework = "*"
+celery = "*"
+"7b96e2b" = {file = "file:///Users/cadams/Projects/concordia/vendor/django-registration.zip"}
```
rstorey: Can this not be an absolute path? How would this translate to another environment?

requirements.txt
```diff
@@ -1,15 +0,0 @@
...
-djangorestframework
-celery
-vendor/django-registration.zip
-Markdown
-coreapi
-confusable-homoglyphs>=3.1.1
-django-machina
-django-haystack
-Whoosh
--r requirements_importer.txt
```
rstorey: Because of incorrect architecture, the concordia app has to install the requirements for the importer app (for the time being until the architecture is fixed). I do want to keep the importer and concordia requirements separate, but, for now all the requirements have to be installed in both docker containers to get it to work (this is one of the major things on my to-do list).

requirements.txt
```diff
@@ -1,15 +0,0 @@
-Django<2.1
```
rstorey: The Dockerfiles for concordia and importer images reference these requirements.txt files. If they're going to be deleted, the Dockerfiles need to be updated. So does the README, probably, with developer environment setup instructions. We could use instruction on how to keep the imports correctly ordered going forward.

requirements.txt
acdha: We might want to see how Pipfile handles optional dependencies or whether it'd be easier just to have them installed in both places. Wheel caching has made that a lot cheaper than it used to be.

requirements.txt
acdha: I hadn't gotten to the Docker files yet – the hope was that we'd use this in both places since I really like the fixity checks on dependencies.

acdha:
Re: Docker, https://docs.pipenv.org/advanced/#deploying-system-dependencies is one way to do this if we don't want to deploy in a virtualenv but I generally like not running as root.  The alternative would be running `pipenv lock --requirements` to generate the requirements files for Docker.

acdha:
The latest version uses `pipenv` for everything. It adds a few other tools but I deferred the code cleanups for other commits to avoid scope creep.

acdha:
I wanted to discuss that more with you first — should we just merge the importer requirements into the main one or try to do something different?

rstorey:
We'll have to merge the requirements for importer into the main one for now. Hopefully in a little while we'll be able to try something else and separate them.

rstorey:
https://github.com/LibraryOfCongress/concordia/pull/97

PR #17, opened 2018-05-31
Add shebang to Docker entrypoint

acdha:
Since this will be run under Bash the shebang will tell tools like [`shellcheck`](https://www.shellcheck.net/) what linting rules to apply

### DIFF
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -1,3 +1,5 @@
+#!/bin/bash
+
 mkdir -p /app/logs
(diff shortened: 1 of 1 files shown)


acdha:
This is an excerpt of https://github.com/acdha/concordia/commit/da4ef48589a7b2ae58e90eb950b94ae6ed0c4d35

acdha:
Yeah, let me merge that from da4ef48589a7b2ae58e90eb950b94ae6ed0c4d35

Issue #19, opened 2018-06-05
Comply with LC coding standard


rstorey:
- [x] config / settings
 - [x] isort
 - [x] black
 - [x] copyright header
 - [ ] pydoc
 - [x] unit tests
 - [x] pipenv

PR #21, opened 2018-06-05
Basic coding standards

acdha:
This addresses the easiest parts of #19. Note that there are two big reformatting commits which we might want to remove from this PR since merging them will probably require all other open branches to be rebased for clarity.


rstorey:
Updated PR is here https://github.com/LibraryOfCongress/concordia/pull/98

PR #100, opened 2018-06-15
Scripts for LC jenkins setup

rstorey:


rstorey:
Our Jenkins server is in our LC-only AWS dev/test environment.

PR #102, opened 2018-06-20
Merge Artemis master fork to LOC master

acj3rd:


concordia/settings.py
```diff
@@ -262,3 +262,4 @@
     }
 ]
 
+USE_POSTGRES = True if 'postgresql_psycopg2' in Config.Get("database")["adapter"] else False
```
acdha: This ternary expression is redundant and could simply be:

 `USE_POSTGRES = 'postgresql_psycopg2' in Config.Get("database")["adapter"]`

… or if you want to emphasize the logic using parentheses:

 `USE_POSTGRES = ('postgresql_psycopg2' in Config.Get("database")["adapter"])`

Also since what the config system refers to as an adapter is what Django refers to as an engine, it would be better to use the non-deprecated `django.db.backends.postgresql` value and perhaps make this check more robust:

`USE_POSTGRES = 'postgres' in Config.Get("database")["adapter"]`

concordia/views.py
```diff
@@ -221,26 +221,23 @@ def post(self, *args, **kwargs):
...
-            c = Collection.objects.create(title=name, slug=name.replace(" ", "-"), description=name)
-            count = 0
-            for root, dirs, files in os.walk(collection_path):
-                for filename in files:
-                    filename = os.path.join(root, filename)
-                    if True:
-                        count += 1
-                        title = '{0} asset {1}'.format(name, count)
-                        media_url = os.path.join(root, filename).replace(settings.MEDIA_ROOT, '')
+                c = Collection.objects.create(title=name, slug=name.replace(" ", "-"), description=name)
```
acdha: This block of code still needs the various improvements in #5.

acdha: I distrust this kind of deeply nested unexplained code since it's not obvious whether it's correct or what the error states are. For example, the `os.system` call doesn't have an error handling path and there's no indication that this is correct.

Using `subprocess.check_call` or `subprocess.check_output` would reliably raise exceptions but it's not immediately obvious what the correct behaviour would be. In this case it's also unnecessary since it would be faster and safer to use `shutil.rmtree` which also has error handling.

acdha: Is it really the case that we want to create the collection after copying files to it? That would also allow replacing the currently duplicate code which creates a collection directory name in three places with a single method call on the `Collection` instance, which in addition to being more readable would avoid the possibility of bugs if the escaping convention every changes in the future but isn't repeated in every location.

concordia/views.py
```diff
@@ -221,26 +221,23 @@ def post(self, *args, **kwargs):
...
+                for root, dirs, files in os.walk(collection_path):
```
acdha: This is enough logic that it seems like we should have it broken into separate functions / methods so the view would clearly communicate the high-level process:

```python

c = Collection.objects.create(…)
c.copy_images_to_collection(…)
c.create_assets_from_filesystem(…)
```

acdha: One advantage to that approach would be highlighting bugs such as the race condition where two simultaneous requests would intermix their files in `/concordia_images` rather than correctly separating them.

concordia/views.py
```diff
@@ -221,26 +221,23 @@ def post(self, *args, **kwargs):
...
+                    for filename in files:
+                        file_path = os.path.join(root, filename)
+                        title = file_path.replace(collection_path + '/', '').split('/')[0]
```
acdha: We should always use `os.path.join` rather than string concatenation.

acdha: Rather than splitting on `/` we should use functions like `os.path.basename` or `os.path.dirname` which communicate the intent.

concordia/views.py
```diff
@@ -221,26 +221,23 @@ def post(self, *args, **kwargs):
...
+                        media_url = file_path.replace(settings.MEDIA_ROOT, '')
+                        sequence = filename.replace('.jpg', '')
```
acdha: Is there any chance that we will ever need to support something which is not a JPEG or which uses a different extension such as `.jpeg` or `.JPG`? Unless we're certain that this will never be true, this should use `os.path.splitext`.

concordia/views.py
```diff
@@ -221,26 +221,23 @@ def post(self, *args, **kwargs):
...
                         Asset.objects.create(title=title,
-                                             slug=title.replace(" ", "-"),
+                                             slug=title + sequence,
                                              description="{0} description".format(title),
                                              media_url=media_url,
                                              media_type='IMG',
+                                             sequence=int(sequence),
```
acdha: This should be converted at the time it's extracted from the filename and there should be a good user-visible error message instead of just bubbling a `ValueError` up. As with the rest of this code, it would also be good to know what the error handling strategy is — i.e. should this block be in a transaction?, what do we do with the files?, do we abort immediately as it currently does or just report and skip the problematic file?, etc.

concordia/views.py
```diff
@@ -221,26 +221,23 @@ def post(self, *args, **kwargs):
...
                                              collection=c)
-            # os.system('rm -rf {0}'.format('/concordia_images/*'))
-            c.is_active=1
-            c.save()
-
-            return redirect('/transcribe/' + name.replace(" ", "-"))
+                c.is_active = 1
+                c.save()
+                return redirect('/transcribe/' + name.replace(" ", "-"))
```
acdha: This should use the Django `reverse` function and, likely, escape the name unless there's another constraint which would prevent it from having non-URL-safe characters.

concordia/models.py
```diff
@@ -1,13 +1,14 @@
...
 from django.contrib.auth.models import User
+from django.conf import settings
 
 
 class UserProfile(models.Model):
     user = models.ForeignKey(User, on_delete=models.CASCADE)
     myfile = models.FileField(upload_to='profile_pics/')
 
-USE_POSTGRES = True
+USE_POSTGRES = settings.USE_POSTGRES
```
acdha: Since this is only used in one place on the very next line, I wouldn't declare a separate variable for it.

acdha:
As a general comment, there's a lot going on in this PR. I would have preferred if this were several separate pull requests — e.g. “Add Prometheus and Grafana to Docker”, “Tidy use of settings”, and whatever the intent of the view changes was — it's not obvious to me why the logic changed in a commit with a message related to PEP-8 cleanup.

That would make it easier to review the changes and it would also make it easier to do things like rebase each branch against master so e.g. the Prometheus/Grafana would would be trivial while the settings work might require actual code changes to match the new upstream design.

acj3rd:
Agree, there is a lot in this pull request. This is the first merge from the Artemis fork, and we are getting familiar with the merge process since switching from bitbucket. Future pull requests will not have as many features (jira tickets) included. At this morning's stand up, we discussed having simpler pull requests.

concordia/settings.py
rstorey: The DB engine setting has been changed upstream, so we should go with 

`USE_POSTGRES = 'postgres' in Config.Get("database")["adapter"]`

acdha: Should we keep the “adapter” term or use “engine” to match Django?

acdha:
It looks like this was merged rather than rebased so the last commit includes a bunch of previously reviewed code. I would like this to be rebased cleanly so the review only includes the actual new code which based on https://github.com/LibraryOfCongress/concordia/compare/master...ArtemisConsulting:master appears to now be only the Prometheus additions.

concordia/views.py
```diff
@@ -1,20 +1,34 @@
...
 from django.urls import reverse
-from django.views.generic import TemplateView
-from registration.backends.simple.views import RegistrationView
+from django.contrib.auth.models import User
+from django.shortcuts import get_object_or_404
+from django.db import transaction
+from django.db.models import Q
+from django.db.models import Count, Sum
+
+PROJECT_DIR = os.path.dirname(os.path.abspath(__file__))
```
rstorey: These lines should have been removed in merge. We use Django settings now (lines 25-31)

concordia/settings.py
```diff
@@ -0,0 +1,265 @@
+# TODO: use correct copyright header
```
rstorey: This file should have been deleted in merge

concordia/models.py
```diff
@@ -66,6 +68,41 @@ class Collection(models.Model):
     def __str__(self):
         return self.title
 
+    def copy_images_to_collection(self, url, collection_path):
+        result2 = None
+        try:
+            result = download_async_collection.delay(url)
+            result.ready()
+            result.get()
+            result2 = check_completeness.delay()
```
rstorey: These lines (77-79) need to be deleted since check_completeness doesn't really do anything

concordia/models.py
```diff
@@ -66,6 +68,41 @@ class Collection(models.Model):
...
+            result2.ready()
+            result2.get()
+        except Exception as e:
+            print(e)
+            pass
+
+        if result2 and not result2.state == 'PENDING':
```
rstorey: This block of code should be a separate method and it should be invoked asynchronously (through chained celery task). The create_assets_from_filesystem method would also need to be invoked asynchronously.

concordia/models.py
```diff
@@ -66,6 +68,41 @@ class Collection(models.Model):
...
+            if os.path.isdir(collection_path):
+                shutil.rmtree(collection_path)
+            shutil.copytree('/concordia_images', collection_path)
```
acdha: Should this path be hard-coded rather than a setting?

concordia/views.py
```diff
@@ -202,59 +216,22 @@ def get_template_names(self):
...
-            return redirect("/transcribe/" + name.replace(" ", "-"))
-        return render(self.request, self.template_name, {"error": "yes"})
+        name = self.request.POST.get('name')
+        url = self.request.POST.get('url')
+        slug = name.replace(' ', '-')
+        collection_path = os.path.join(settings.MEDIA_ROOT, "concordia", slug)
+        c = Collection.objects.create(title=name, slug=slug, description=name)
+        c.copy_images_to_collection(url, collection_path)
+        c.create_assets_from_filesystem(collection_path)
+        c.is_active=1
```
rstorey: Rather than setting c.is_active=1, we will need a feature for admins to "release" a collection. Not sure if that is already ticketed somewhere. No change needed here immediately, but it will need to change in the next sprint.

concordia/models.py
acdha: Is it the case that we will only ever have one pending image copy operation enforced somewhere else in the code? If not, won't this be a problem if two operations are happening simultaneously and they're both copying files to the same hard-coded path?

concordia/models.py
```diff
@@ -66,6 +68,41 @@ class Collection(models.Model):
...
+            for the_dir in os.listdir('/concordia_images'):
+                shutil.rmtree(os.path.join('/concordia_images', the_dir))
+
+    def create_assets_from_filesystem(self, collection_path):
+        for root, dirs, files in os.walk(collection_path):
+            for filename in files:
+                file_path = os.path.join(root, filename)
+                title = file_path.replace(collection_path + '/', '').split('/')[0]
```
acdha: Use `os.path.join` instead of string concatenation. 
Use `os.path.dirname` instead of splitting the strings.

Does something else in the codebase prevent directories from being nested more than one level deep?

concordia/models.py
```diff
@@ -66,6 +68,41 @@ class Collection(models.Model):
...
+                media_url = file_path.replace(settings.MEDIA_ROOT, '')
+                sequence = int(os.path.splitext(filename)[0])
```
acdha: This seems like it should have some error handling if the filename isn't numeric so there'd be more context to debug with

concordia/models.py
```diff
@@ -66,6 +68,41 @@ class Collection(models.Model):
...
+                Asset.objects.create(title=title,
```
acdha: What happens if this fails? Should this be in a transaction to rollback everything which has already been created when a failure happens?

concordia/settings.py
```diff
@@ -0,0 +1,265 @@
...
+if Config.mode == "production":
+    # TODO: production we can not have DEBUG = True
+    DEBUG = False
+    # TODO: For final deployment to production, when we are running https, uncomment this next line
+    #    CSRF_COOKIE_SECURE = True
+else:
+    DEBUG = True
+    CSRF_COOKIE_SECURE = False
+
+sys.path.append(PROJECT_DIR)
```
acdha: Manually hacking at the system path shouldn't be done at all and definitely not in the settings file.

concordia/settings.py
acdha: This seems to conflict with the settings updates which @rstorey recently made and since there's no comment on the commit I can't tell whether that's intentional

concordia/settings.py
```diff
@@ -0,0 +1,265 @@
...
+EMAIL_HOST = 'localhost'
+EMAIL_PORT = 25
+LANGUAGE_CODE = 'en-us'
+LOGIN_REDIRECT_URL = '/'
+LOGOUT_REDIRECT_URL = '/'
+ROOT_URLCONF = 'concordia.urls'
+STATIC_ROOT = 'static'
+STATIC_URL = '/static/'
+STATICFILES_DIRS = [os.path.join(PROJECT_DIR, 'static'),
+                    os.path.join('/'.join(PROJECT_DIR.split('/')[:-1]), 'concordia/static')]
```
acdha: This could be more cleanly written using `os.path.join` & `os.path.normpath`:

`os.path.normpath(os.path.join(PROJECT_DIR, "..", "concordia", "static"))`

concordia/views.py
```diff
@@ -1,20 +1,34 @@
 import csv
 import os
 import sys
+import shutil
+from django.http import HttpResponse
```
acdha: See previous about using isort & Black before submitting

concordia/views.py
```diff
@@ -1,20 +1,34 @@
...
+from django.contrib.auth.models import User
+from django.shortcuts import get_object_or_404
+from django.db import transaction
+from django.db.models import Q
+from django.db.models import Count, Sum
+
+PROJECT_DIR = os.path.dirname(os.path.abspath(__file__))
+BASE_DIR = os.path.dirname(PROJECT_DIR)
+
+sys.path.append(BASE_DIR)
```
acdha: Tinkering with the system path is usually a mistake. It is definitely a mistake in view code or to do so unconditionally.

concordia/views.py
```diff
@@ -1,20 +1,34 @@
...
+
+sys.path.append(os.path.join(BASE_DIR, 'config'))
+from config import Config
```
acdha: This looks like it hasn't been reconciled with the configuration simplification @rstorey recently made. In particular I don't see how this code could have been tested since there's no config module in the project any more.

It's also not used so it should a) be removed and b) `flake8` should always be used to clear out undeclared or unused variables before submitting code:

```
$ pipenv run flake8 concordia/views.py
concordia/views.py:1:1: F401 'csv' imported but unused
concordia/views.py:4:1: F401 'shutil' imported but unused
concordia/views.py:16:1: F811 redefinition of unused 'HttpResponse' from line 5
concordia/views.py:16:1: F401 'django.http.HttpResponse' imported but unused
concordia/views.py:19:1: F811 redefinition of unused 'User' from line 11
concordia/views.py:20:1: F811 redefinition of unused 'get_object_or_404' from line 17
concordia/views.py:22:1: F811 redefinition of unused 'Q' from line 15
concordia/views.py:23:1: F811 redefinition of unused 'Count' from line 15
concordia/views.py:23:1: F811 redefinition of unused 'Sum' from line 15
concordia/views.py:23:1: F401 'django.db.models.Sum' imported but unused
concordia/views.py:31:1: E402 module level import not at top of file
concordia/views.py:31:1: F401 'config.Config' imported but unused
concordia/views.py:33:1: E402 module level import not at top of file
concordia/views.py:34:1: E402 module level import not at top of file
concordia/views.py:36:1: E402 module level import not at top of file
concordia/views.py:36:1: F401 'importer.importer.tasks.check_completeness' imported but unused
concordia/views.py:36:1: F401 'importer.importer.tasks.download_async_collection' imported but unused
concordia/views.py:165:9: F841 local variable 'context' is assigned to but never used
concordia/views.py:221:9: F841 local variable 'context' is assigned to but never used
concordia/views.py:229:20: E225 missing whitespace around operator
```

concordia/views.py
```diff
@@ -202,59 +216,22 @@ def get_template_names(self):
 class CollectionView(TemplateView):
     template_name = "transcriptions/create.html"
 
+    @transaction.non_atomic_requests
```
acdha: This could use a comment of some sort – since the intention is to disable automatic transactions I was expecting to see transaction handling in the code but did not.

concordia/views.py
```diff
@@ -202,59 +216,22 @@ def get_template_names(self):
...
-                            media_type="IMG",
-                            collection=c,
-                        )
-            # os.system('rm -rf {0}'.format('/concordia_images/*'))
-            c.is_active = 1
-            c.save()
-
-            return redirect("/transcribe/" + name.replace(" ", "-"))
-        return render(self.request, self.template_name, {"error": "yes"})
+        name = self.request.POST.get('name')
```
acdha: Any value received from the client has to be validated

concordia/views.py
```diff
@@ -202,59 +216,22 @@ def get_template_names(self):
...
+        url = self.request.POST.get('url')
+        slug = name.replace(' ', '-')
```
acdha: Why is this not using either Django's standard `slugify` function? 

See previous comment about validation rules, also – e.g. this could be empty, it could contain nothing but spaces (are consecutive values supposed to be collapsed?), do we support unicode (e.g. `slugify(…, allow_unicode=True)`? 

Since it's being passed directly to the filesystem there's also the question of what happens if it contains a value such as a `/`, `:`, NULL, `.`, `..`, etc. If the to-be-added validation logic won't preclude those values we should probably run it through something like a URL encoder to prevent confusion or attacks.

concordia/views.py
```diff
@@ -202,59 +216,22 @@ def get_template_names(self):
...
+        collection_path = os.path.join(settings.MEDIA_ROOT, "concordia", slug)
+        c = Collection.objects.create(title=name, slug=slug, description=name)
+        c.copy_images_to_collection(url, collection_path)
+        c.create_assets_from_filesystem(collection_path)
+        c.is_active=1
+        c.save()
```
acdha: Shouldn't this call `full_clean()` first to perform validation?

Both the save & full clean calls should be in some sort of `try` block which logs errors and, since we opted out of atomic requests, presumably in a transaction block as well.

rstorey:
1. concordia/settings.py needs to be deleted (The whole file)
2. The call to check_completeness in concordia/models.py needs to be removed
3. print exception statements need to be changed to logger.error() calls
4. The USE_POSTGRES setting issue still hasn't been addressed (See #4) or just delete the setting
5. Please explain why  @transaction.non_atomic_requests is used in concordia/views.py

concordia/views.py
krishnakommineni: Regarding @transaction.non_atomic_requests:
1. This will not rollback any changes made to database, will not behave as atomic_request, which will treat each view as a transaction
2. To not to rollback everything when there is an error in processing images in task. This is for initial testing until we handle all error scenarios.

static/js/contribute.js
```diff
@@ -0,0 +1,42 @@
+//Make the DIV element draggagle:
```
acdha: This file should be run through Prettier on a regular basis

static/js/contribute.js
```diff
@@ -0,0 +1,42 @@
...
+dragElement(document.getElementById(("contribute-box")));
+
+function dragElement(elmnt) {
+  var pos1 = 0, pos2 = 0, pos3 = 0, pos4 = 0;
+  if (document.getElementById(elmnt.id + "-tabs")) {
```
acdha: Since this element is being queried repeatedly, it should be looked up once and stored so we don't need to duplicate the element ID construction string on two subsequent lines:

```
var tabElement = document.getElementById(elmnt.id + "-tabs");
if (tabElement) {
    tabElement.onmousedown …
}
```

static/js/contribute.js
```diff
@@ -0,0 +1,42 @@
...
+    /* if present, the header is where you move the DIV from: */
+    document.getElementById(elmnt.id + "-tabs").onmousedown = dragMouseDown;
```
acdha: Rather than setting `onmousedown` directly, use [`addEventListener`](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener) so this won't clobber any other code which happens to be listening for events and call [`removeEventListener`](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/removeEventListener) when necessary.

static/js/contribute.js
```diff
@@ -0,0 +1,42 @@
...
+  } else {
+    /* otherwise, move the DIV from anywhere inside the DIV: */
+    elmnt.onmousedown = dragMouseDown;
+  }
+
+  function dragMouseDown(e) {
+    e = e || window.event;
```
acdha: When do we need `window.event`? That's a non-standard Internet Explorer API which normally wouldn't be needed on any supported browser.

static/css/base.css
```diff
@@ -0,0 +1,649 @@
...
+.p-latestblog {
+  font-size: .875rem;
+}
+
+.font-weight-bolder {
+  font-weight: 600;
+}
+
+.offwhite-text {
+  color: rgba(250,250,250,1); /* #FAFAFA */
```
acdha: Rather than having some comments indicating the hex counterparts for a particular `rgb()` / `rgba()` declaration, which will have to be diligently updated, I would prefer to see symbolic names indicating why those color values exist.

I don't know what our browser support strategy is (perhaps @rstorey can help?) but while the old-school approach would just be a text comment (“primary text color”, “highlighted element title color”, etc.), a more recent approach would be to use something like SCSS with a preprocessor (which I would support), and since it's 2018 we could go straight to using native CSS variables if we don't support IE11:

https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_variables
https://caniuse.com/#feat=css-variables

(this could also be preprocessed or polyfilled)

static/css/base.css
```diff
@@ -0,0 +1,649 @@
...
+}
+
+.blue-text {
+  color: rgba(0,97,142,1); /* #00618E */
+}
+blue {
```
acdha: Can someone confirm that this selector exists because we're actually using a custom TAG named `blue` rather than a typo for a CLASS named `blue`?

concordia/views.py
acdha: As discussed in the meeting, let's create a separate issue for this independent of this PR to discuss error reporting & handling. I would vote for removing `non_atomic_requests` now so it will follow the default behaviour until that strategy is implemented.

krishnakommineni: Sure, will create a separate issue.

static/css/base.css
rstorey: I think we'll have to support whatever browsers the public loc.gov website supports. I'm not sure what their strategy is but it would be easiest to stay in sync with that.

static/css/base.css
elainekamlley: @cameroneshgh can you confirm this?

static/css/base.css
rstorey: Also, I'd like to extract this as a separate issue so it can be addressed throughout the CSS code and not hold up the pull request.

elainekamlley:
The two remaining comments will be turned into tickets

static/css/base.css
cameroneshgh: This is intentional (from the home page in the trending assets):
image

PR #107, opened 2018-06-26
Pipenv lock update

rstorey:
- Fixes the problem where Docker container has wrong version of Pipfile.lock

- Update bagIt dependency to latest version

- Change path to `vendor/django-registration` dependency to absolute Docker-compatible path (requires developers create the same path in their local devel environments)

- Removes `--deploy` from `pipenv `command in Dockerfiles (it seems to work the first time, but fail when repeated in subsequent containers).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,12 +11,13 @@ RUN apt-get update -qy && apt-get install -o Dpkg::Options::='--force-confnew' -
     locales
 
 RUN locale-gen en_US.UTF-8
+ENV LC_ALL=en_US.UTF-8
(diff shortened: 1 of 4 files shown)


acdha:
This seems reasonable but it does allow the possibility of shipping untested package combinations. The idea with the lockfile freeze in `--deploy` is that you want the container build to fail when it's outdated so you can rebuild & retest it. I'm not sure whether we're staffed at a level to handle that.

rstorey:
I agree, #108 is for testing and figuring out why it doesn't work, and adding --deploy back in

PR #109, opened 2018-06-26
Cameron chc 168 ui for transcribing

cameroneshgh:
Addressed:
CHC-214 - As a volunteer, I want to click on a thumbnail so that I can view the image
CHC-215 - As a volunteer, I want to be able to type words into a text field that corresponds to with the image
CHC-216 - As a volunteer, I want to be able to save my work so that I can submit a finished transcription

Fixed abnormal sizing of images and whitespace in collection view. Moved contribute box to right-hand side—downsized box navigation text size and fixed the color of text in the transcription and tagging boxes.
Fixed CSS and JS related to contriubute box, as well as some HTML changes to accommodate those changes—hiding status percentage when transcription not chosen on contribute box.
Fixed showing and hiding of transcription status label and percent based on which task is selected in the contribute box as well as the color of the label for the status select.
Made the decision to hide save button when viewing asset-related discussion. Also removed verbiage for status label as it's only shown when transcribing task is selected and changed the layout and spacing of the status and save row of items. Finally, also fixed the alignment of the tags input so the cursor didn't default to the bottom of the input box. Fixed a small matter of the textarea for transcription and tagging thinking it was resizable.

### DIFF
--- a/concordia/static/css/base.css
+++ b/concordia/static/css/base.css
@@ -134,8 +134,8 @@ a.gray-text:hover {
   padding-right: 0;
 }
 
-input {
-  color: rgba(51,51,51,.7); /* #333333 */
+input, select, textarea {
+  color: rgba(51,51,51,1); /* #333333 */
 }
 
 .py-default {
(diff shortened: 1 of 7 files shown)


elainekamlley:
Closing due to: Need to go through Artemis master prior to LOC

PR #139, opened 2018-08-15
Create CONTRIBUTING.md

elainekamlley:
@rstorey @acdha 

Took a stab at the contributing policy. Take a look and make any changes. I didn't add Python versions and I'm unsure about the docstring. Rosie, if those apply to our project, mind adding them? 

I'll get the PR and ISSUE templates going in the next sprint.

### DIFF
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -43,10 +43,9 @@ If you are proposing a feature:
 * Remember that this is a volunteer-driven project, and that contributions
   are welcome :)
 
-Get Started!
-------------
+### Get Started!
 
-Ready to contribute? Check out our [README](https://github.com/LibraryOfCongress/concordia/blob/master/README.rst) on how to set up your machine for local development.
+Ready to contribute? Check out our [README](https://github.com/LibraryOfCongress/concordia#concordia) on how to set up your machine for local development.
 
 1. Fork the `concordia` repo on GitHub.
 2. Clone your fork locally:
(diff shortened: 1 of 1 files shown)


docs/CONTRIBUTING.md
```diff
@@ -0,0 +1,93 @@
...
+
+If you are proposing a feature:
+
+* Explain in detail how it would work.
+* Keep the scope as narrow as possible, to make it easier to implement.
+* Remember that this is a volunteer-driven project, and that contributions
+  are welcome :)
+
+Get Started!
+------------
```
acdha: I'm tempted to run the entire thing through prettier since switching heading markup styles is slightly confusing but it also won't matter to the reader.

docs/CONTRIBUTING.md
```diff
@@ -0,0 +1,93 @@
...
+
+Ready to contribute? Check out our [README](https://github.com/LibraryOfCongress/concordia/blob/master/README.rst) on how to set up your machine for local development.
```
acdha: I don't think this will ever matter but if we used "../README.rst" it should work on branches other than master or forks, which might avoid a little confusion if someone is updating the docs.

docs/CONTRIBUTING.md
```diff
@@ -0,0 +1,93 @@
...
+
+1. Fork the `concordia` repo on GitHub.
+2. Clone your fork locally:
+
+    ```
+    $ git clone git@github.com:your_name_here/concordia.git
+    ```
+
+3. Install your local copy into a virtualenv. Assuming you have virtualenvwrapper installed, this is how you set up your fork for local development::
```
acdha: This block should be resynced with the README which reflects the switch to having `pipenv` manage the virtual environment:

https://github.com/LibraryOfCongress/concordia/blob/2b211d3054fb681edb28adfab37928ad80ff859c/README.rst#serve

I'm wondering whether this documentation should link there, also whether we should be including links to the documentation for any of the tools we're using.

docs/CONTRIBUTING.md
```diff
@@ -0,0 +1,93 @@
...
+Pull Request Guidelines
+-----------------------
+
+Before you submit a pull request, check that it meets these guidelines:
+
+1. The pull request should include tests.
+2. If the pull request adds functionality, the docs should be updated. Put
+   your new functionality into a function with a docstring, and add the
+   feature to the list in README.rst.
+3. The pull request should work for Python [add versions here]
```
acdha: Currently this is Python 3.6

rstorey: Chris is correct - Python 3.6 is the version. We don't have any coding standard around docstrings other than that they should be used.

rstorey:
@elainekamlley  I really like the Mozilla docs. Example here: https://github.com/mozilla/kitsune/ - see the link to the docs from their main README.

PR #175, opened 2018-09-17
Enable maintenance mode page

acdha:
This allows us to toggle the server into a maintenance mode
where all requests will get 503 responses using the new
maintenance mode template. 

Maintenance mode can be activated by toggling the boolean 
`MAINTENANCE_MODE` setting or by visiting 
`/maintenance-mode/on/` or `/maintenance-mode/off/` while
logged in as a superuser.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,3 +33,4 @@ mss*
 *.swp
 config-optional-override.json
 env/
+concordia/maintenance_mode_state.txt
--- a/Pipfile
+++ b/Pipfile
@@ -29,6 +29,7 @@ python-dotenv = "*"
 django-elasticsearch-dsl = "*"
 elasticsearch-dsl = "==6.1.0"
 raven = "*"
+django-maintenance-mode = "*"
 
 [dev-packages]
 invoke = "*"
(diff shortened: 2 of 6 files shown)


acdha:
@rstorey By default it creates a state file in the current directory. Does that work with our server deployment or should we set [`MAINTENANCE_MODE_STATE_FILE_PATH`](https://github.com/fabiocaccamo/django-maintenance-mode#configuration-optional) to something under `/tmp`?

rstorey:
I think ideally we'd want to point to a page on S3?

acdha:
Perhaps — I'm wondering what our threshold would be for having a switch to make this change at the CDN level, or perhaps simply just dropping the targets out of the load-balancer with a suitably nice 503 page on the CDN.

rstorey:
I added a couple to-do items to the go-live checklist about customizing the 503 page.

rstorey:
As far as the directory, I don't think it matters too much which directory the state file is in. I'm not sure whether that will actually work with our ECS deployment though - using a state file at all, that is. I will test it and find out.

PR #208, opened 2018-09-19
Enable Adobe DTM for Analytics (closes #160)

### LINKED ISSUE #160: Add Adobe Analytics Script
Acceptance Criteria: 
- [ ] Add script to the site
- [ ] Ensure that dev and production traffic are not comingled

acdha:
This embeds the code but Adobe’s instructions violate web
performance guidelines and we should review this carefully
to see how much Adobe is affecting site performance.


rstorey:
By review carefully, you mean we should evaluate the performance after I click approve and merge? :)

acdha:
The original instructions didn't mention this but we need to make this conditional based on the environment. Do you want to make it a setting or just have a if check on the CONCORDIA_ENVIRONMENT setting?

rstorey:
Let's use the CONCORDIA_ENVIRONMENT setting. I also have an ENV_NAME setting introduced in the deployment branch which is used to determine which secrets to retrieve.

elainekamlley:
I'll check in with Leah about the placement of the script. That seems odd.

acdha:
@elainekamlley apparently it's Adobe's requirement since they want to be able to run earlier. I think it'll be covered by monitoring it and we can assess strategies for mitigation if it becomes one of the primary slow points.

Issue #203, opened 2018-09-19
Admin UI improvements


acdha:
Low priority items which shouldn't be forgotten:

* [x] Every slug field should be [linked to the source field](https://docs.djangoproject.com/en/2.1/ref/contrib/admin/#django.contrib.admin.ModelAdmin.prepopulated_fields)
* [x] Set list filters on status
* [x] Enable text search
* [x] Consider enabling autocomplete for some related fields such as campaign / project
* [x] Add navigation links to related objects – e.g. jump from a project to its assets
* [x] Truncate large fields in list displays
* [x] Add publish/unpublish admin list actions
* [x] Add active/inactive admin list actions
* [x] In admin page for a user, link to their transcription activity
* [x] In admin transcriptions page be able to search on usernames
* [x] Admins are able to filter users by when their account was created and whether it is activated

elainekamlley:
Moving @lalgee comments from another ticket: 

As a community manager I need to be able to find the assets/transcripts a specific user has worked on.

- Suggested soluntion:  In admin page for a user, link to their actions (like on user profile page) AND be able to search and sort by user in the admin transcriptions page


As a community manager, I want to remove accounts that have not been activated after a certain period of time. I recieve Mailer Daemon email when users have entered their email addresses incorrectly and therefore know that those profiles will never be activated.

- Suggested solution:  Admins are able to filter users by when their account was created and if it is activated so they can be deleted.  Alternately, admins can set up automated deletion of accounts that have not been activated within a certain time window (for example 30 or 60 days)

acdha:
@elainekamlley @lalgee I'm deferring the to-do item for user/item/project counts on the transcription change list until we discuss more what would be useful to display and whether that's the right page for it.

VVH:
This all looks pretty good. Just one little thing now opened as new bug issue #691

elainekamlley:
@VVH You are da 💣 dot com (The kids still stay this right?). @acdha Good work! We can close this issue.

PR #248, opened 2018-09-20
WIP: import refactor and campaign import cleanup

acdha:
This is not yet ready to merge as there's a ton of cleanup needed on the importer still but it's at the point where you can import items into a project from the admin (on the project change form there's an “Import Items” button). There are also a bunch of admin usability improvements for #203.

Major todo items left:
* [ ] Finish implementing collection support as a group of item imports
* [ ] Finish refactoring the data model
* [ ] Finish removing spaghetti code which downloads and saves items and images

At some point we should remove the built-in S3 support and use django-storages so changing backends would be a simple settings change and we wouldn't need to have all of these branching code-paths to support.


acdha:
The spreadsheet importer is now usable, if not pretty:

screenshot 2018-09-21 13 28 50
screenshot 2018-09-21 13 29 05

This is enough to make fixing the importer more important

acdha:
@rstorey I'm inclined to close this in favor of #248 which used this as a base and is complete enough that I've been able to import the starting projects and walk through most of the UI

PR #261, opened 2018-09-24
Adding issue templates

elainekamlley:
Adding standard issue templates for 
- New features
- Report bug


elainekamlley:
@rstorey Made the sections in to questions rather than headers for clarity. Does that help split the sections?

Issue #219, opened 2018-09-20
Review for website caching


rstorey:
Check pages to see whether they will be able to be cached by cloudflare. Make changes as necessary to ensure that we can use the cache.

rstorey:
In particular, the top right nav (which changes from "Login | Register" to display a dropdown "Account" menu when someone is signed in) needs to change.

Issue #135, opened 2018-08-07
As a community manager, I want to be able to update static content that is not HTML


elainekamlley:
Update content should be user friendly and does not require coding. Either through a markdown file or another easy text format.

rstorey:
- instructions
- about
- home page text

The content can be edited as Markdown. When the site is deployed, the new markdown is incorporated into the site.

rstorey:
See #154 and #241

PR #268, opened 2018-09-25
Basic static pages serving Markdown (see #135)

acdha:
This is a simple implementation of the Markdown static pages strategy we discussed:

* The content lives in a top-level `static-pages` directory
* The `static_page` view uses the URL path component with the addition of ".md" for the filename (e.g. `/about/` becomes `static-pages/about.md`) or the URL can be registered using an extra `base_name` parameter when those values should not match
* The Markdown `title:` attribute will be used to set the page title. If not present, a converted form of the path hierarchy will be used instead

This currently requires each page to be registered in urls.py for the target location. We could handle dynamic pages by integrating with the 404 handler to check for a Markdown file before returning an error if there's a need for that.

This moves #135 into our normal development cycle. We could further improve matters by making the custom path be an S3 bucket.

This does not set set cache headers — I'm deferring that to #219.


acdha:
@rstorey I just replaced the FAQ app with a static page in this branch. @elainekamlley, @colefemeno, et al. may want to start reviewing the content which can be done at https://github.com/LibraryOfCongress/concordia/blob/d5249fbdb2f39e924368dc958023ff5e2faa43a2/static-pages/faq.md

PR #276, opened 2018-09-25
User newsletter signup

### LINKED ISSUE #155: Receive updates sign in in Registration
User stories:

- As a registered user, I want to opt in to a news letter so that I can get regular updates on crowdsourcing at the library
- As a community manager, I would like to export a list of registered users that want to sign up for our newsletter

Potential implementation:

- [x] Create a new account, tick box to get newsletter
- [x] Add a field to the database if a registered user wants to sign up for the newsletter that captures this info
- [ ] Export csv list of users who want to sign up by date
- [x] Use the django admin page for exporting

rstorey:
When a user ticks the newsletter sign-up box on registration, they are added to the Newsletter group.

This closes #155 except there is no special CSV export in the admin view. What is already built in is filtering based on Newsletter group, and then copy and paste. I think the admin export should be a separate ticket.


acdha:
For admin export, we should test whether this requires changes for Django 2.x since it could be a simple option:

https://github.com/LibraryOfCongress/django-tabular-export#admin-integration

concordia/forms.py
```diff
@@ -56,11 +68,15 @@ def save(self, commit=True):
             setattr(instance, role_dict[role], 1)
         if commit:
             instance.save()
+
+        if "newsletterOptIn" in self.data and self.data["newsletterOptIn"]:
+            # Add this user to the Newsletter group if they opted in.
+            post_save.connect(add_user_to_newsletter_group, sender=User)
```
acdha: I'm a little confused by the ordering here — since we're just calling save on the line before, why not just add the user directly?

Also, do we have a process for opt-ing out?

Makefile
```diff
@@ -24,7 +24,7 @@ firstup: increase-elk-max-map-count create-docker-sentry-network
 
 .PHONY: adminuser
 adminuser:
-	docker-compose -f docker-compose.yml run --rm app ./manage.py shell -c "from django.contrib.auth.models import User;from django.contrib.auth.models import Group; User.objects.create_superuser('admin', 'crowd@loc.gov', '${CONCORDIA_ADMIN_PW}');Group.objects.create(name='CM')"
+	docker-compose -f docker-compose.yml run --rm app ./manage.py shell -c "from django.contrib.auth.models import User; User.objects.create_superuser('admin', 'crowd@loc.gov', '${CONCORDIA_ADMIN_PW}')"
```
acdha: This looks functional as long as the password doesn't have shell characters in it. The alternative would be `manage.py createsuperuser --no-input --username=admin --email=crowd@loc.gov` and forcing people to use the password-reset process on the initial install.

concordia/forms.py
rstorey: Yeah, I thought that would work too. I guess there's a bit of weirdness with how Django clears and refreshes many-to-many relationships. At this point the id for newly-created Users doesn't exist, although it seems like it should.
GovDelivery manages the newsletter - I think the opt-out thing is handled through that.

Makefile
rstorey: The main point of this change was to get rid of creating the special "CM" group here (for Community Managers) and put it in a migration. The only reason I haven't changed the process to what you described is because I don't have access to crowd@loc.gov.

rstorey:
Created #278 for the admin export thing

Makefile
acdha: Okay, works for me.

rstorey:
Ok to merge then? Did you want to look more into that post-save thing?

rstorey:
Still to do:

- [ ] Confirm that the signal does not persist for the life of the process so this would mean that all future users would be opted-in as well.

- [ ] Switch django-registration to the release package instead of vendor/

- [ ] Looking at https://django-registration.readthedocs.io/en/3.0/signals.html#django_registration.signals.user_registered makes me wonder whether we should hook on that instead since it looks like we'd the HttpRequest with the form fields and the actually-created user in scope.

concordia/forms.py
VVH: Users have to explicitly opt-in to the newsletter, as per wider Library policy. They can opt out through gov delivery emails or going to the gov delivery site.

Pipfile
```diff
@@ -20,7 +20,7 @@ Pillow = "*"
 confusable_homoglyphs = ">=3.1.1"
 Whoosh = "*"
 bagit = "*"
-"5c736a0" = {path = "./vendor/django-registration.zip"}
+django-registration = "*"
```
acdha: 🎉

PR #277, opened 2018-09-25
Refs #251, remove HTML from account confirm email

### LINKED ISSUE #251: Modify account confirmation email
Steps to reproduce:
1. Visit the site as an unauthenticated user.
2. Complete the registration form and use a valid email address.
3. Wait to receive the email and read it.

Observed behavior:
1. There is HTML source code in the email text.
2. Cannot click any button.

Expected behavior:
1. The account activation email should be plain-text formatted
2. The account activation email should have a link in plain text

rstorey:
Closes #251

### DIFF
--- a/concordia/templates/registration/activation_email.txt
+++ b/concordia/templates/registration/activation_email.txt
@@ -4,7 +4,7 @@ Hi {{ user }},
 Thank you for joining Make Your Mark! Please verify your email address in the next {{ expiration_days |multiply:24 }} hours.
 
 To complete activation, please click the following link:
-<button class=button><a href='http://{{ site }}{% url "registration_activate" activation_key %}'>Verify Now</a></button>
+http://{{ site }}{% url "registration_activate" activation_key %}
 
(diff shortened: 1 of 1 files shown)


concordia/templates/registration/activation_email.txt
```diff
@@ -4,7 +4,7 @@ Hi {{ user }},
 Thank you for joining Make Your Mark! Please verify your email address in the next {{ expiration_days |multiply:24 }} hours.
 
 To complete activation, please click the following link:
-<button class=button><a href='http://{{ site }}{% url "registration_activate" activation_key %}'>Verify Now</a></button>
+http://{{ site }}{% url "registration_activate" activation_key %}
```
acdha: This should be `https` for production. I'm inclined to say we do that now and either get ACM setup on our load-balancers or get some LC internal certs for every other tier.

rstorey: Changed to https, although the load balancers still need to catch up

Issue #270, opened 2018-09-25
Finalize content for soft launch


elainekamlley:
Refine content and scope down:

- [x] homepage
- [x] about
- [x] faq
- [x] campaigns descriptions 
- [x] quick tips
- [x] notifications/messages
- [x] register for account
- [x] user account page
- [x] contact us
- [x] latest (to include letters to lincoln challenge)
- [x] For educators page

elainekamlley:
@VVH working on a updated spreadsheet based on #346

elainekamlley:
By weds update content for comms meeting

VVH:
Pages created in Confluence/Jira with drafts of Campaign Descriptions (including links etc) and 'Latest' page. New Quick Tips content generated as well.

elainekamlley:
This is done

Issue #154, opened 2018-09-13
Create static pages


elainekamlley:
Acceptance criteria: 

Create view and markdown of 
- [x] FAQ
- [x] Instructions
- [x] Latest
- [x] For educators

elainekamlley:
TODO: 
Convert to markdown
Top level directory - URLS

PR #281, opened 2018-09-26
WIP: Content updates 

elainekamlley:
Addresses parts of #154 and #270 

- [ ] about/faq
- [ ] homepage

### DIFF
--- a/static-pages/about.md
+++ b/static-pages/about.md
@@ -1,16 +1,57 @@
-title: Empowering Readership and Transcription for the “We” Generation
+title: About Make Your Mark
 
-We, the people, have a vision for a Library transcription and tagging tool to empower you, the user, to interact with historic materials in new ways. While typeset text can be scanned by optical character recognition software (OCR), to quickly and inexpensively add full text search capability, there’s no simple way to do that for hand-written diaries, journals, manuscripts, un-tagged photographs of this kind, or other primary source material of that ilk.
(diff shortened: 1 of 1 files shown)


elainekamlley:
I would like to make more updates to the home page but didn't feel confident doing them without previewing it on my local host. Need to set that up.

acdha:
Are we certain that we want to delete the FAQ page? If so, we'll want to update the site navigation templates as well.

acdha:
I just pushed a commit with some minor Markdown tidying using [Prettier](https://prettier.io/)

acdha:
Where do we want the instructions page to live? Is that going to be a separate page, loaded inline on the transcription page, or both?

elainekamlley:
@acdha yes, I think we want to try and merge About and FAQ together. There is another ticket to confirm the top level nav and footer to reflect pages we want. 

As for the instructions page, its going to be a separate page for now and we will do some user testing on how we can refine how users will get to them, use them, etc. I could use some help in creating the view and path.

acdha:
Basically adding it will be the reverse of what I just did to remove the FAQ link. In the main `urlpatterns` block of `concordia/urls.py` we just add a path matching the hierarchy in the `static-pages` folder. For example, to match where it is currently that could be as simple as this:

```python
path("instructions/", views.static_page, name="instructions")
```

We could, of course, move that under a sub-directory if you wanted to have the URL be something like `/about/instructions/`, etc. and there's the option of overriding it completely if we need something truly custom.

PR #290, opened 2018-09-27
Update services tasks

rstorey:
Reduce CPU requirements in task definitions so ECS doesn't have such a hard time placing tasks on instances.


concordia/migrations/0026_creategroups.py
```diff
@@ -0,0 +1,28 @@
...
+from django.conf import settings
+
+
+def create_groups(apps, schema_editor):
+    Group = apps.get_model("auth", "Group")
+
+    cm_group_exists = False
+    newsletter_group_exists = False
+
+    try:
```
acdha: I don't really care that much but this could be `Group.objects.get_or_create()` without the `try`/`except` block

concordia/settings_template.py
```diff
@@ -299,3 +299,8 @@
 # When the MAINTENANCE_MODE setting is true, this template will be used to
 # generate a 503 response:
 MAINTENANCE_MODE_TEMPLATE = "maintenance-mode.html"
+
+# Names of special django.auth Groups
+
+COMMUNITY_MANAGER_GROUP_NAME = "CM"
```
acdha: Should this be “Community Managers”?

rstorey:
Please recheck @acdha

rstorey:
I'm not sure either. I need to carefully set up a  new stack somewhere and test this (apart from our dev / test / security assessment instances).

Issue #159, opened 2018-09-13
Add Tag RegEx and rules


elainekamlley:
**Acceptance Criteria:** 

- Add these following rules to tags:

- [ ] Max char length: 50

 **Allowed**
- [ ] - dash
- [ ] - spaces
- [ ] - underscore
- [ ] - apostrophe
- [ ] - letters
- [ ] - numbers

**Not allow**
- [ ] - No special char
- [ ] - URLs
- [ ] - naughty words
- [ ] - This list can be edited by community manager

** How to close a tag:**
- [ ] - Enter
- [ ] - semicolon
- [ ] - colon
- [ ] - comma
- [ ] - tab

acdha:
Are we defining letters in the sense that Unicode does or ASCII? I’m assuming the former since there are many potential terms with accents.

acdha:
@colefemeno Should we have a minimum length – maybe 1 or 2 characters?

acdha:
#304 implements most of this. Were the unchecked values in “how to close a tag” supposed to be applied? I can add those as well.

acdha:
@elainekamlley Is there any reason to keep this open?

PR #296, opened 2018-09-27
Asset tagging UI

acdha:
This implements a basic UI for tagging assets

### DIFF
--- a/concordia/static/css/base.css
+++ b/concordia/static/css/base.css
@@ -1375,10 +1375,6 @@ div.nav-secondary {
     border: 0px solid transparent;
 }
 
-#test-percent {
-    height: 0;
-}
-
 .contribute-box {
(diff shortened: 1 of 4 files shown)


acdha:
Here's an example of the basic functionality in action:



It prevents duplicates on both sides but I have not yet implemented typeahead suggestions from other people's tags.

Two issues which @elainekamlley or @colefemeno might have answers to:

1. Do we allow spaces in tags? If so, I should remove that code. It's a little less common than traditional tags but would allow library-style “Lincoln, Abraham” type tags.

2. Do we have any thought about case-sensitivity? I made the duplicate check case sensitive but not the database lookups.

elainekamlley:
@acdha There is a RegEx for tags ticket #159. We allow spaces but have not made rules about case sensitivity. My instinct is to not be case sensitive, what do you think @colefemeno

PR #306, opened 2018-09-28
Update account profile page

acdha:
The user profile page now allows users to change their email address and has a nicer display of their contribution history.

This also removes a never-used user profile model field and redundant form/view code


acdha:
Here's what the UI looks like currently:



When someone attempts to use an email address which conflicts with another user it looks like this:

image

acdha:
@rstorey Don't merge this quite yet – I have a couple of fixes for the password change pages, too

acdha:
Okay, I just pushed the updates for the password change templates so they follow the site look and feel

PR #308, opened 2018-09-28
Populate the campaign list on the home page from the DB

### LINKED ISSUE #291: Link campaigns to project pages on the homepage
In the campaign section: 

- [x] Remove the get started button
- [x] link image to go to the projects page for that campaign

rstorey:
Closes #291 

Note that the original design of the home page is meant for exactly six (or some other multiple of three) campaigns to be displayed. I have adjusted that in this PR so that it now will display 4. I understand for soft launch that number will be 2 and that number may be changing. We need to decide how we want to handle that on the home page.


concordia/views.py
```diff
@@ -145,6 +145,19 @@ def form_valid(self, form):
         return super().form_valid(form)
 
 
+class HomeView(TemplateView):
```
acdha: It's kind of moot for this but we could change this to be a ListView

concordia/views.py
```diff
@@ -145,6 +145,19 @@ def form_valid(self, form):
...
+    template_name = "home.html"
+
+    def get_context_data(self, *kws):
+        queryset = Campaign.objects.published().order_by("title")[:4]
```
acdha: Should we hard-code the `[:4]` in here or let the template do the slicing so we don't need to touch the view if it changes?

concordia/templates/home.html
```diff
@@ -51,49 +51,48 @@ <h4 class="offwhite-text mt-half">REVIEW</h4>
...
-                <h3 class="mt-half">Abraham Lincoln Papers</h3>
-                <a class="btn btn-outline-primary" href="#" role="button">Get Started!</a>
-            </div>
-            <div class="col-12 col-md-4 mb-default text-center border-left-home-contribute">
-                <img class="img-home-contribute d-block mx-auto img-fluid" src="{% static 'img/MaryChurchTerrell.svg' %}"
-                    alt="Mary Church Terrell Papers">
-                <h3 class="mt-half">Mary Church Terrell Papers</h3>
-                <a class="btn btn-outline-primary" href="#" role="button">Get Started!</a>
+            {% comment %}
+            FIXME: col-md-6 when there are 4 campaigns to display, col-md-4 when there are 6
```
acdha: I realize you inherited this but it really looks to me like we could simplify all of this to a simply using flexbox and not trying to abuse the column layout classes to set item widths

concordia/templates/home.html
```diff
@@ -51,49 +51,48 @@ <h4 class="offwhite-text mt-half">REVIEW</h4>
...
+                <a class="btn btn-primary" href="{% url 'transcriptions:campaigns' %}" role="button">Browse All Campaigns</a>
             </div>
         </div>
-        <div class="col mx-auto d-flex justify-content-center my-default">
-            <a class="btn btn-primary" href="{% url 'transcriptions:campaigns' %}" role="button">Browse More Campaigns</a>
-        </div>
     </div>
-</div>
-{% endblock main_content %}
+    {% endblock main_content %}
```
acdha: whitespace

concordia/templates/home.html
```diff
@@ -51,49 +51,48 @@ <h4 class="offwhite-text mt-half">REVIEW</h4>
         <div class="col-12 mb-half">
             <h2 class="px-default px-md-0 ml-md-default">Campaigns: <small>Choose what to transcribe</small></h2>
         </div>
+
+
```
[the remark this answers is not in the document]
rstorey: Ha! yeah.

concordia/views.py
rstorey: I didn't know the template could do that. It would be preferable to put it all in the template.

concordia/templates/home.html
rstorey: Is there a VS Code / prettier setting for django template language? I noticed your indentations are different than mine.

rstorey: Prettier is automatically indenting that when I save the file. :(

acdha: I've been using https://github.com/vscode-django/vscode-django but that doesn't format at all. I've been okay with that since many things don't handle the indentation of HTML + Django template tags well.

Re: the indentation, there was a missing close tag which is why Prettier was adding the extra indent

concordia/templates/home.html
acdha: https://github.com/LibraryOfCongress/concordia/pull/310 has my experiment in this direction

concordia/templates/home.html
```diff
@@ -51,49 +51,48 @@ <h4 class="offwhite-text mt-half">REVIEW</h4>
...
-            <div class="col-12 col-md-4 mb-default text-center border-left-home-contribute">
-                <img class="img-home-contribute d-block mx-auto img-fluid" src="{% static 'img/WilliamOlandBourne.svg' %}"
-                    alt="William Oland Bourne Papers">
-                <h3 class="mt-half">William Oland Bourne Papers</h3>
-                <a class="btn btn-outline-primary" href="#" role="button">Get Started!</a>
+            <div class="col mx-auto d-flex justify-content-center my-default">
+                <a class="btn btn-primary" href="{% url 'transcriptions:campaigns' %}" role="button">Browse All Campaigns</a>
             </div>
         </div>
-        <div class="col mx-auto d-flex justify-content-center my-default">
```
acdha: I think this should be reverted – it causes the browse all button to go up to the collections row on wide screens

concordia/views.py
acdha: Hmmm, yes. Is this a requirement that we always break it into two even rows rather than letting the browser wrap it?

https://getbootstrap.com/docs/4.1/layout/grid/#equal-width-multi-row suggests that we can do this with a simple `{% if forloop.count == column_count %}<div class="w-100"></div>{% endif %}`  which would cut down on the template complexity but doesn't help with the need to calculate that value. We do have that `custom_math` template tag library so we could simply put in `campaigns|length|multiply_by:0.5`.

rstorey: I think it's fine if they're not evenly distributed - if the browser decides. Your implementation is much cleaner and this way we don't have to modify anything except template code if we decide to change the number of campaigns we want to list on the home page. I changed it back to 6 since the flex implementation works for anything up to 6 items. So we don't need to change anything in the code unless we decide we want to show more than 6.

acdha: It's also super-easy to do consistent columns using CSS Grid:

```diff
diff --git a/concordia/static/css/base.css b/concordia/static/css/base.css
index 950e6fd..8e71402 100644
--- a/concordia/static/css/base.css
+++ b/concordia/static/css/base.css
@@ -1898,3 +1898,9 @@ div.nav-secondary {
 .image-icon.itunes {
     background-position: -114px 0;
 }
+
+.section-homepage #campaign-list {
+    display: grid;
+    grid-template-columns: repeat(2, 1fr);
+    grid-gap: 0.6rem;
+}
diff --git a/concordia/templates/home.html b/concordia/templates/home.html
index fdf5e06..dd7b331 100644
--- a/concordia/templates/home.html
+++ b/concordia/templates/home.html
@@ -47,12 +47,12 @@
             </div>
         </div>
     </div>
-    <div class="row my-5">
+    <div class="row my-5 justify-content-center">
         <div class="col-12 mb-half">
             <h2 class="px-default px-md-0 ml-md-default">Campaigns: <small>Choose what to transcribe</small></h2>
         </div>
 
-        <div class="d-flex flex-wrap col-12 justify-content-around">
+        <div id="campaign-list" class="mx-auto">
             {% for campaign in campaigns %}
                 <div class="text-center campaign m-1">
                     <a href="{% url 'transcriptions:campaign' campaign.slug %}">
```

Maybe what we should do is add the grid bit above so we get a nice 2 column layout on modern browsers and IE11 falls back to flexbox?

rstorey: I like having 3 columns and 2 rows when there are six items - this would make it always 2 columns? I don't think I want that, but it's really up to @elainekamlley and @colefemeno

acdha: That would always set two columns. We could either make it responsive (i.e. set a threshold for going from 1 to 2 to 3 columns) or do the calculation based on the current number of campaigns.

concordia/templates/home.html
rstorey: I'm not able to reproduce that. How wide of a screen do you have?

acdha: This was fixed in https://github.com/LibraryOfCongress/concordia/commit/437d83748aa45a54564699adcbbfa9bc72d42d83 (and possibly other commits) by moving the browse all button outside of the same flex container as the campaigns.

PR #309, opened 2018-09-28
Overhaul Contact Us form

acdha:
* [x] Use type-specific Django form fields
* [x] Use bootstrap4 instead of hand-coding everything
* [x] Always include the HTTP Referer distinct from the optional  link field
* [x] Use Django’s form API to validate inputs
* [x] Use Django’s form API to provide the template context  rather than building it by hand
* [x] Don’t HTML escape the plain-text content
* [x] Add HTML alternative message
* [x] Handle errors rather than ignoring them

### DIFF
--- a/concordia/forms.py
+++ b/concordia/forms.py
@@ -39,17 +39,10 @@ class UserProfileForm(forms.Form):
 
 
 class ContactUsForm(forms.Form):
-    email = forms.CharField(
-        label="Your email",
-        required=True,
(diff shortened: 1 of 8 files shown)


acdha:
@elainekamlley Feel free to suggest any changes to the email message format while we're working on this. 

One questions I had: subject is optional, but we include it in the email. I changed it so we'll never send an email with a blank subject but I'm not sure whether it makes more sense to simply remove the subject field, make it required, or add a default.

PR #317, opened 2018-10-01
Travis integration

rstorey:
https://travis-ci.org/LibraryOfCongress/concordia


.travis.yml
```diff
@@ -0,0 +1,17 @@
+language: python
+env:
+  - DJANGO_SETTINGS_MODULE=concordia.settings_test
+python:
+  - "3.6"
+  - "3.6-dev"  # 3.6 development branch
+  - "3.7-dev"  # 3.7 development branch
```
acdha: I would include 3.7 release as well as the dev branch. We may need to add something like `allowed_failures` for the dev branches, too

.travis.yml
```diff
@@ -0,0 +1,17 @@
...
+# command to install dependencies
+install:
+  - pip install pipenv
+  - pipenv install --dev --deploy
+# command to run tests
+script:
+  - mkdir logs
+  - touch ./logs/concordia-celery.log
```
acdha: Based on the failures, we probably need to add `./manage.py collectstatic --noinput` here, too

acdha: https://github.com/LibraryOfCongress/concordia/pull/317/commits/e4529233a5f08d401bf24017d1fc019da8a11845

.travis.yml
rstorey: See https://github.com/travis-ci/travis-ci/issues/9815

.travis.yml
rstorey: I tested with the mkdir and touch commands removed (after collectstatic was added) and I had to add them back in.

rstorey:
The tests are still erroring / failing, but I did get the 3.7 python release added.

.travis.yml
```diff
@@ -1,11 +1,14 @@
...
-  - "3.7"
-  - "3.7-dev"  # 3.7 development branch
+matrix:
+  include:
+    - python: "3.6"
+    - python: "3.6-dev"
+    - python: "3.7-dev"
+    - python: "3.7"
+      dist: xenial
+      sudo: true
```
michael-k: You need the same config for 3.7-dev, otherwise 3-7-dev is just an old beta of 3.7.0 and not a pre release of the next minor version.

acdha: Oh, that's quite a footnote: https://docs.travis-ci.com/user/languages/python/#development-releases-support

I'm guessing that means we can go back to a simple `python:` block because we should use xenial everywhere.

rstorey:
I think this is ready now. @acdha

PR #327, opened 2018-10-02
Model relationship cleanup

acdha:
This is the big cleanup branch:

* No more unvalidated `user_id` fields
* `Transcription`'s unused`parent` field is removed
* The `Campaign` -> `Project` -> `Item` -> `Asset` hierarchy is now enforced. We will want to add `select_related` / `prefetch_related` calls as appropriate but we're no longer hitting so many points of tedium due to crossing the hierarchy
* `Item`s no longer have slugs; the `item_id` field is used instead
* `s3_storage` is gone, along with the last of the branchy code checking it


acdha:
@rstorey This works to the point where I imported items and transcribed them but I'm working on getting the entire test suite to pass

acdha:
@rstorey One other big one: `Asset.media_url` is misleading since it's really just the filename. I'm inclined to rename that to indicate that it's just the filename and that the filenames should all be `<sequence>.ext`

docker-compose.yml
```diff
@@ -21,7 +21,7 @@ services:
         environment:
             POSTGRES_PASSWORD: ${POSTGRESQL_PW}
             POSTGRES_USER: concordia
-            POSTGRES_MULTIPLE_DATABASES: concordia
+            POSTGRES_MULTIPLE_DATABASES: test_concordia
```
rstorey: I think this would have to be `concordia,test_concordia` in order for the various settings options to work with docker? e.g. the `concordia.settings_dev` and `concordia.settings_prod` mode (which is used in docker) have the db name as `concordia`

Nevermind - I tested and it works fine

acdha: That's what I thought, too, but if you list the primary database in there the create command will crash due to it already existing

PR #328, opened 2018-10-02
Use IAM role with ECS tasks

rstorey:
This update creates a Task Role and attaches it to the ECS tasks. However, the application and importer still can't retrieve secrets from the secrets manager even with this role added to the task. Does it also need to be a "taskExecutionRole"  in addition to "taskRole" ?


rstorey:
This is ready now

PR #329, opened 2018-10-03
Update docs

elainekamlley:
Proposed changes: 

- README: Added an intro, moved code quality to how we work
- update how we work


elainekamlley:
@rstorey completely remove it or replace it with other instructions?

rstorey:
I guess we should replace it with other instructions... let me come up with something.

Issue #173, opened 2018-09-14
Flatten all migrations before launch


elainekamlley:
https://docs.djangoproject.com/en/2.0/topics/migrations/

Before we deploy to production, we'll want to take all our django migrations and flatten them into one initialization migration for the database.

PR #331, opened 2018-10-03
Squash 0001 through 0040 migrations

rstorey:
Refs #173


rstorey:
I was thinking we could delete those next week, after this has a chance to get deployed wherever / after I have a chance to reset the dev / test DBs.

PR #333, opened 2018-10-04
Flatten migrations

rstorey:
Deployment fixes including squashed migrations for importer.


importer/migrations/0001_squashed_0015_auto_20180925_1851.py
```diff
@@ -0,0 +1,328 @@
...
+        ("importer", "0012_auto_20180923_0231"),
+        ("importer", "0013_auto_20180924_1318"),
+        ("importer", "0014_auto_20180924_1943"),
+        ("importer", "0015_auto_20180925_1851"),
+    ]
+
+    initial = True
+
+    dependencies = [
+        ("concordia", "0019_auto_20180920_1503"),
```
acdha: Should we change the three concordia dependencies to the squashed one now?

rstorey: Django recognizes that the squashed concordia dependency replaces any references to the dependencies it squashes. Which is why I had to remove the dependency in one of the non-squashed importer migrations - it errored out because it had been applied prior to the concordia squashed migration. So I think the answer is, we don't have to. I did deploy this  branch to AWS and tested it there with DBs that were pre-existing. I also tested it locally after wiping out my docker DB volume.

acdha: Okay, we'll need to update this when we remove the old migrations but we can do that then since we'll need to update the concordia migrations, too.

PR #334, opened 2018-10-04
Update README to remove obsolete .env file stuff

rstorey:
Now that I look at it, I don't think we need any info about passwords in the readme. They're not needed to run a development environment. If there are specific questions you think need to be answered, let me know and I'll write something.


rstorey:
Yep, sentry instructions were removed too.

PR #335, opened 2018-10-04
WIP: Add setuptools_scm to make application version available in Django

rstorey:
#322


setup.py
```diff
@@ -26,4 +26,9 @@
     scripts=SCRIPTS,
     install_requires=INSTALL_REQUIREMENTS,
     classifiers=CLASSIFIERS,
+    use_scm_version={
+        "write_to": "version.txt",
+        "tag_regex": r"^(?P<prefix>v)?(?P<version>[^\+]+)(?P<suffix>.*)?$",
```
acdha: Is this just to avoid needing to prefix version tags with “v”?

concordia/views.py
```diff
@@ -67,6 +55,8 @@ def healthz(request):
     # at last once:
     status["database_has_data"] = Campaign.objects.count() > 0
 
+    status["application_version"] = get_version()
```
acdha: I'm wondering whether there's enough of a performance impact here that we should toss `get_version()` into a function decorated with `functools.lru_cache` – is there a chance that this fails back to querying Git or does it just ask pkg resources?

setup.py
rstorey: I think it would be useful, but I'm still puzzling through how we're going to get that version number into the docker tag. We currently don't have `setup.py build` as part of our deployment process. If we did, the version.txt file would be created. I'm thinking of adding that, but that introduces some complexities for the build server.

concordia/version.py
```diff
@@ -0,0 +1,7 @@
+from setuptools_scm import get_version
+import functools
+
+
+@functools.lru_cache(maxsize=None)
```
acdha: I'm not sure maxsize matters since we have no arguments but it also doesn't matter

Issue #325, opened 2018-10-02
Refactor page-in-use implementation


acdha:
The page in use display on the asset detail page needs to be reviewed and brought into line with the UI style

PR #338, opened 2018-10-04
Finish fixing tests

acdha:
This reaches the point where the tests pass locally on my system. The page-in-use tests are partially fixed but since the API is poorly designed I chose to mark those tests as expected failures since all of the view code will change in #325. This PR does include a migration to fix the model design.


acdha:
That was my main goal: now we can treat Travis breaking as a problem rather than the natural order of things

PR #344, opened 2018-10-04
WIP: Refactor page in use

### LINKED ISSUE #325: Refactor page-in-use implementation

acdha:
This still needs tests but it's considerably faster than the old approach and safer. We can decide how much to work with @jbresner on the conflict UI in this ticket.

Closes #325


rstorey:
This is looking good so far. Personally I think we should remove the "You have exclusive permission" message that comes up when someone loads the page... is it actually meant to exclude everyone else from being able to enter transcriptions? I don't think we should do it that way, since there are many reasons why someone might want to load the page other than to actively work on a transcription. I like the message that comes up when you're the second user that says "Someone else is transcribing this page." That part is exactly what we want. The only other thing we might want to add there is a link to some other page (or maybe prompt users to click the Next Page button).

acdha:
This now has 100% test coverage, including branches, and I rewrote the view to use direct SQL to save a couple of queries per call. The main concern I have now is that the query counts vary depending on the `SESSION_BACKEND` setting, which is fine in the sense that I want to change that to use the cache anyway but we should setup Redis/Memcache as the cache backend first to avoid problems once the number of running instances goes from 1 to many.

Regarding the UI, I'm wondering whether the best answer here is to merge what we have and then iterate again.

elainekamlley:
@acdha Can you add a screenshot of the UI?

rstorey:
You can see the changes on the crowd-dev site now. @elainekamlley

PR #347, opened 2018-10-05
Update report view

acdha:
This page isn't linked anywhere so I missed it the first time in #327. I also moved it to a consistent URL hierarchy.

I was thinking about adding a link somewhere — is there any reason why this shouldn't be a public link? (cc: @colefemeno @elainekamlley)


elainekamlley:
@acdha what kind of report is this?

acdha:
It's the campaign-level report which lists the projects and their asset statuses — here's a screenshot:

image

PR #352, opened 2018-10-05
updates to CF templates

rstorey:


cloudformation/add_cloudflare_ips_to_sgs.py
```diff
@@ -0,0 +1,36 @@
+#!/usr/bin/env python3
+
+# TODO: move this into Ansible or at least use Boto
```
acdha: I'm okay checking this in but sorely tempted to rewrite it to use boto

rstorey: Done

rstorey:
I'd like to merge this soon-ish, since new deployments (in the new AWS account) will need this stuff.

cloudformation/add_cloudflare_ips_to_sgs.py
```diff
@@ -1,10 +1,12 @@
...
 
-# TODO: move this into Ansible or at least use Boto
-
-import json
-import subprocess
+import boto3
 import requests
+from botocore.exceptions import ClientError
+
+SECURITY_GROUP_ID = "sg-0e07161e54ca34212"
```
acdha: We could probably just query the security group list for a tag like `CloudFlareIngress` and iterate over the results, which would come in handy if e.g. we wanted to have staging match production in this aspect, too.

rstorey: Oh I like that idea. I'm going to ticket that.

Issue #222, opened 2018-09-20
Status fields need on-save trigger


rstorey:
Completion status rolls up to the campaign level from the transcription / asset level.

There should be an on-save trigger that checks whether the last item in a Project is done, and if so marks the Project as done. Whether the last Project in a Campaign is done, and if so marks the Campaign as done.

elainekamlley:
**Campaign**
-  Inactive = UNPUBLISHED
-  Active and Needs Work = PUBLISHED and Open 
-  Active but Completed = Closed
-  _Closed Status determined by the published and completed projects (see projects)_

**Project**
- Inactive = UNPUBLISHED
- Active and Needs Work = PUBLISHED and Open 
- Active but Completed = Closed
- _Closed Status determined by the published and completed items (see items)_

**Item**
-  Inactive = UNPUBLISHED
-  Active and Needs Work = PUBLISHED and Open 
-  Active but Completed = Closed
-  _Closed Status determined by the cumulative status of assets - when completed_ 

**Asset** 
- Bulk status of assets dictated on publish at Item level     
- Inactive = Unpublished
- Active but Unpublished (“Hidden”) = Closed/Complete

- Active but no work so far = Blank (New)
- Active but Needs Work = Edit 
- Active and Ready to Review = Submit
- Active and Ready to Complete = Complete 
- Active but no action can be taken = Read

acdha:
With #358 this is technically no longer necessary but we should review performance before deciding whether we want to denormalize or cache the higher level objects’ calculated status to avoid a query.

acdha:
#358 now has a roll-up signal handler for Transcriptions which updates the Asset's `transcription_status` field so we don't need to do those calculations in views. The rest of these statuses can be calculated from that so I think the next step would be deciding where we apply those filters to determine whether there's a significant performance issue. 

@elainekamlley do we have wireframes / documentation for how we're handling campaigns and projects which have no work remaining to be done? Also, how does that interact with tagging?

PR #358, opened 2018-10-09
Update status model

acdha:
This commit removes the old status fields from the models which don't directly track the status and adds validation to the transitions using several new fields to track key decisions and the user who made them.

See #222


acdha:
@rstorey The major parts of this are done. The main thing which is left is updating the various places which filter on status values and measuring the performance impact to decide whether or not we need to cache the asset transcription status values or add some indexes (likely).

acdha:
@rstorey I updated the migration as part of adding a rollup transcription status field to Assets to reduce the complexity of the queries in the item detail view.

rstorey:
The migrations are working now, but travis is complaining.

acdha:
Yeah, there are some tests I haven't updated yet since the actual code is working correctly

acdha:
As of the last commits, there are tests covering the basic functionality and most of the branches. If you want to merge, that's fine by me.

PR #360, opened 2018-10-10
Campaign project thumbnails

rstorey:


concordia/models.py
```diff
@@ -63,13 +40,13 @@ class Campaign(MetricsModelMixin("campaign"), models.Model):
...
     published = models.BooleanField(default=False, blank=True)
-    status = models.CharField(
-        max_length=10, choices=Status.CHOICES, default=Status.DEFAULT
-    )
 
     title = models.CharField(max_length=80)
     slug = models.SlugField(max_length=80, unique=True)
     description = models.TextField(blank=True)
+    thumbnail_image = models.ImageField(
+        upload_to="campaign-thumbs", blank=True, null=True
```
acdha: I'd probably vote for “campaign-thumbnails” rather than the shorthand

concordia/models.py
```diff
@@ -88,15 +65,17 @@ class Project(MetricsModelMixin("project"), models.Model):
     objects = PublicationManager()
 
     campaign = models.ForeignKey(Campaign, on_delete=models.CASCADE)
+
+    published = models.BooleanField(default=False, blank=True)
+
     title = models.CharField(max_length=80)
     slug = models.SlugField(max_length=80)
+    thumbnail_image = models.ImageField(
+        upload_to="project-thumbs", blank=True, null=True
```
acdha: Same for “project-thumbnails”

concordia/settings_dev.py
```diff
@@ -21,10 +21,30 @@
...
 S3_BUCKET_NAME = "concordia-staticpages"
 
+DEFAULT_FILE_STORAGE = "storages.backends.s3boto3.S3Boto3Storage"
+AWS_STORAGE_BUCKET_NAME = S3_BUCKET_NAME
+AWS_DEFAULT_ACL = None  # Don't set an ACL on the files, inherit the bucket ACLs
+
+AWS_ACCESS_KEY_ID = os.getenv("AWS_ACCESS_KEY_ID")
+AWS_SECRET_ACCESS_KEY = os.getenv("AWS_SECRET_ACCESS_KEY")
+
+IMPORTER = {
```
acdha: We don't use `IMPORTER` at all any more

concordia/settings_dev.py
```diff
@@ -21,10 +21,30 @@
...
+    # /concordia_images is a docker volume shared by importer and concordia
+    "IMAGES_FOLDER": "/concordia_images/",
+    "S3_BUCKET_NAME": S3_BUCKET_NAME,
+}
+
+MEDIA_URL = "https://%s.s3.amazonaws.com/" % S3_BUCKET_NAME
+
+AWS_S3 = {
```
acdha: We should remove this - it's an artifact of the old code

concordia/settings_dev.py
```diff
@@ -21,10 +21,30 @@
...
 
-IMPORTER = {"IMAGES_FOLDER": "/tmp/concordia_images/"}
-
 S3_BUCKET_NAME = "concordia-staticpages"
 
+DEFAULT_FILE_STORAGE = "storages.backends.s3boto3.S3Boto3Storage"
+AWS_STORAGE_BUCKET_NAME = S3_BUCKET_NAME
+AWS_DEFAULT_ACL = None  # Don't set an ACL on the files, inherit the bucket ACLs
+
+AWS_ACCESS_KEY_ID = os.getenv("AWS_ACCESS_KEY_ID")
```
acdha: Do we need to manage these in the settings file versus just letting boto3 pick them up by default?

concordia/migrations/0006_thumbnail_images.py
```diff
@@ -0,0 +1,28 @@
+# Generated by Django 2.0.9 on 2018-10-10 18:20
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [("concordia", "0005_auto_20181010_1714")]
```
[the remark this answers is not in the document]
rstorey: Okay, I wasn't sure which was going to go first. I have another branch with a migration in it as well.

concordia/settings_dev.py
rstorey: boto3 picks them up from the IAM role in ECS, but when I'm testing locally they're in my environment.

concordia/settings_dev.py
rstorey: It looks like this is actually used in serializers.py

acdha: Those are also subject to removal – we could correct the media URL handling there but I'd want to check since most of those serializers are no longer called by a view in use. It's fine to defer that to a separate cleanup ticket, too.

concordia/settings_dev.py
acdha: Doesn't boto also pick up the environmental variables without us having to put them in the settings file?

rstorey:
I think i actually need the changes in your branch so this branch can run - it wants those ws views to still be existing in here.

PR #376, opened 2018-10-11
WIP: Item asset status display

acdha:
This starts improving the list of assets on the item detail page. It drops the use of the asset “title” (which is just the item id + sequence number) and displays the sequence number at the top in a partially transparent overlay which darkens on hover and the relevant action button (i.e. link to the asset detail page) in a similar format at the bottom. All assets will be displayed in a consistent size with the image centered as necessary depending on its aspect ratio. 

screen shot 2018-10-11 at 18 51 59-fullpage

Several other changes impact other pages:
* https://github.com/LibraryOfCongress/concordia/commit/1075c66733e6368360f583765865b127919fb639 removes the rules which overrode Bootstrap's default container behaviour to make `.container` effectively like `.container-fluid` with a fixed margin. Now it follows the normal responsive centering behaviour; if we want `.container-fluid`, we'll ask for it. This may require tweaks to margin/padding on pages which were relying on this customization but we can use the standard Bootstrap margin utilities for that.
* https://github.com/LibraryOfCongress/concordia/commit/f6b4126bdc0c236fde379e4d176fef7e580c74ed changes the project listing to be a little more similar in style.

Still to do: 
* displaying an overall progress indicator at the top
* displaying some indication for when an asset has tags


acdha:
@rstorey If you want to merge this now, we can get closer to @jbresner's design tomorrow in another pull request

jbresner:
@acdha - here's the only "Item Record" page I created:
item record

Let me know if you need this updated (based on previous conversations I may not have been part of...).

PR #377, opened 2018-10-12
Item detail progress bar

### LINKED ISSUE #158: Progress percentage to show level of completion on an item
As a user, I want to know how complete an item is so that I can decide if I want to click it and transcribe

Acceptance criteria:
- [x] A function that does the math to generates a percentage based on how many items were completed and items that are open for editing
- [x] UI of percentage based on wireframe

https://github.com/LibraryOfCongress/concordia/issues/41

acdha:
This adds a basic progress bar to the item detail page next to the title above the asset listing:

screenshot_2018-10-12 crowd rosa parks papers general correspondence 1928-2006 letters and drawings from children 1986 4

I tried including text labels in the bar but the text will be truncated when the values aren't fairly large (e.g. >10%) so I made them tooltips instead:

screenshot_2018-10-12 crowd rosa parks papers general correspondence 1928-2006 letters and drawings from children 1986

Displaying the edit state including items which have not been transcribed yet makes our progress bar always add up to 100% which felt misleading, so the final version above only lists assets which have at least some transcription progress in the “In Progress” state:

screenshot_2018-10-12 crowd rosa parks papers general correspondence 1928-2006 letters and drawings from children 1986 1

Closes #158

### DIFF
--- a/concordia/forms.py
+++ b/concordia/forms.py
@@ -79,18 +79,8 @@ class AssetFilteringForm(forms.Form):
         widget=forms.Select(attrs={"class": "form-control"}),
     )
 
-    def __init__(self, asset_qs, *args, **kwargs):
+    def __init__(self, status_counts, *args, **kwargs):
         super().__init__(*args, **kwargs)
-
-        # We want to get a list of all of the available asset states in this
-        # item's assets and will return that with the preferred display labels
-        # including the asset count to be displayed in the filter UI
-        asset_state_qs = asset_qs.values_list("transcription_status")
-        asset_state_qs = asset_state_qs.annotate(
-            Count("transcription_status")
-        ).order_by()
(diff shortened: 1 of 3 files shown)


acdha:
One policy question: do we display 0% or just not show the progress bar when there is no progress?
screenshot_2018-10-12 crowd rosa parks papers writings notes and statements 1956-1998 drafts of early writings account

acdha:
That's what I figured, too, but I wanted to run it by @colefemeno as well

colefemeno:
Agree with Rosie - no need to show 0% and we can anticipate items will go quickly into progress

acdha:
What does “no need to show 0%” mean — hide the progress bar entirely if we don't have anything to show?

Issue #168, opened 2018-09-14
Password strength validation


elainekamlley:
As a user registering for an account, I want to know how strong my password is so that I can feel confident that my profile is as safe

https://pwned-passwords-django.readthedocs.io/

https://github.com/dropbox/zxcvbn

Backend validation for password strength. It’ll get things like Password123 which satisfy most algorithmic requirements but are common on widely shared dictionaries

acdha:
pwned-passwords-django uses an outside service which tries to be privacy preserving but may still require approval:

https://pwned-passwords-django.readthedocs.io/en/1.3.1/faq.html#how-can-this-be-secure-it-s-sending-passwords-to-some-random-site

zxcbn will be a huge improvement over our existing validator which accepts very weak common passwords

jbresner:
The current registration page has both recommendations and requirements for a users password:

- Your password can't be too similar to your other personal information.
- Your password must contain at least 8 characters.
- Your password can't be a commonly used password.
- Your password can't be entirely numeric.
- Your password fails to meet our complexity requirements.

Some of these look like requirements (2 and 4) but some feel a lot like recommendations (1, 3, and 5...what are the LOC's complexity requirements?). Do we plan on providing feedback as a user enters a password? Unless we can provide clear guidelines on how to create a password, I'd consider removing the recommendations.

acdha:
The plan is to provide feedback once that bit of JavaScript is cleaned up rather than just telling you rules

acdha:
We do most of what's in this ticket now that #386 has merged. I'm voting that we close this and create a ticket to switch to zxcvbn later.

PR #386, opened 2018-10-15
Update password complexity UI

acdha:
This adds basic warnings for our complexity requirements but is not as strong as using [zxcvbn](https://github.com/dropbox/zxcvbn).

See #168

### DIFF
--- a/concordia/static/js/password-validation.js
+++ b/concordia/static/js/password-validation.js
@@ -1,0 +1,73 @@
+/* global jQuery */
+
+(function($) {
+    var requirements = [
+        {
+            id: 'pw-length',
+            text: 'At least 8 characters long',
+            test: function(i) {
+                return i.match(/.{8,}/);
+            }
+        },
+        {
+            id: 'pw-uppercase',
+            text: '1 or more uppercase characters',
+            test: function(i) {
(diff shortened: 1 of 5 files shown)


concordia/static/js/password-validation.js
```diff
@@ -0,0 +1,73 @@
...
+        {
+            id: 'pw-digits',
+            text: '1 or more digits',
+            test: function(i) {
+                return i.match(/[0-9]/);
+            }
+        },
+        {
+            id: 'pw-special',
+            text: '1 or more non-unique special characters',
```
rstorey: Do we really have to say "non-unique" here, if we're allowing a single special character to satisfy the requirement? Are the digits non-unique as well?

concordia/static/js/password-validation.js
```diff
@@ -0,0 +1,73 @@
...
+            }
+        });
+
+        if (validity) {
+            this.removeAttribute('aria-invalid');
+            this.setCustomValidity('');
+        } else {
+            this.setAttribute('aria-invalid', 'true');
+            this.setCustomValidity(
+                'Your password does not meet our requirements'
```
rstorey: I'd prefer this to read "the requirements" instead of "our requirements"

PR #390, opened 2018-10-15
Add pre-commit hooks

acdha:
This provides an easy way to get black, isort, flake8, and prettier to run on every commit using [pre-commit](https://pre-commit.com/) (`brew install pre-commit`)

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,18 @@
+exclude: '.*/vendor/.*'
+repos:
+    - repo: https://github.com/pre-commit/mirrors-isort
+      rev: v4.3.4
+      hooks:
+          - id: isort
+    - repo: https://github.com/ambv/black
+      rev: stable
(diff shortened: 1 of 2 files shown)


acdha:
@rstorey I just made https://github.com/LibraryOfCongress/concordia/pull/394/commits/84b674c3e2bb47efb1e54a74db55288c38f0a7ea to deal with the fact that Prettier's HTML support hasn't quite been released – see https://github.com/prettier/prettier/issues/5098

acdha:
Okay, this ended up being more annoying than it should have been — I needed to make 8165fbf86e63ca3a2bb85c14cb673cefff29e2f8 to avoid isort & black conflicting on some files (they disagree on how to handle multi-line import blocks). I decided that black was more important to me.

https://pre-commit.com/hooks.html has some other things we might want to enable for JSON, line-termination, lines at the ends of files, etc.

Issue #350, opened 2018-10-05
Simplify ALB configuration


acdha:
Currently we're setting up a default rule to a target group which doesn't work:

https://github.com/LibraryOfCongress/concordia/blob/master/cloudformation/infrastructure/load-balancers.yaml#L36-L44

The service still works because we add a /* path pointing to the group which is setup as part of the ECS service:

https://github.com/LibraryOfCongress/concordia/blob/master/cloudformation/services/concordia-service/service.yaml#L148-L173

I think the way to resolve this might just be not to create a listener at all in load-balancers.yaml so the one created by the concordia-service will be the default and only listener.

rstorey:
Except you can't create a listener with zero rules, as far as I know.

acdha:
I had to create a load balancer with no listeners and then let the app stack add the first (only) listener

rstorey:
Ah ok. I know if you try to delete all the rules through the console UI, it won't let you do that.

rstorey:
This is done

PR #395, opened 2018-10-15
Change cluster type

### LINKED ISSUE #323: Enable Fargate ECS deployments
Our CloudFormation template needs to specify Fargate as the deployment type

That's probably:

Task definition:
```
"RequiresCompatibilities": ["FARGATE"],
```

Service definition:
```
"LaunchType": "FARGATE",
```

See also #236 as this will restrict the network setup types

rstorey:
Closes #323 and #350


cloudformation/infrastructure/load-balancers.yaml
```diff
@@ -32,27 +28,6 @@ Resources:
             Tags: 
                 - Key: Name
                   Value: !Ref EnvironmentName
-
-    LoadBalancerListener:
```
acdha: This looks like it should be a nice simplification since that will avoid the possibility of anything ending up in a partially-deployed state with the default listener

rstorey:
I don't think we can close #350 yet with this PR, at least not as it is currently. Although I was able to get a Fargate cluster up and running with these templates, I had to add another non-default listener rule to the load balancer listener in order to do so. That seems to me like an AWS weirdness / bug. These templates will yield an environment where that one additional rule still needs to be added, which means #350 isn't really solved. Would appreciate another set of eyes on it - or maybe we punt on #350 for now.

cloudformation/create_secrets.sh
```diff
@@ -0,0 +1,19 @@
...
+
+export ENV_NAME=dev-fg
+
+export DJANGO_SECRET_KEY=
+export DB_HOST=
+export DB_IDENTIFIER=
+export DB_PASSWORD=
+export KMS_KEY_ARN=
+
+aws secretsmanager create-secret --name "crowd/${ENV_NAME}/Django/SecretKey" --kms-key-id "${KMS_KEY_ARN}" --secret-string '{"DjangoSecretKey": "${DJANGO_SECRET_KEY}"}'
```
rstorey: These commands also don't work (well, they execute, but they create secrets that have shell syntax instead of the actual secret value)

cloudformation/create_secrets.sh
```diff
@@ -0,0 +1,19 @@
...
+
+aws secretsmanager create-secret --name "crowd/${ENV_NAME}/DB/MasterUserPassword" --kms-key-id "${KMS_KEY_ARN}" --secret-string '{"username": "concordia","engine": "postgres","port": 5432,"dbname": "concordia","password": "${DB_PASSWORD}","dbInstanceIdentifier": "${DB_IDENTIFIER}","host": "${DB_HOST}"}'
```
rstorey: This one is also problematic

cloudformation/infrastructure/fargate-cluster.yaml
```diff
@@ -0,0 +1,282 @@
...
+        Type: AWS::EC2::SecurityGroup::Id
+
+    VpcId:
+        Description: The Id of the VPC for this cluster
+        Type: AWS::EC2::VPC::Id
+
+    ConcordiaVersion:
+        Type: String
+        Description: version of concordia, concordia/importer, and rabbitmq docker images to pull and deploy
+        Default: latest
```
rstorey: We're going to want to change this for prod / stage / test

cloudformation/infrastructure/fargate-cluster.yaml
```diff
@@ -0,0 +1,282 @@
...
+      ListenerArn: !Ref ExternalLoadBalancerListener
+      Priority: 1
+      Conditions:
+          - Field: path-pattern
+            Values:
+                - /*
+      Actions:
+          - TargetGroupArn: !Ref ConcordiaExternalTargetGroup
+            Type: forward
+
```
rstorey: The SSL listener rule will have to be added manually because getting the cert is done manually

cloudformation/infrastructure/fargate-cluster.yaml
```diff
@@ -0,0 +1,282 @@
...
+              awslogs-stream-prefix: ConcordiaWorker
+          Environment:
+              - Name: AWS
+                Value: '1'
+              - Name: ENV_NAME
+                Value: !Ref EnvName
+              - Name: S3_BUCKET_NAME
+                Value: concordia-staticpages
+              - Name: CELERY_BROKER_URL
+                Value: pyamqp://guest@rabbit:5672
```
rstorey: This actually needs to be localhost, not rabbit

cloudformation/infrastructure/fargate-cluster.yaml
```diff
@@ -0,0 +1,282 @@
...
+              awslogs-stream-prefix: ConcordiaServer
+          Environment:
+              - Name: AWS
+                Value: '1'
+              - Name: ENV_NAME
+                Value: !Ref EnvName
+              - Name: S3_BUCKET_NAME
+                Value: concordia-staticpages
+              - Name: CELERY_BROKER_URL
+                Value: pyamqp://guest@rabbit:5672
```
rstorey: This needs to be localhost, not rabbit

cloudformation/create_secrets.sh
acdha: That's due to the double-quoting:

```
cadams@jupiter:~/Projects/concordia/cloudformation [git fargate-cluster] $ shellcheck create_secrets.sh 

In create_secrets.sh line 17:
aws secretsmanager create-secret --name "crowd/${ENV_NAME}/Django/SecretKey" --kms-key-id "${KMS_KEY_ARN}" --secret-string '{"DjangoSecretKey": "${DJANGO_SECRET_KEY}"}'
                                                                                                                           ^-- SC2016: Expressions don't expand in single quotes, use double quotes for that.


In create_secrets.sh line 19:
aws secretsmanager create-secret --name "crowd/${ENV_NAME}/DB/MasterUserPassword" --kms-key-id "${KMS_KEY_ARN}" --secret-string '{"username": "concordia","engine": "postgres","port": 5432,"dbname": "concordia","password": "${DB_PASSWORD}","dbInstanceIdentifier": "${DB_IDENTIFIER}","host": "${DB_HOST}"}'
                                                                                                                                ^-- SC2016: Expressions don't expand in single quotes, use double quotes for that.
```

(`brew install shellcheck`, I also use https://github.com/timonwong/vscode-shellcheck)

cloudformation/infrastructure/fargate-cluster.yaml
```diff
@@ -0,0 +1,282 @@
...
+            - 'prod'
+            - 'dev-fg'
+        ConstraintDescription: Must match a location for secret storage in secretsmanager
+
+
+Resources:
+
+  ConcordiaServiceSecretAccessPolicy:
+      Type: AWS::IAM::Policy
+      Properties:
```
acdha: This should have a `PolicyName` declaration

cloudformation/infrastructure/fargate-cluster.yaml
```diff
@@ -0,0 +1,282 @@
...
+    Properties:
+      DefaultActions:
+        - TargetGroupArn: !Ref 'ConcordiaExternalTargetGroup'
+          Type: forward
+      LoadBalancerArn:  !Ref LoadBalancer
+      Port: 80
+      Protocol: HTTP
+    Type: AWS::ElasticLoadBalancingV2::Listener
+
+  HttpListenerRule:
```
acdha: I don't think we need the listener rule since the default forwarder will take care of it

cloudformation/infrastructure/fargate-cluster.yaml
acdha: Can't we reference a pre-defined cert in IAM? I'm assuming that'll be something like `!Sub 'arn:aws:iam::${AWS::AccountId}':server-certificate/${CanonicalHostName}`, assuming that we also set a property which defines the public hostname we want to use which would also be handy to pass to the container to seed the sites framework.

rstorey:
Remaining TODOs:
- [ ] Add redis / elasticache
- [ ] Automated process for deploying tagged versions to test / stage / prod without tearing down the stack
- [ ] Figure out load balancer listener rule issues
- [ ] Automated process for SSL certificate and https listener

rstorey:
I propose you evaluate this PR as-is and we ticket the remaining two TODOs separately. This branch is running in the repo AWS account - cluster crowd-dev-fg2

rstorey:
Tickets have been created

rstorey:
I'm deploying it to crowd-prod now, and I already know what you'll encounter when you test. I'd like to get your help with it. The instructions are in the cloudformation/README.md file. Essentially we need to streamline how the DB password and host get saved into secretsmanager. I wanna do it more like how you have it in your sentry templates so the create_secrets.sh script can be eliminated.

acdha:
Yeah, I've found that secrets manager is a LOT more work than using the systems manager properties and it costs more, too.

rstorey:
I thought it would be the opposite since systems manager seems designed for managing systems that we aren't running, but secrets manager is just too much.

PR #400, opened 2018-10-16
help center build out

rstorey:
This doesn't have uniform sizes for the Guides cards on the help center page; and it doesn't include subnav for the help center 2nd level pages yet.


acdha:
https://github.com/LibraryOfCongress/concordia/pull/401/commits/a081c7924f171eb3bd0f45dace55cb2b9228dd1f is a simpler implementation of the generic breadcrumb loading – it handles more than one level deep but does not attempt to load the markdown title

PR #404, opened 2018-10-17
More UI updates

acdha:
Some updates reducing CSS leakage from the homepage styles and making the static pages nicer. Here are some screenshots:

image
image
image
image

### DIFF
--- a/concordia/static/css/base.css
+++ b/concordia/static/css/base.css
@@ -53,11 +53,13 @@ body {
 
 h1 {
     font-family: 'Roboto Slab', serif;
-    font-size: 3rem;
-    line-height: 125%;
     font-weight: bold;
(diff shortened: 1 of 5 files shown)


jbresner:
One minor UI fix...the home page banner area's text box is oversized (legacy from when it was first built). The text container at the largest breakpoint (1200px) should be 420x242px in height (the height is variable based on the amount of text). The box also has a left-margin of 82px. As for it being responsive, I was envisioning it following what we're doing on the LOC.gov home page with the text moving below the image. If you need specific CSS/breakpoints, let me know.

acdha:
I don’t think much has been done for responsiveness yet. I’ll review later. 

I’m hoping to get major UI work done today so we can focus on deployment updates tomorrow.

Issue #381, opened 2018-10-12
Admin importer changes


elainekamlley:
@rstorey some things I noticed when working in the importer for the demo 

- [x] Still getting a 500 error when doing a bulk import. The initial thought as to why, was that there was no project description in the spreadsheet. I added p. description and still got a 500 error. Still not sure why. However the first line in the spreadsheet was imported.  Could sit down with you and do a walk through. 
- [x] Implement hierarchal publishing at the item level to publish all assets in an item. CMs may not want to publish every item in a project but every asset in an item should. 
- [x] Add project description field in the project admin 
- [x] Map the project description to the project page

rstorey:
Did you try deleting all the blank lines that were in your test spreadsheet? There were over a thousand lines in there.

rstorey:
Nevermind - I found the reason is that the project titles are too long (or the way we're creating slugs which have a limit of 80 characters needs to be adjusted). Once I reduced the length of the project titles, I was able to get your spreadsheet to import in crowd-dev.

elainekamlley:
Ohhhhh, ok. Mind making that update on the instructions? We'll make adjustment on our end.

elainekamlley:
Nah, project titles should be short. The project description will provide more context

PR #405, opened 2018-10-17
Admin importer fixes for #381

rstorey:


concordia/admin.py
```diff
@@ -103,11 +103,12 @@ def admin_bulk_import_view(request):
                 import_url_blob = row["Import URLs"]
 
                 if not all((campaign_title, project_title, import_url_blob)):
-                    messages.add_message(
-                        request,
-                        messages.WARNING,
-                        f"Skipping row {idx}: at least one required field (Campaign, Project, Import URLs) is empty",
+                    warning_message = (
+                        f"Skipping row {idx}: at least one required field \
```
acdha: Putting the line continuation in there puts all of the subsequent indentation into the string. I'd just split this into two implicitly-concatenated strings:

```suggestion
                        f"Skipping row {idx}: at least one required field" \
```

concordia/admin.py
```diff
@@ -103,11 +103,12 @@ def admin_bulk_import_view(request):
...
+                        (Campaign, Project, Import URLs) is empty",
```
acdha:
```suggestion
                        " (Campaign, Project, Import URLs) is empty",
```

concordia/admin.py
rstorey: Done

rstorey:
See #381

concordia/admin.py
```diff
@@ -30,6 +30,46 @@
...
 
+def publish_item_action(modeladmin, request, queryset):
+    """
+    Mark all of the selected items and their related assets as published
+    """
+
+    count = queryset.filter(published=False).update(published=True)
+    asset_count = 0
+    for item in queryset:
+        asset_count += Asset.objects.filter(item=item, published=False).update(
```
acdha: I think we could avoid the loop by making this `asset_count += Asset.objects.filter(item__in=queryset, …)`

rstorey: I was going to ask if there was a way to avoid the loop. Thanks

PR #406, opened 2018-10-17
Consistent page titles

acdha:
* All titles are now prefixed with “Crowd: ”
* Titles within hierarchy use the most specific value first followed by the parent title(s) in parentheses:
  * Crowd: Home
  * Crowd: About Make Your Mark
  * Crowd: Abraham Lincoln
  * Crowd: Untranscribed Correspondence (Abraham Lincoln)
  * Crowd: Abraham Lincoln papers: Series 1. General Correspondence. 1833-1916: Asher Robbins, Friday, June 17, 1836 (Printed Speech) (Abraham Lincoln: Untranscribed Correspondence)
  * Crowd: mal0003400-1 (Abraham Lincoln: Untranscribed Correspondence)

### DIFF
--- a/concordia/templates/404.html
+++ b/concordia/templates/404.html
@@ -5,7 +5,7 @@
 <head>
     <meta charset="UTF-8">
     <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no">
-    <title>{% block title %}404 error{% endblock title %}</title>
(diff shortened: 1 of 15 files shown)


acdha:
Sounds good – we can truncate them but that tends to have its own sources of complaint so I'd want to see how much users care about this

Issue #216, opened 2018-09-20
Keep MS Word characters out of transcriptions


rstorey:
Steps to reproduce:
1. Load an asset that needs to be transcribed.
2. Write a transcription in MS Word or equivalent that contains double quotes.
3. Paste the transcription into the text area and save it.

Observed behavior:
1. Ambiguous

Expected behavior:
1. If a user types in or pastes in certain characters such as MS Word specific characters, they should be scrubbed or converted.

rstorey:
We have #410 if it turns out that we need it.

PR #410, opened 2018-10-17
change MS Word smart quotes to ASCII equivalents in transcription text

rstorey:
Refs #216

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -36,6 +36,7 @@ markdown = ">=3.0.1"
 "django-bootstrap4" = "*"
 django-robots = "*"
 setuptools-scm = "*"
+"beautifulsoup4" = "*"
 
 [dev-packages]
 invoke = "*"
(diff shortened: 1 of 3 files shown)


acdha:
Do we have an example of the problem this is trying to solve? Unless someone's browser is extremely broken we should be getting UTF-8 in and out.

rstorey:
So all modern browsers will automatically convert Microsoft characters to UTF? I don't use Windows or Word much, but the idea is that if someone copies and pastes from MS Word on Windows we don't want windows-1252 chars in there. If it's really not an issue on Windows anymore, we can just close this PR.

acdha:
By default, forms are submitted using the same encoding as the document, which is set to UTF-8 by default (`Content-Type: text/html; charset=utf-8`). Clipboard contents should be converted on paste if they're not originating as as Unicode. I would want to try to verify a problem before making a change.

Issue #213, opened 2018-09-20
Prevent URLs in transcriptions


rstorey:
Steps to reproduce:

1. As an anonymous user, paste a URL into a transcription and save it.
2. Repeat for as many transcriptions as you can!

Observed behavior:
1. We know that URLs won't be present in these historical source materials. 
2. We can use the presence of URLs to judge the quality of a transcription.

Expected behavior:
1. The application should somehow flag transcriptions with URLS for community manager review.
2. Perhaps we should not allow URLs to be saved in transcriptions at all.

PR #414, opened 2018-10-17
don't allow users to put URLs in transcription text

rstorey:
Refs #213

### DIFF
--- a/concordia/views.py
+++ b/concordia/views.py
@@ -404,7 +404,7 @@ def save_transcription(request, *, asset_pk):
     # If so, ask the user to correct the transcription by removing the URLs
     transcription_text = request.POST["text"]
     url_match = re.search(URL_REGEX, transcription_text)
-    if urls:
+    if url_match:
         return JsonResponse(
             {
                 "error": "It looks like your text contains URLs. "
--- a/concordia/views.py
+++ b/concordia/views.py
@@ -1,5 +1,6 @@
 import json
 import os
+import re
 import time
 from datetime import datetime, timedelta
 from logging import getLogger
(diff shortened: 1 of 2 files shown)


concordia/views.py
```diff
@@ -50,6 +51,7 @@
 ASSETS_PER_PAGE = 36
 PROJECTS_PER_PAGE = 36
 ITEMS_PER_PAGE = 36
+URL_REGEX = "http[s]?://"
```
acdha: I try to always prefix regular expressions with `r` to mark them as [raw string literals](https://docs.python.org/3/reference/lexical_analysis.html#string-and-bytes-literals). In addition to not needing to escape back-slashes, in many editors and on GitHub that activates regex syntax highlighting:

```suggestion
URL_REGEX = r"http[s]?://"
```

concordia/views.py
```diff
@@ -398,6 +400,19 @@ def save_transcription(request, *, asset_pk):
     else:
         user = request.user
 
+    # Check whether this transcription text contains any URLs
+    # If so, ask the user to correct the transcription by removing the URLs
+    transcription_text = request.POST["text"]
+    urls = re.search(URL_REGEX, transcription_text)
```
acdha:
```suggestion
    url_match = re.search(URL_REGEX, transcription_text)
```

concordia/views.py
```diff
@@ -398,6 +400,19 @@ def save_transcription(request, *, asset_pk):
...
+    if urls:
```
acdha:
```suggestion
    if url_match:
```

concordia/views.py
rstorey: I meant to do that, but I forgot

Issue #265, opened 2018-09-24
Revise footer 


colefemeno:
We have the full scope of the footer - we need to revisit the design and what is communicated in this space. 

A footer provides related navigation and information to other Library of Congress web entities, suggests social media and other places to find information, and communicates up to date information about activity on the site. 

To Do: 

- [x] Reach out to MS to confirm what we absolutely need to keep
- [x] remove links/icons we don't need 
- [x] update links to correct accounts

acdha:
On a more mundane level, many of the links in https://github.com/LibraryOfCongress/concordia/blob/master/concordia/templates/base.html#L94-L134 are broken because they're relative paths intended to link to pages like https://www.loc.gov/connect/. If you want to review these I can fix those at the same time.

elainekamlley:
Reached out to Michelle and we need legal, IG, accessibility & external links disclaimer. We will utilize the other links for navigating on Make Your Mark and going to History Hub. @acdha other than the four identified we can update those links names and links to where they need to go. 

Proposed footer - @colefemeno what do you think?  

Under `follow us`
- Twitter icon that goes to crowd_loc
- Add GitHub??
- Everything else will be removed 

Links: 

- Accessibility
- Legal
- IG
- External link disclaimer
- Contact (goes to MYM contact) 
- Discuss (goes to History Hub)

Fix congress.gov and copyright logos to link

jbresner:
Here are a few footer layouts. I removed the "Follow Us" since it was only ever a requirement on the LOC.gov page (it's actually a separate link). Let me know if you have any thoughts on the mockups.
![footer - v1](https://user-images.githubusercontent.com/4275957/46711941-a4277500-cc1c-11e8-987e-b4e4ba3cc32d.png)
![footer - v2](https://user-images.githubusercontent.com/4275957/46711942-a4277500-cc1c-11e8-9ce3-e26154088247.png)
![footer - v3](https://user-images.githubusercontent.com/4275957/46711943-a4277500-cc1c-11e8-91a9-774575ffe719.png)

elainekamlley:
@jbresner Now that I think of it we may be able to remove congress.gov and copyright. I'll follow up with MS

elainekamlley:
Newsletter link: https://updates.loc.gov/accounts/USLOC/subscriber/new?topic_id=USLOC_175
History hub link: https://historyhub.history.gov/community/make-your-mark/

jbresner:
[...]

PR #422, opened 2018-10-18
Revise footer

acdha:
This is a first pass at the updates in #265. It still has a fair amount of duplication with the standard Bootstrap classes but I don't think it's worth the time to align them more.

Screenshots at various sizes:
image
image
image
image
image
image


rstorey:
Is the all-caps FOLLOW US intentional? I thought it was meant to be Title Case

acdha:
No, that's one of the overly-broad customizations we inherited

PR #424, opened 2018-10-18
Web performance updates

### LINKED ISSUE #218: Ensure home page is eligible for caching
Steps to reproduce:
1. Set up a website cache for the home page
2. Attempt to use it

Observed behavior:
1. The home page is not cacheable because the top right nav block changes on user login.

Expected behavior:
1. Use javascript front end code to populate the top right nav block so the home page can be cached

### LINKED ISSUE #219: Review for website caching

acdha:
This branch will make our main landing pages cacheable and otherwise review site performance

* [x] Move user login check into JavaScript so base.html doesn't check the session
* [x] Review page resource loading performance

Closes #218 
Closes #219

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -10,7 +10,6 @@ gunicorn = "*"
 djangorestframework = "*"
 celery = "*"
 coreapi = "*"
-django-machina = "*"
 django-haystack = "*"
 "boto3" = ">=1.9.16"
 botocore = '>=1.12.16'
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -1,7 +1,7 @@
 {
     "_meta": {
         "hash": {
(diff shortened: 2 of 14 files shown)


acdha:
At this point I should be able to mark the static pages & everything but the asset detail view as cacheable but that will have to wait until after the meeting.

acdha:
At this point, all of the public views are cacheable. The asset transcription view is the major exception so it's still `never_cache()`-ed until we have time to work on that.

A quick review of the pages suggests one win would be consolidating from multiple CDNs to a single one since Bootstrap and jQuery are available on CDNJS as well. That can be a separate ticket.

PR #439, opened 2018-10-19
Item page edits

rstorey:
Move the Filter label from on top of the input group to part of the input group, remove the word "Assets" from the form.

Also adjusts the caption of the progress bar to make it smaller and lighter.

### DIFF
--- a/concordia/templates/transcriptions/item_detail.html
+++ b/concordia/templates/transcriptions/item_detail.html
@@ -51,9 +51,12 @@
     </div>
     <div class="row">
         <form method="get" class="w-50 mx-auto">
-            <h6>Filter Assets:</h6>
             <div class="form-group row">
+            
                 <div class="input-group input-group-sm">
+              <div class="input-group-prepend">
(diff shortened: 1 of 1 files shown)


concordia/templates/transcriptions/item_detail.html
```diff
@@ -51,9 +51,11 @@ <h1 class="m-3">{{ item.title }}</h1>
     </div>
     <div class="row">
         <form method="get" class="w-50 mx-auto">
-            <h6>Filter Assets:</h6>
             <div class="form-group row">
                 <div class="input-group input-group-sm">
+                    <div class="input-group-prepend">
+                        <span class="input-group-text" id="inputGroup-sizing-sm">Filter:</span>
```
acdha: We're not using this ID, are we?

```suggestion
                        <span class="input-group-text">Filter:</span>
```

rstorey:
Looks like this now:
screen shot 2018-10-19 at 1 06 23 pm

Issue #437, opened 2018-10-19
Add National Digital Trust details to About page


colefemeno:
To do

- [x] Update About page with National Digital Trust details

colefemeno:
Also updated READ.ME and Help Center/FAQ

PR #445, opened 2018-10-19
Add details of National Digital Trust to About

colefemeno:
Updating the About page to include support from National Digital Trust and collaboration


colefemeno:
closes #437

PR #450, opened 2018-10-19
Fix bulk project creation error message reporting

acdha:
One message.error call was missing the request argument in e375af570c829b55e4503afc459e948e19057a4b


rstorey:
But what about the underlying issue there? If the project and campaign with that slug already exist, should it not be retrieved in the `get_or_create` method? Or was I being a bad user?

PR #459, opened 2018-10-21
Add links to view item / assets on loc.gov

### LINKED ISSUE #432: Expose item metadata from button on transcription page
This is something we've talked about I'm not sure it ever made it into a ticket. Other than showing the breadcrumb info, can we also expose the underlying metadata for a particular image? If not, then we should add to the instructions how people can find a page they're transcribing back over in P1.

rstorey:
Closes #432


importer/tasks.py
```diff
@@ -385,6 +387,7 @@ def get_asset_urls_from_item_resources(resources):
     """
 
     assets = []
+    item_resource_url = resources[0]["url"] or ""
```
acdha: Do we have to worry about getting more than one of these? It seems like we might want to at least log if there are discrepancies

rstorey: I'm not sure, actually. I was hoping to have a chance to test this with the real spreadsheet in dev / test.

PR #463, opened 2018-10-22
Paginated filters

acdha:


acdha:
Note that this builds on your loc.gov view branch since I forgot to back out the database migration first and that should be merged soon. Feel free to cherry-pick if not since this one doesn't touch the database at all.

PR #467, opened 2018-10-22
Add env vars for cache endpoints to legacy deployments

rstorey:


cloudformation/services/concordia-service/service.yaml
```diff
@@ -39,31 +39,34 @@ Parameters:
 
 Mappings:
     ClusterMapping:
+        RedisEndpointMap:
+            dev: 'crowd-dev-redis.hfkoms.0001.use1.cache.amazonaws.com'
```
acdha: Does the stack which creates these resources export their outputs? If there's a reason why we can't do that, I'm thinking the answer might be migrating to Route53 so we could say "redis.{environment}.internaltld"

cloudformation/services/concordia-service/service.yaml
```diff
@@ -39,31 +39,34 @@ Parameters:
...
+            stage: 'crowd-stage-redis.hfkoms.0001.use1.cache.amazonaws.com'
+        MemcachedEndpointMap:
+            dev: 'crowd-dev-cache.hfkoms.cfg.use1.cache.amazonaws.com'
+            test: 'crowd-test-cache.hfkoms.cfg.use1.cache.amazonaws.com'
+            stage: 'crowd-stage-cache.hfkoms.cfg.use1.cache.amazonaws.com'
         AutoScalingServiceRoleMap:
             dev: 'arn:aws:iam::351149051428:role/crowd-dev-ECS-RQZSJ89GYY4-ECSServiceAutoScalingRol-KJ5SO8E1KIWF'
             test: 'arn:aws:iam::351149051428:role/crowd-test2-ECS-1I34ZK7Z3-ECSServiceAutoScalingRol-U5GIQOZ1RQ8Z'
             stage: 'arn:aws:iam::351149051428:role/crowd-stage-ECS-BU57ZZJOQ-ECSServiceAutoScalingRol-1EK5KYLQQTVVU'
-            prod: 'arn:aws:iam::351149051428:role/crowd-prod-ECS-1501XCPSDM-ECSServiceAutoScalingRol-JUB3TZXYH65T'
```
acdha: It's unclear to me why the prod mappings were removed?

rstorey:
I should have put this in the description: This template is only used for the existing dev, test and stage deployments. The prod deployment (and all new deployments) uses the fargate-cluster.yaml template and not this one. I'm only keeping this template around as long as we have the existing deployments to support - hopefully not much longer.

acdha:
Ah, that makes it much more clear

PR #468, opened 2018-10-22
WIP: search indexing

rstorey:


concordia/models.py
```diff
@@ -240,6 +240,9 @@ class UserAssetTagCollection(
     def __str__(self):
         return "{} - {}".format(self.asset, self.user)
 
+    def tags_to_string(self):
+        return ", ".join(t.value for t in self.tags.all())
```
acdha: I thought ElasticSearch supported multi-value fields — can't this just be a list?

rstorey: Maybe it can? I'll look into it after we get through upcoming deployments.

rstorey: Done

PR #478, opened 2018-10-22
Expose environment to RavenJS

acdha:
This allows the Sentry reports to be filtered based on the environment
just as we're doing on the backend.


acdha:
Indeed: we need to add that to the task definition

PR #480, opened 2018-10-22
Ratelimit

rstorey:
First pass at django-ratelimit.

Not sure what the actual rate values should be, or how to rate-limit the django-registration module.

### DIFF
--- a/concordia/views.py
+++ b/concordia/views.py
@@ -34,6 +34,7 @@ from django.views.decorators.http import require_POST
 from django.views.decorators.vary import vary_on_headers
 from django.views.generic import DetailView, FormView, ListView, TemplateView
 from django_registration.backends.activation.views import RegistrationView
+from ratelimit.decorators import ratelimit
 
 from concordia.forms import (
     AssetFilteringForm,
@@ -47,10 +48,10 @@ from concordia.models import (
     Campaign,
     Item,
     Project,
+    Tag,
     Transcription,
     TranscriptionStatus,
     UserAssetTagCollection,
-    Tag,
 )
 from concordia.version import get_concordia_version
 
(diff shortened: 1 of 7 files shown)


concordia/views.py
```diff
@@ -795,6 +804,10 @@ def add_transcription_status_summary_to_projects(self, projects):
             project.transcription_statuses
 
 
+reserve_rate = lambda g, r: None if r.user.is_authenticated else "1/m"
```
acdha: This might want to be a bit higher since we trigger requests at 1/m and timing skew could hit it. I'm thinking 2-3/m should suffice for a single page but how much do we care about people having multiple windows open?

acdha:
For rate-limiting the authentication functions, I think that means something like adding a path for the login view in urls.py which uses the decorated `ratelimit(django.contrib.auth.views.login, …)`, and the same for the registration view, using the request.post.username field as the key for the per-user limits and the IP address for a second tier.

concordia/views.py
```diff
@@ -804,7 +806,8 @@ def add_transcription_status_summary_to_projects(self, projects):
             project.transcription_statuses
 
 
-reserve_rate = lambda g, r: None if r.user.is_authenticated else "1/m"
+def reserve_rate(g, r):
+    return None if r.user.is_authenticated else "12/m"
```
acdha: The concern I have here is that we're basically saying 12 anonymous users can work at a time. That number seems low but I'm not sure how much higher we want to make it before it's not worth having the rate-limit on this view at all.

rstorey: Doesn't the key='ip' bit mean this rate is 12/m per ip? Maybe I read the docs wrong.

rstorey: Ah yes. Flake8 didn't like the lambda functions, and black puts those def's a couple lines above the @ratelimit decorator. Not the most beautiful thing in the world. Should I wait to merge this until we have a chance to integrate with registration and login, and do the HTTP 429 thing?

PR #483, opened 2018-10-22
Add CONCORDIA_ENVIRONMENT to task definition

rstorey:


cloudformation/infrastructure/fargate-cluster.yaml
```diff
@@ -42,6 +42,15 @@ Parameters:
             - 'prod'
         ConstraintDescription: Must match a location for secret storage in secretsmanager
 
+    FullEnvironmentName:
```
acdha: I'm thinking we could migrate to using these values everywhere but it's not a priority.

rstorey: Yeah, I was thinking that too. The full names are nicer.

PR #491, opened 2018-10-23
Update hero image and links in navs

elainekamlley:
Switched out hero image and updated navs with correct History hub link


elainekamlley:
TODO: Change out help center image

PR #493, opened 2018-10-23
Set SECURE_PROXY_SSL_HEADER

rstorey:


rstorey:
All 4 use settings_prod, so it will

PR #530, opened 2018-10-26
Update CloudFlare SecurityGroup update script

acdha:
This applies the list to every group tagged with `AllowCloudFlareIngress` and is closer to what we need to be able to run this using Lambda with scheduled events. Still TBD is tagging the security groups with this — I was wondering whether we want to make that conditional on the environment name but am also wondering whether we should just enable it for all of the environments so testing can be done using the CDN configuration as desired.

### DIFF
--- a/cloudformation/add_cloudflare_ips_to_sgs.py
+++ b/cloudformation/add_cloudflare_ips_to_sgs.py
@@ -1,32 +1,69 @@
 #!/usr/bin/env python3
+"""
+Ensure that every security group tagged with “AllowCloudFlareIngress” has
+permissions for every public CloudFlare netblock
+"""
+
+import sys
 
 import boto3
 import requests
 from botocore.exceptions import ClientError
 
-SECURITY_GROUP_ID = "sg-0cf16b045e14f5fad"
+EC2_CLIENT = boto3.client("ec2")
 
-ec2 = boto3.client("ec2")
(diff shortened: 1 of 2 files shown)


rstorey:
I tagged the load balancer security groups for all four environments. I also added the tag to the cloudformation template. Would you check that I applied those tags correctly?

acdha:
Hah, this hit an interesting bug in the GitHub UI: because we have the option set to allow reviewers to push commits to the same branch, when I tried to review your commit it gave me the “authors can't review their own pull requests” message.

rstorey:
Yeah, I wanted to add you as a reviewer too and couldn't. If it looks good to you, we can merge this. I can run the script manually today, let's talk about how we want to set it up to run on a scheduled basis.

acdha:
Hmmm, this didn't do what I expected:

screenshot 2018-10-29 13 20 43

acdha:
I'm thinking we just change it to `AllowCloudFlareIngress=true`

rstorey:
So the tag name would be AllowCloudFlareIngress with a value of true?

acdha:
Yes — right now we don't care about the value since we don't have a reason where you'd want that tag to exist without including that SG in the process but if we come up with a scenario where that's not true we could add a truthiness check.

rstorey:
I updated the tags on the four existing security groups.

rstorey:
Ran the script, groups are updated.

rstorey:
Also see https://github.com/LibraryOfCongress/concordia/issues/534

Issue #515, opened 2018-10-24
Address Internet Explorer Compatibility 


elainekamlley:
Decide on course of action to make crowd.loc.gov compatible in IE

acdha:
We should identify specific areas and prioritize them. Much of the site is functional in IE11 but the homepage has some custom CSS which needs updating — basically anything in https://github.com/LibraryOfCongress/concordia/blob/master/concordia/static/css/base.css which mentions ”flex” needs reviewing since without something like autoprefixer those don't have the fallbacks which the Bootstrap classes include.

colefemeno:
As a user coming to the site using IE, I would like to be notified that my experience might not be optimal with this browser. I could see this message in a banner or modal when I land on the page.

acdha:
@colefemeno Do we want a simple modal or something like the https://browser-update.org/ ribbon?

rstorey:
See https://github.com/LibraryOfCongress/concordia/pull/539#issuecomment-434758192

rstorey:
If we want to do more with IE compatibility, let's ticket specific issues separately.

PR #539, opened 2018-10-29
Banner for IE 11 and older

rstorey:
This is a first draft at a banner and message for IE11 and older users. Refs #515

### DIFF
--- a/concordia/static/js/base.js
+++ b/concordia/static/js/base.js
@@ -68,6 +68,29 @@ function displayMessage(level, message, uniqueId) {
     $messages.append($newMessage);
 }
 
+function detectOldIE() {
+    /* 
+        Returns true if user is running IE 11 or older. 
+        Returns false if user is running Edge or non-IE browser.
+    */
+
+    var ua = window.navigator.userAgent;
+    var msie = ua.indexOf('MSIE ');
+    var trident = ua.indexOf('Trident/');
+
+    if (msie > 0 || trident > 0) {
+        return true;
+    } else {
+        return false;
+    }
+}
+
+$(function() {
+    if(detectOldIE()) {
+        displayMessage("danger", "Internet Explorer 11 and older are not fully supported by this website. Please use a different browser such as Chrome or Firefox.")
+    }    
+});
+
 if (screenfull.enabled) {
     $('#go-fullscreen')
         .removeAttr('hidden')
--- a/concordia/static/js/base.js
+++ b/concordia/static/js/base.js
@@ -86,9 +86,13 @@ function detectOldIE() {
 }
 
 $(function() {
-    if(detectOldIE()) {
-        displayMessage("danger", "Internet Explorer 11 and older are not fully supported by this website. Please use a different browser such as Chrome or Firefox.")
-    }    
+    if (detectOldIE()) {
+        displayMessage(
+            'danger',
+            'Internet Explorer 11 and older are not fully supported by this website. ' +
+                'Please use a different browser such as Chrome or Firefox.'
+        );
+    }
 });
 
 if (screenfull.enabled) {


concordia/static/js/base.js
```diff
@@ -68,6 +68,33 @@ function displayMessage(level, message, uniqueId) {
...
+        return false;
+    }
+}
+
+$(function() {
+    if (detectOldIE()) {
+        displayMessage(
+            'danger',
+            'Internet Explorer 11 and older are not fully supported by this website. ' +
+                'Please use a different browser such as Chrome or Firefox.'
```
acdha: I'm tempted to change this to something like this:

> This website fully supports the current  version of every major browser (Microsoft Edge, Google Chrome, Mozilla Firefox, and Apple Safari). See <a href="https://crowd.loc.gov/help-center/#headingTwelve">our browser support policy</a> for more information and ongoing development efforts.

acdha: Basically the change I'd most want to make would be not singling out a single browser and giving them a link for more information.

rstorey: HTML in the message isn't working (without further changes) but I changed the wording so it directs them where to go.

concordia/static/js/base.js
```diff
@@ -68,29 +68,21 @@ function displayMessage(level, message, uniqueId) {
...
-    if (detectOldIE()) {
+    if (isOutdatedBrowser()) {
         displayMessage(
             'danger',
-            'Internet Explorer 11 and older are not fully supported by this website. ' +
-                'Please use a different browser such as Chrome or Firefox.'
+            'You are using an outdated browser. This website fully supports the current ' +
+                'version of every major browser ' +
+                '(Microsoft Edge, Google Chrome, Mozilla Firefox, and Apple Safari). See ' +
+                ' our browser support policy in the Help Center for more information.'
```
acdha: I'd like this to be a link but `displayMessage()` uses text for safety:

https://github.com/LibraryOfCongress/concordia/blob/master/concordia/static/js/base.js#L66

I'm tempted to just make this do it directly:

```javascript
$('#messages').find('#message-template .alert')
        .clone()
        .removeAttr('hidden id')
        .addClass('alert-danger')
        .prepend(theMessage)
        .appendTo("#messages");
```

rstorey: Ok, I want the link too. I'll update it to do it this way.

concordia/static/js/base.js
```diff
@@ -68,6 +68,44 @@ function displayMessage(level, message, uniqueId) {
...
+        $('#' + uniqueId).remove();
+        $newMessage.attr('id', uniqueId);
+    }
+
+    $newMessage.prepend(message);
+
+    $messages.append($newMessage);
+}
+
+function displayMessage(level, message, uniqueId, isText = true) {
```
acdha: Unfortunately, this won't work in IE since it doesn't support ES5:

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/default_parameters

concordia/static/js/base.js
acdha: Use this snippet: 

```javascript
$("#messages")
  .removeAttr("hidden")
  .find("#message-template .alert")
  .clone()
  .removeAttr("hidden id")
  .addClass("alert-danger")
  .prepend(theMessage)
  .appendTo("#messages");
```

concordia/static/js/base.js
```diff
@@ -68,6 +68,44 @@ function displayMessage(level, message, uniqueId) {
     $messages.append($newMessage);
 }
 
+function displayHTMLMessage(level, message, uniqueId) {
```
acdha: This is more duplication than I'd like. The way I would probably do this is to rely on the fact that [`jQuery.prepend()`](http://api.jquery.com/prepend/) accepts both HTML strings and elements and refactor this so we have a `displayHtmlMessage()` function which is what we have now except without the `createTextNode` call and then make `displayMessage()` basically just be `function displayNewMessage(level, message, uniqueId) { return displayHtmlMessage(level, document.createTextNode(message), uniqueId); }` so most calls will safely escape HTML but you can call `displayHtmlMessage()` directly when you know you have safe content.

rstorey:
So I tested this in Edge and IE11. Unfortunately, this bit of code (note the addition of the not operator in the return statement) returns true for Edge.
https://github.com/LibraryOfCongress/concordia/blob/master/concordia/static/js/base.js#L75-L80

acdha:
Does `CSS.supports("display", "flex")` work correctly in Edge?

acdha:
Ah, it's an Edge bug: they only support the two-argument form:

image

I'll file MDN & CanIUse.com bug reports so the compatibility lists can be updated.

PR #543, opened 2018-10-29
Update progress bars

acdha:
This updates the progress bar display to be reusable on pages other than the item detail view


acdha:
Funny you should mention that 😉

acdha:
With that shuffle, here's what it looks like:
image
image
image

rstorey:
I like this better. I guess what I meant when I said that the in progress caption shouldn't be displayed if it's zero, was not for the case when everything is zero, but when complete is 100%. If we remove the In progress caption there, then perhaps the complete caption won't wrap like it's doing and it'll look cleaner. I think when everything is zero, it makes sense to show all the zeroes.

acdha:
Oh, yeah, we can just tweak the CSS to prevent wrapping which I find preferable since there's some variability across browsers and operating systems when you're trying to fit things too tightly.

acdha:
With the latest commits the Campaign, Project, and Item detail views all share the same code for populating the context variables.

image
image
image

acdha:
@rstorey @elainekamlley Currently the progress bars display above the related links:

image

I'm wondering whether that would be better with the title full width and the related links and stats in the same side column:

image

rstorey:
I like the second option with the full width title better.

acdha:
I'm wondering whether we want the related links above or below

acdha:
Current screenshot:

image

acdha:
Here's a combined version with #554:

image

acdha:
This is now in #554 since I had to work on similar parts of the same page.

PR #557, opened 2018-10-31
Update public Sentry DSN

acdha:
Now that we have a stable HTTPS URL we can deploy it:

https://48ec47f15e484502a29879e40ed2e0c3@crowd-sentry.loc.gov/3

### DIFF
--- a/cloudformation/infrastructure/fargate-cluster.yaml
+++ b/cloudformation/infrastructure/fargate-cluster.yaml
@@ -1,6 +1,6 @@
 Description: >
     This template deploys a fargate cluster to the provided VPC and subnets
-    
+
 Parameters:
 
     EnvironmentName:
@@ -18,7 +18,7 @@ Parameters:
     SecurityGroup:
         Description: Select the Security Group to use for the ECS cluster hosts
         Type: AWS::EC2::SecurityGroup::Id
-    
+
     LoadBalancerSecurityGroup:
         Description: The SecurityGroup for load balancer
         Type: AWS::EC2::SecurityGroup::Id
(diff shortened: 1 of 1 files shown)


cloudformation/infrastructure/fargate-cluster.yaml
```diff
@@ -80,273 +79,271 @@ Parameters:
...
+                        Value: !Ref FullEnvironmentName
+                      - Name: S3_BUCKET_NAME
+                        Value: !Ref S3BucketName
+                      - Name: CELERY_BROKER_URL
+                        Value: pyamqp://guest@localhost:5672
+                      - Name: AWS_DEFAULT_REGION
+                        Value: !Ref AWS::Region
+                      - Name: SENTRY_BACKEND_DSN
+                        Value: http://34db819263f34c28809da045f841f045@sentry-internal.devops.cloud/2
+                      #              - Name: SENTRY_FRONTEND_DSN
```
rstorey: Need to add it here, too.

rstorey:
Other than adding the env var to both containers (importer and app) the only issue I have with this is that because the whole file is reformatted, it needs to be run through CloudFormation to test it. Not really something I was going to put in the test plan for this release, particularly since we have our stacks set up and we made a decision not to deploy stacks without defined canonical host names and certificates.

cloudformation/infrastructure/fargate-cluster.yaml
acdha: I just pushed an update after disabling auto-formatting. I'm not sure `SENTRY_FRONTEND_DSN` being set is relevant to the importer currently but in case that changes in the future it's now in both places.

PR #558, opened 2018-10-31
Use meaningful names in the help center

### LINKED ISSUE #546: Use meaningful names in the help center
Currently https://crowd.loc.gov/help-center/ has anchors with names like “headingTwelve” which aren't meaningful and could change in the future. It would be nice if we could make durable links such as https://crowd.loc.gov/help-center/#browser-support-policy.

Additionally, we might want to review the JavaScript to have it automatically expand the accordion when the URL fragment matches one of those anchors.

rstorey:
Also expands the accordion slide for the section if there is a hash in the location.

Closes #546

### DIFF
--- a/concordia/static/js/base.js
+++ b/concordia/static/js/base.js
@@ -85,7 +85,7 @@ $(function() {
             'You are using an outdated browser. This website fully supports the current ' +
             'version of every major browser ' +
             '(Microsoft Edge, Google Chrome, Mozilla Firefox, and Apple Safari). See ' +
-            'our <a href="/help-center/#headingTwelve">browser support policy</a> ' +
+            'our <a href="/help-center/#browserSupport">browser support policy</a> ' +
             'for more information.';
 
         displayHtmlMessage('danger', theMessage);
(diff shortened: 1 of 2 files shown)


concordia/static/js/base.js
```diff
@@ -85,11 +85,15 @@ $(function() {
...
             'version of every major browser ' +
             '(Microsoft Edge, Google Chrome, Mozilla Firefox, and Apple Safari). See ' +
-            'our <a href="/help-center/#headingTwelve">browser support policy</a> ' +
+            'our <a href="/help-center/#browserSupport">browser support policy</a> ' +
             'for more information.';
 
         displayHtmlMessage('danger', theMessage);
     }
+
+    if (location.hash && $('#faqAccordion')) {
```
acdha: I think `$('#faqAccordion')` will always evaluate to true. This could be a simple length check:

```suggestion
    if (location.hash && $('#faqAccordion').length > 0) {
```

Alternatively, you could just use jQuery to scope it implicitly since the only elements we care about should all be children of the accordion element anyway: 

```javascript
if (location.hash) {
    $('#faqAccordion').find(location.hash).collapse('show');
}
```

rstorey: The second option didn't work so I used the first option. Tested in the debugger.

PR #560, opened 2018-10-31
show a message in IE11 but not in Edge

rstorey:
Refs #515

### DIFF
--- a/concordia/static/js/base.js
+++ b/concordia/static/js/base.js
@@ -73,10 +73,11 @@ function displayMessage(level, message, uniqueId) {
 }
 
 function isOutdatedBrowser() {
+    /* See https://caniuse.com/#feat=css-supports-api */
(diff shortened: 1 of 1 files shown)


acdha:
The second check was intentional to catch any other browsers which we may not have tested but we can omit it for now since all of the major ones do have this.

concordia/static/js/base.js
```diff
@@ -73,10 +73,11 @@ function displayMessage(level, message, uniqueId) {
 }
 
 function isOutdatedBrowser() {
+    /* See https://caniuse.com/#feat=css-supports-api */
     if (typeof CSS == 'undefined' || !CSS.supports) {
         return true;
     }
-    return !CSS.supports('display: flex');
+    return false;
```
acdha: Since we're changing this to use only a single check we could replace the function body with a simple `return (typeof CSS == 'undefined' || !CSS.supports);`

PR #561, opened 2018-10-31
Commands I used for DB dump and restore

rstorey:
Wanna keep track of these so I don't have to remember the commands next time.


postgresql/restore.sh
```diff
@@ -0,0 +1,18 @@
...
+# Before running the restore, you'll have to stop the ECS task to close any open connections.
+# Then, run the following to drop the database.
+# psql -U concordia -h $POSTGRES_HOST
+# \c postgres
+# drop database concordia;
+
+export POSTGRES_HOST=
+export DUMP_FILE=concordia.dmp
+
+pg_restore --create --clean -U concordia -h $POSTGRES_HOST -Fc --dbname=postgres --no-owner --no-acl $DUMP_FILE
```
acdha: This may be version specific but with 10.5 this command failed and told me to use `psql` instead for a plain text file.

rstorey: Yeah this command is for a dump created with the -Fc flag.

acdha: Oh, that's not the dump I pulled from S3?

rstorey: No, I used the one created by the command in `dump.sh` after my attempt to use the plain text SQL to reload failed.

PR #573, opened 2018-11-01
Login & registration ratelimit

rstorey:
Besides customizing the view for ratelimit exceptions, the login rate limit behavior isn't working quite as expected. It seems that `request.user.is_authenticated` does not get populated when `login_rate` evaluates it - it gets set to True later on in the request processing. I guess I'll have to customize the login form processing to check the rate limit after processing the form rather than before?

### DIFF
--- a/concordia/settings_template.py
+++ b/concordia/settings_template.py
@@ -95,7 +95,10 @@ MIDDLEWARE = [
     "django.contrib.messages.middleware.MessageMiddleware",
     "django.middleware.clickjacking.XFrameOptionsMiddleware",
     "maintenance_mode.middleware.MaintenanceModeMiddleware",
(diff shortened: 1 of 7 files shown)


rstorey:
I guess I thought we only wanted to rate limit failed login attempts, not all login attempts...However, maybe it's sufficient to rate limit all login attempts.

rstorey:
Guess I need to disable rate limiting for settings_test

acdha:
I just pushed c6d3444 to disable rate-limiting globally for testing. I was wondering whether it might make sense to only do so for the tests which break but there's (I think) an equally valid argument that we should make the rate-limit values into settings so we could raise them for testing.

Issue #541, opened 2018-10-29
Update developer docs and add pre-commit webhook


rstorey:
Some of the developer guide needs to be updated accurate steps and information on how to install and use pre-commit.

It would also be nice if recommended VS Code plugins and settings could be documented in our dev guide as well.

jkueloc:
@rstory @acdha @rabiloc - Do you have specific VS Code plugins and settings that you would like to recommend in addition to the Python extension.

acdha:
Most of them follow the same tools we're using in our Pre-Commit configuration. The most important one is the Editor Config extension, which causes VSCode to use the `.editorconfig` settings in the repo, and the Prettier one is handy for ensuring that you don't have to wait for pre-commit to run to reformat JavaScript, CSS, etc.

```
amazonwebservices.aws-toolkit-vscode
dbaeumer.vscode-eslint
shinnn.stylelint
EditorConfig.EditorConfig
esbenp.prettier-vscode
mrmlnc.vscode-scss
```

PR #577, opened 2018-11-01
Updated broken links and how to get started with development

rstorey:
Updated developer guide and contributing. Note that we still need to do something about #541


docs/for-developers.md
```diff
@@ -48,14 +47,14 @@ development server:
...
+Once the database, rabbitMQ service, importer and the application
+are running, you're ready to import data.
+First, [create a Django admin user](https://docs.djangoproject.com/en/2.1/intro/tutorial02/#creating-an-admin-user)
 and log in as that user.
-Then, Click on Campaigns and then Add Campaign. Enter a name for the campaign 
-and project and a loc.gov item URL such as 
-[https://www.loc.gov/item/mss859430011](https://www.loc.gov/item/mss859430011) and click Create.
-
+Then, go to the Admin area (under Account) and click "Bulk Import Items".
+Upload a spreadsheet populated according to the instructions. Once all the import
```
acdha: Since we don't have a source for that spreadsheet, should we have a simple test one? That might be handy for anyone who doesn't want to download tens of thousands of images, too.

rstorey: Yes, I had the same thought. A small test spreadsheet is a good idea. Populating any test content without a spreadsheet is pretty tedious.

PR #578, opened 2018-11-01
Media url slash

rstorey:


acdha:
We should at some point turn this into a template tag which calls urljoin so this class of problem isn't an issue

PR #579, opened 2018-11-02
Improved IE11 compatibility

acdha:
This improves the IE11 experience but raises the question of how we want to handle compatibility in the future: do we want to adopt a heavyweight tool like Webpack, Rollup, etc. or add some light-weight tooling so e.g. we process CSS with PostCSS + autoprefixer, JS with Babel, etc. without changing the entire project structure, and whether we serve the same bundle to everyone or take the polyfill.io approach of only serving support code to the older browsers which need it.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -13,3 +13,11 @@ repos:
       hooks:
           - id: prettier
             files: \.(css|less|scss|ts|tsx|graphql|gql|json|js|jsx|md)$
+    - repo: https://github.com/pre-commit/mirrors-eslint
+      rev: v5.8.0
+      hooks:
+          - id: eslint
+            additional_dependencies:
+                - eslint@5.8.0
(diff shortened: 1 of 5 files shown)


acdha:
Here's the homepage in IE11 currently:

screenshot 2018-11-02 10 41 30

acdha:
Campaign listing:

screen shot 2018-11-02 at 10 52 00 am

Campaign detail:

screen shot 2018-11-02 at 10 52 59 am

acdha:
Project detail:

screen shot 2018-11-02 at 11 56 46 am

acdha:
Item detail:

screen shot 2018-11-02 at 11 57 53 am

acdha:
Asset detail:

screen shot 2018-11-02 at 12 20 16 pm

Asset detail fullscreen:

screen shot 2018-11-02 at 12 20 42 pm

PR #586, opened 2018-11-02
Proper CSV export at campaign level

rstorey:
FIrst of a few PRs from this branch for export functions.

### DIFF
--- a/exporter/views.py
+++ b/exporter/views.py
@@ -21,16 +21,6 @@ class ExportCampaignToCSV(TemplateView):
 
     @method_decorator(login_required)
     def get(self, request, *args, **kwargs):
-        headers = [
-            "Campaign",
-            "Project",
-            "Item",
-            "ItemId",
-            "Asset",
-            "AssetStatus",
-            "DownloadUrl",
-            "Transcription",
-        ]
         qs = Campaign.objects.filter(slug=self.kwargs["campaign_slug"]).select_related()
         headers, data = flatten_queryset(
             qs,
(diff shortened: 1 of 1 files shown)


exporter/views.py
```diff
@@ -8,57 +7,48 @@
...
     def get(self, request, *args, **kwargs):
-        campaign = Campaign.objects.get(slug=self.kwargs["campaign_slug"])
-        asset_list = Asset.objects.filter(item__project__campaign=campaign).order_by(
-            "title", "sequence"
-        )
-        # Create the HttpResponse object with the appropriate CSV header.
-        response = HttpResponse(content_type="text/csv")
-        response["Content-Disposition"] = 'attachment; filename="{0}.csv"'.format(
-            campaign.slug
+        qs = Campaign.objects.filter(slug=self.kwargs["campaign_slug"]).select_related()
```
acdha: I generally like to start with the class I'm returning (e.g. `Asset` in this case) but I don't think that matters here since we're returning the whole table anyway.

acdha: Don't we need to filter for the most recent transcription so we don't return multiple rows for every revision? Since we're using Django 2, I think that's just a [`Subquery`](https://docs.djangoproject.com/en/2.1/ref/models/expressions/#subquery-expressions) annotation.

exporter/views.py
```diff
@@ -8,57 +7,48 @@
...
+            qs,
+            field_names=[
+                "title",
+                "project__title",
+                "project__item__title",
+                "project__item__item_id",
+                "project__item__asset__title",
+                "project__item__asset__transcription_status",
+                "project__item__asset__download_url",
+                "project__item__asset__transcription__text",
```
acdha: According to https://support.office.com/en-us/article/excel-specifications-and-limits-1672b34d-7043-467e-8e27-269d656771c3 Excel can't handle more than 32K characters in a table cell. Currently the largest transcription appears to have 17K characters so it's not inconceivable that we'd approach that limit. Do we care?

rstorey: I think so. I believe we should detect when a transcription has to be truncated to fit into a CSV and report a warning or error if so.

exporter/views.py
rstorey: Yes, I was trying to figure out how to do that

exporter/views.py
acdha: It would only be an issue when using Excel rather than other tools so perhaps we could just do a single max length query and if it's near 32K call messages.warning with a note to use caution with Excel?

PR #589, opened 2018-11-02
OpenSeadragon custom controls

### LINKED ISSUE #582: OpenSeadragon tweaks
Make the controls always visible
Make the controls look like the rest of the site

acdha:
* [x] Use custom buttons for all of the OpenSeadragon controls with colors & icons matching the rest of the site
* [ ] Add keyboard shortcuts for controls and make them discoverable

Closes #582

### DIFF
--- a/concordia/static/js/contribute.js
+++ b/concordia/static/js/contribute.js
@@ -12,6 +12,36 @@ function unlockControls($container) {
     $container.find('button').removeAttr('disabled');
 }
 
+$(document).on('keydown', function(evt) {
+    /*
+        Global keyboard event handlers
+
+        * F1 and ? open help
+        * Control-I focuses on the image viewer
+        * Control-T focuses on the transcription text field
+
+        n.b. jQuery interferes with setting the focus so our handlers use the
+        DOM directly
+    */
+
+    if (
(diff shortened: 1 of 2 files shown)


acdha:
All of the custom controls are implemented but it looks like something in OpenSeadragon is preventing the normal behaviour enabling and disabling buttons which aren't applicable (e.g. disabling zoom out when you're already zoomed out) so more work will be needed to update the UI.

This also omits the next/previous page buttons since we are using single-image sources.

rstorey:
I'd like to merge this before the end of the week and separately ticket any work or issues that remain.

acdha:
That seems reasonable. I was looking at the Seadragon source and it appears to be non-trivial to adjust the zoom button state because there doesn't appear to be a way to get the maximum supported zoom level out of the viewer.

We can move the keyboard support into a separate ticket. The main issue we need to decide is how to handle switching keyboard focus between the viewer & transcription view since the existing shortcuts are also valid text controls. I'm thinking we might end up with something like a toggle focus command for people who find the mouse inconvenient and a help dialog bound to F1/? and a help button which shows you that and the list of options once it has focus.

PR #595, opened 2018-11-06
Add coverage run command

rstorey:
One more thing needed to get coveralls to actually work.


rstorey:
Also adds exporter to tests being executed in travis.

Issue #169, opened 2018-09-14
Update Exporter config to include original image file name


elainekamlley:
To make ETL simple, the export of the transcription data needs to match the original format of the image name and directory name.

PR #597, opened 2018-11-06
WIP: Exporter

### LINKED ISSUE #580: Refactor exporter - look at past tickets and start prelim tests

rstorey:
Closes #580 
Refs #169


rstorey:
I haven't gotten this going to S3 yet, but there are unit test fixes in here.

concordia/tests/utils.py
```diff
@@ -104,6 +104,11 @@ def create_asset(
     return asset
 
 
+def clean_up():
+    campaign = Campaign.objects.get(title="Test Campaign")
```
acdha: Is this actually necessary? Django's test case normally resets the database between runs

exporter/tests/test_view.py
```diff
@@ -37,67 +65,45 @@ def login_user(self):
...
 
         response = self.client.get(
-            reverse(
-                "transcriptions:export-csv", args=(asset.item.project.campaign.slug,)
-            )
+            reverse("transcriptions:export-csv", args=(campaign_slug,))
         )
 
         self.assertEqual(response.status_code, 200)
+        response_content = ""
```
acdha: Couldn't this be `"".join(response.streaming_content)`?

exporter/views.py
```diff
@@ -10,8 +11,33 @@
...
+
+
+def get_original_asset_id(download_url):
+    """
+    Extract the bit from the download url
+    that identifies this image uniquely on loc.gov
+    """
+    if download_url.startswith("http://tile.loc.gov/"):
+        pattern = r"/service:([A-Za-z0-9:]*)/"
+        asset_id = re.search(pattern, download_url)
```
acdha: Shouldn't we handle the case where this fails with at least an `assert asset_id`?

concordia/tests/utils.py
rstorey: Well, without the clean up it fails because it tries to run setUp() twice and fails because stuff already exists.

exporter/views.py
rstorey: Yes, this part is still under development. I need to check all the download_urls for all the assets to see which ones if any did not come from the tile server using this URL convention.

concordia/tests/utils.py
acdha: That sounds like we might want to use `setUpTestData()` instead: https://docs.djangoproject.com/en/2.1/topics/testing/tools/#django.test.TestCase

exporter/tests/test_view.py
rstorey: I don't think so. I tried that and this is what I got:
```
Traceback (most recent call last):
  File "/Users/rstorey/workspace/concordia/exporter/tests/test_view.py", line 67, in test_csv_export
    response_content = "".join(response.streaming_content)
TypeError: sequence item 0: expected str instance, bytes found
```

acdha: Add the `b` prefix to mark it as bytes (`b"".join(…)`) or `"".join(map(str, response.streaming_content))` but I normally prefer to keep bytes as bytes unless we're explicitly decoding them into text.

rstorey: Trying the `b"".join()` method doesn't work because you can't mix bytes and string literals.

rstorey:
So I checked and all the download URLs we have for assets we imported are of the form that getting the unique asset ID should work. I added the assert also.

PR #601, opened 2018-11-07
Fix docker container build for pylibmc dependency

rstorey:
Docker containers won't build since the addition of pylibmc. Adding pre-req for ubuntu.

### DIFF
--- a/concordia/templates/admin/bulk_import.html
+++ b/concordia/templates/admin/bulk_import.html
@@ -79,7 +79,7 @@
             <h4>Messages</h4>
             <ul>
                 {% for message in messages %}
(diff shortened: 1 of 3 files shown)


concordia/templates/admin/bulk_import.html
```diff
@@ -79,7 +79,7 @@ <h2>Import Tasks</h2>
             <h4>Messages</h4>
             <ul>
                 {% for message in messages %}
-                    <li class="message {% if message.level >= DEFAULT_MESSAGE_LEVELS.error %}message-error{% elif message.level >= DEFAULT_MESSAGE_LEVELS.warning %}message-warning{% endif %}">{{ message }}</li>
```
acdha: This seems unrelated?

rstorey: It is, except it was cluttering up my console output.

PR #604, opened 2018-11-08
Export: upload bags to S3

### LINKED ISSUE #169: Update Exporter config to include original image file name

rstorey:
- only export completed transcriptions in BagIt format, all transcriptions in CSV format
- upload created bags to S3 bucket
- created bags uploaded to S3 are named with campaign + timestamp
- filenames in the bag reflect the original download URL

Closes #169

### DIFF
--- a/cloudformation/infrastructure/fargate-cluster.yaml
+++ b/cloudformation/infrastructure/fargate-cluster.yaml
@@ -79,6 +79,11 @@ Parameters:
         Type: String
         Description: name of the S3 bucket (public) where collection images will be stored
 
+    ExportS3BucketName:
+        Type: String
+        Description: name of the S3 bucket (public) where exported transcriptions will be stored
+
+
 Resources:
 
   ConcordiaServiceSecretAccessPolicy:
@@ -244,6 +249,8 @@ Resources:
                 Value: !Ref FullEnvironmentName
               - Name: S3_BUCKET_NAME
                 Value: !Ref S3BucketName
+              - Name: EXPORT_S3_BUCKET_NAME
+                Value: !Ref ExportS3BucketName
               - Name: CELERY_BROKER_URL
                 Value: pyamqp://guest@localhost:5672
               - Name: AWS_DEFAULT_REGION
@@ -300,6 +307,8 @@ Resources:
                 Value: !Ref FullEnvironmentName
               - Name: S3_BUCKET_NAME
                 Value: !Ref S3BucketName
+              - Name: EXPORT_S3_BUCKET_NAME
+                Value: !Ref ExportS3BucketName
               - Name: CELERY_BROKER_URL
                 Value: pyamqp://guest@localhost:5672
               - Name: AWS_DEFAULT_REGION
--- a/concordia/settings_dev.py
+++ b/concordia/settings_dev.py
@@ -20,6 +20,7 @@ CELERY_BROKER_URL = "pyamqp://guest@localhost"
 CELERY_RESULT_BACKEND = "rpc://"
 
 S3_BUCKET_NAME = "crowd-dev-content"
+EXPORT_S3_BUCKET_NAME = "crowd-dev-export"
 
 DEFAULT_FILE_STORAGE = "storages.backends.s3boto3.S3Boto3Storage"
 AWS_STORAGE_BUCKET_NAME = S3_BUCKET_NAME
--- a/concordia/settings_dev.py
+++ b/concordia/settings_dev.py
@@ -19,7 +19,7 @@ ALLOWED_HOSTS = ["127.0.0.1", "0.0.0.0", "*"]
 CELERY_BROKER_URL = "pyamqp://guest@localhost"
 CELERY_RESULT_BACKEND = "rpc://"
 
-S3_BUCKET_NAME = "concordia-staticpages"
+S3_BUCKET_NAME = "crowd-dev-content"
 
 DEFAULT_FILE_STORAGE = "storages.backends.s3boto3.S3Boto3Storage"
 AWS_STORAGE_BUCKET_NAME = S3_BUCKET_NAME
--- a/concordia/settings_dev.py
+++ b/concordia/settings_dev.py
@@ -19,15 +19,6 @@ ALLOWED_HOSTS = ["127.0.0.1", "0.0.0.0", "*"]
 CELERY_BROKER_URL = "pyamqp://guest@localhost"
 CELERY_RESULT_BACKEND = "rpc://"
 
-S3_BUCKET_NAME = "crowd-dev-content"
-EXPORT_S3_BUCKET_NAME = "crowd-dev-export"
-
-DEFAULT_FILE_STORAGE = "storages.backends.s3boto3.S3Boto3Storage"
-AWS_STORAGE_BUCKET_NAME = S3_BUCKET_NAME
-AWS_DEFAULT_ACL = None  # Don't set an ACL on the files, inherit the bucket ACLs
-
-MEDIA_URL = "https://%s.s3.amazonaws.com/" % S3_BUCKET_NAME
-
 EMAIL_BACKEND = "django.core.mail.backends.console.EmailBackend"
 EMAIL_FILE_PATH = "/tmp/concordia-messages"  # change this to a proper location
 DEFAULT_FROM_EMAIL = os.environ.get("DEFAULT_FROM_EMAIL", "")
--- a/exporter/views.py
+++ b/exporter/views.py
@@ -1,3 +1,4 @@
+from datetime import datetime
 import os
 import re
 import shutil
@@ -128,13 +129,21 @@ class ExportCampaignToBagit(TemplateView):
         # Download zip
         with open("%s.zip" % export_base_dir, "rb") as zip_file:
             response = HttpResponse(zip_file, content_type="application/zip")
-        response["Content-Disposition"] = "attachment; filename=%s.zip" % campaign_slug
+
+        export_filename = "%s-%s.zip" % (
+            campaign_slug,
+            datetime.today().isoformat(timespec="minutes"),
+        )
+
+        response["Content-Disposition"] = "attachment; filename=%s" % export_filename
 
         # Upload zip to S3 bucket
-        if "S3_BUCKET_NAME" in settings and settings.S3_BUCKET_NAME:
+        s3_bucket = getattr(settings, "S3_BUCKET_NAME", None)
+
+        if s3_bucket:
             s3 = boto3.resource("s3")
-            s3.Bucket(settings.S3_BUCKET_NAME).upload_file(
-                "%s.zip" % export_base_dir, "exporter/%s.zip" % campaign_slug
+            s3.Bucket(s3_bucket).upload_file(
+                "%s.zip" % export_base_dir, "exporter/%s" % export_filename
             )
 
         # Clean up temp folders & zipfile once exported
--- a/exporter/views.py
+++ b/exporter/views.py
@@ -34,7 +34,7 @@ def get_original_asset_id(download_url):
     that identifies this image uniquely on loc.gov
     """
     if download_url.startswith("http://tile.loc.gov/"):
-        pattern = r"/service:([^/]?)/"
+        pattern = r"/service:([A-Za-z0-9:/-]?)/"
         asset_id = re.search(pattern, download_url)
         assert asset_id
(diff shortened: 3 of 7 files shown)


concordia/settings_dev.py
```diff
@@ -19,7 +19,7 @@
 CELERY_BROKER_URL = "pyamqp://guest@localhost"
 CELERY_RESULT_BACKEND = "rpc://"
 
-S3_BUCKET_NAME = "concordia-staticpages"
+S3_BUCKET_NAME = "crowd-dev-content"
```
acdha: I'm a bit tired of having to manage these around the settings files. What do you think about changing `settings_template.py` to default to `S3_BUCKET_NAME = os.getenv("S3_BUCKET_NAME", f"crowd-{CONCORDIA_ENVIRONMENT}-content")` (after the environment name is pulled in, of course)?

I'm also wondering whether this is a good time to setup more than one storage backend with the goal towards eventually having bags, etc. available as public downloads which might not go through Django.

exporter/views.py
```diff
@@ -14,7 +16,7 @@
 from concordia.models import Asset, Transcription
 
 
-def get_latest_transcription_data(campaign_slug):
+def get_latest_transcription_data(campaign_slug, filter_status=False):
```
acdha: The name of this argument doesn't really tell me what it does and it seems likely that at some point we might want to have other filters. What do you think about changing this signature to be the asset queryset:

```suggestion
def get_latest_transcription_data(campaign_slug, asset_qs=None):
```

Then we could change the `Asset.objects` call below to use `asset_qs` and the caller could pass in an asset QuerySet with arbitrary filtering, ordering, select/prefetch related, etc.

exporter/views.py
```diff
@@ -33,7 +41,7 @@ def get_original_asset_id(download_url):
     that identifies this image uniquely on loc.gov
     """
     if download_url.startswith("http://tile.loc.gov/"):
-        pattern = r"/service:([A-Za-z0-9:]*)/"
+        pattern = r"/service:([A-Za-z0-9:\-]*)/"
```
acdha: Looking at the code for the other side of this, I see this comment:

> The API places no restrictions on the form of the identifiers that a server may use or support. All special characters (e.g. ? or #) must be URI encoded to avoid unpredictable client behaviors. The URI syntax relies upon slash (/) separators so any slashes in the identifier must be URI encoded (also called "percent encoded").

That is basically passed in to this code:

```python
    identifier = urllib.unquote_plus(raw_id).replace(':', '/')
```

Since that's using [Flask's `path` pattern](http://exploreflask.com/en/latest/views.html#built-in-converters) I'm interpreting that as potentially accepting any character which a POSIX filesystem will accept.

exporter/views.py
```diff
@@ -92,9 +100,10 @@ class ExportCampaignToBagit(TemplateView):
     @method_decorator(login_required)
     def get(self, request, *args, **kwargs):
         campaign_slug = self.kwargs["campaign_slug"]
-        assets = get_latest_transcription_data(campaign_slug)
+        assets = get_latest_transcription_data(campaign_slug, True)
```
acdha: I would really want this to name the keyword argument for clarity. I now prefer using the newer Python 3 syntax for the function declaration (`def foo(bar, *, baaz):`) which makes it a syntax error to pass a keyword argument by position.

exporter/views.py
```diff
@@ -120,9 +129,22 @@ def get(self, request, *args, **kwargs):
...
+        s3_bucket = getattr(settings, "S3_BUCKET_NAME", None)
+
+        if s3_bucket:
+            s3 = boto3.resource("s3")
+            s3.Bucket(s3_bucket).upload_file(
+                "%s.zip" % export_base_dir, "exporter/%s" % export_filename
+            )
 
         # Clean up temp folders & zipfile once exported
         shutil.rmtree(export_base_dir)
```
acdha: We need to handle the case where an error prevents this final cleanup step from running. That could either be a try-finally block or using the context-manager form of [`TemporaryDirectory`](https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryDirectory) but it perhaps also says that we should move the the core of this into a separate function so it'd basically be something like:

```python
        campaign_slug = self.kwargs["campaign_slug"]
        assets = get_latest_transcription_data(campaign_slug, True)

        with TemporaryDirectory(prefix=campaign_slug) as export_base_dir:
            return do_the_export(…)

    def do_the_export(…):
        for asset in assets:
            …

        …
        
        if s3_bucket:
            s3 = boto3.resource("s3")
            s3.Bucket(s3_bucket).upload_file(
                "%s.zip" % export_base_dir, "exporter/%s" % export_filename
            )

        return response
```

exporter/views.py
```diff
@@ -120,9 +129,22 @@ def get(self, request, *args, **kwargs):
...
+        )
+
+        response["Content-Disposition"] = "attachment; filename=%s" % export_filename
 
         # Upload zip to S3 bucket
+        s3_bucket = getattr(settings, "S3_BUCKET_NAME", None)
+
+        if s3_bucket:
+            s3 = boto3.resource("s3")
+            s3.Bucket(s3_bucket).upload_file(
```
acdha: As far as I can tell, we're uploading the file but not using it. That raises the possibility of things like not setting the MIME type or security options on the file without noticing it. I'm wondering whether the process should basically be:

```python
try:
    if s3_bucket:
        # do upload
        # return redirect to the S3 object's public URL in the response returned by `upload_file`
    else:
    return response
finally:
    shutil.rmtree(export_base_dir)
```

exporter/views.py
```diff
@@ -92,9 +100,10 @@ class ExportCampaignToBagit(TemplateView):
     @method_decorator(login_required)
     def get(self, request, *args, **kwargs):
         campaign_slug = self.kwargs["campaign_slug"]
-        assets = get_latest_transcription_data(campaign_slug)
+        assets = get_latest_transcription_data(campaign_slug, True)
 
         export_base_dir = os.path.join(settings.SITE_ROOT_DIR, "tmp", campaign_slug)
```
acdha: The old code here was unsafe and we'd get some level of corruption if, say, someone double-clicked the link. I would use [`mkdtemp`](https://docs.python.org/3/library/tempfile.html#tempfile.mkdtemp) or the context-manager form of [`TemporaryDirectory`](https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryDirectory) with using a prefix, which almost certainly is not `SITE_ROOT_DIR`.

concordia/settings_dev.py
rstorey: Okay, I'm game to set up a separate S3 bucket for the exported content. I think it's also a good idea to figure out the S3 bucket names based on the environment. However the buckets are named dev test stage and prod, so it'd have to be based on ENV_NAME instead of CONCORDIA_ENVIRONMENT

rstorey: On second look, I don't think we need to change this particular setting. This is only for local runserver development using `settings_dev` where presumably no dedicated storage backend has been set up. In docker deployments, the S3 bucket name is passed as an env var to the docker container. I'll set up a separate bucket for exported content, but I think we should just stick with passing both the bucket names as env vars to docker like we do for other deployment-specific settings. If we want to programmatically generate the bucket names, the place to do that would be in the cloud formation template.

acdha: > where presumably no dedicated storage backend has been set up

Does that mean this should just be blank then so the `if s3_bucket` check will fall back to local storage?

rstorey: Yes, let's make it blank on the checked-in version.

exporter/views.py
rstorey: `shutil.rmtree` isn't necessary with `TemporaryDirectory` so it's even easier than this.

exporter/views.py
rstorey: Do you think `pattern = r"/service:([^/]?)/" ` would be too permissive?

cloudformation/infrastructure/fargate-cluster.yaml
```diff
@@ -244,6 +249,8 @@ Resources:
                 Value: !Ref FullEnvironmentName
               - Name: S3_BUCKET_NAME
                 Value: !Ref S3BucketName
+              - Name: EXPORT_S3_BUCKET_NAME
```
acdha: Seeing this name makes me want to refactor `S3_BUCKET_NAME` to describe the usage, too, but that's probably a separate iteration.

exporter/views.py
acdha: Should that be `r"/service:([^/]+)/"` to require one or more non-path separators? It seems reasonable to me but the main question I would have is how we'd know if that belief is incorrect — since this isn't in a public blocking path the big thing I'd focus on is getting a clear error report.

exporter/views.py
```diff
@@ -120,34 +125,37 @@ def get(self, request, *args, **kwargs):
...
-
-        if s3_bucket:
-            s3 = boto3.resource("s3")
-            s3.Bucket(s3_bucket).upload_file(
-                "%s.zip" % export_base_dir, "exporter/%s" % export_filename
+        try:
+            # Upload zip to S3 bucket
+            s3_bucket = getattr(settings, "EXPORT_S3_BUCKET_NAME", None)
+            export_filename = "%s-%s.zip" % (
+                campaign_slug,
```
acdha: It looks like we could refactor this not to need the `campaign_slug` variable to be passed in if this ZIP call simply used `os.path.basename(export_base_dir)`

exporter/views.py
```diff
@@ -120,34 +125,37 @@ def get(self, request, *args, **kwargs):
...
+                # Download zip
+                with open("%s.zip" % export_base_dir, "rb") as zip_file:
+                    response = HttpResponse(zip_file, content_type="application/zip")
+                response["Content-Disposition"] = (
+                    "attachment; filename=%s" % export_filename
+                )
+                return response
+        finally:
+            # Clean up temp folders & zipfile once exported
+            shutil.rmtree(export_base_dir)
```
acdha: Since this is called within a `TemporaryDirectory` context it should be cleaned up automatically without us needing to do anything here.

rstorey: It looks like the last round of comments wasn't on the latest version of the code?

acdha: They were current at the time I wrote them — not sure whether that overlapped with more changes

exporter/views.py
rstorey: But then the zip file would have the extra chars in it that are in the temporarily-created directory, wouldn't it?

rstorey: In my testing (I checked our current data for all the download URLs we have so far) the regex that I just committed works to get all the asset IDs. For whatever reason, if I change the * to a ? it fails the unit tests. I'm not really sure I care enough to figure out exactly why. The other regex I tried (allowing all characters except slash) also failed in the unit test.

exporter/views.py
rstorey: Weird cause I could've sworn I took out the try / finally before I went to the doctor.

acdha:
I guess my question is whether we should expose that S3 endpoint as the public download option, leaving this trigger behind a login gate or simply making it completely public and allowing anyone to trigger an export with some sort of throttling.

bagit-python should have S3 support but I definitely want time to do that right since it will solve a number of other requests like bagging to a different destination and potentially even streaming.

rstorey:
I think those are questions we need to get the CMs to weigh in on. I wouldn't have any problem letting the public create CSV exports, but for bags I think we should pre-create them and then maybe have a page where we list them for the public along with a little bit of data about them. In order to use the data effectively from the bag export, you kinda need the CSV as well.

exporter/views.py
```diff
@@ -47,7 +47,12 @@ class ExportCampaignToCSV(TemplateView):
     Exports the most recent transcription for each asset in a campaign
     """
 
-    @method_decorator(login_required)
+    @method_decorator(permission_required("concordia.add_campaign"))
+    @method_decorator(permission_required("concordia.change_campaign"))
```
acdha: Do we care about these permissions specifically or should we just use `staff_member_required`?

rstorey: I guess it could be simplified to `staff_member_required` - at this point they're all the same but that could change.

rstorey:
I want to get this in so I can test it with our latest production data. Please open tickets for anything else you think needs to be addressed. Thanks.

PR #620, opened 2018-11-13
WIP: improved asset detail UI

acdha:
* [ ] Add splitter handle from @jbresner 
* [ ] Add vertical layout for tall, narrow displays


acdha:
@jbresner This has the SVG file but it doesn't render as expected because the SVG file assumes that the FontAwesome 5 Pro font is available and that doesn't work even if I switch the font name to the free font we're using elsewhere:

image

I'm not sure whether there's a better way to use that font, embed the SVG versions of those icons, or simply draw two triangles.

jbresner:
@acdha - Sorry about that. I converted the Font Awesome font to an outline so it should render properly now.
[handle.svg.zip](https://github.com/LibraryOfCongress/concordia/files/2578042/handle.svg.zip)

acdha:
@jbresner Thanks – with the updated image it looks like this:

image

Might be worth merging the WIP so far, @rstorey?

Issue #441, opened 2018-10-19
Find a new page should exclude current page from set of potential pages


rstorey:
**How can we reproduce the bug?**
1. Import a collection and publish everything.
2. Make sure no one else is logged into the site.
3. Go to a random asset page in the collection.
4. Click Find a new page.
5. Click Find a new page again.

**What behavior did you observe? Please describe the bug**
The page I was on refreshed and I didn't go anywhere new.

**What is the expected behavior?**
I expected to go to one of the many other pages which are currently open for edit and have no one transcribing. The "Find a New Page" function should ensure that the page it finds is actually a new one, not simply any eligible page which is open for edit.

acdha:
This is a subset of the changes discussed with @colefemeno last week where we need to update that view to add biases in favor of keeping you within the same item, project, or campaign.

acdha:
9b7bdde6c7e10dadfd203506070d59cb84606d28 changes the logic to avoid hitting the current asset unless there are no others available.

rstorey:
Right on, it works now

PR #623, opened 2018-11-14
Update homepage “Let's Go” target

acdha:
This makes two changes:

* The homepage “Let's Go” button is now treated as a “find next transcribable page” request for the Letters to Lincoln campaign
* The redirect view now requires only the campaign slug and uses database sorting to favor finding an asset which is later than the current asset, within the same item, within the same project in that order.


acdha:
We could try that but order by random can be more expensive so we probably want to measure that.

rstorey:
Yeah, there's already a ticket for this which we can look at later: #441

acdha:
#441 is already done in 9b7bdde6c7e10dadfd203506070d59cb84606d28

PR #627, opened 2018-11-15
Latest content changes

rstorey:
Merge already-released changes into master for next release.


rstorey:
Yeah it probably should

PR #626, opened 2018-11-14
WIP: more progress indicators

acdha:
* [x] De-emphasize completed items in list views by making them partially transparent until they're hovered over:
  image
* [x] Show state counts on the progress bar's text summary table:  image
* [ ] Show item completion on progress detail page's item list
* [ ] Show project completion on campaign detail page's project list

Issue #566, opened 2018-10-31
Indicator of how much work is done on a item from the project page


elainekamlley:
User story: 
As a user, when I am looking at a project page with all of the thumbnails of the item, I want to know how much work has been done on them so that I can determine which item to work on. 
![projectview](https://user-images.githubusercontent.com/7362915/47819707-aec3be80-dd31-11e8-87b2-37bda787ae48.png)

**Is your feature request related to a problem? Please describe.**
When I am in a project, I see all the items but can't tell what I should work on. I would like to know from that page what is complete or needs work. 

**Describe the solution you'd like**
An indicator of level of completeness, either in the form of a percentage or a circular representation of the progress bar in the item view.

acdha:
We should think about how much it's worth to have a granular progress breakdown (with the extra work calculating and displaying a progress bar) versus a simple “complete” boolean display. 

One question which @jbresner may have ideas on is how we can indicate that transcription is complete but you can & should still tag things.

jbresner:
In one of the original wireframes created (see below) for the Project page, I added an indicator that showed the percentage of what was complete for a specific item. I agree with Chris that it's work discussing further about what a user gains from a more detailed view of the percentage (or another TBD calculation). I can work on updating this design with some thoughts on which approach may benefit the user in selecting items to transcribe/tag.

![project-detail](https://user-images.githubusercontent.com/4275957/47856510-c8124c80-ddbd-11e8-8d0b-a16a635d1f27.png)

elainekamlley:
To start, just apply the design of the progress bar to the items on the project page. Also persistent filter from projects page

jbresner:
![project - items listed - gby](https://user-images.githubusercontent.com/4275957/48530978-4bc14400-e868-11e8-9c8c-68503d901194.png)
![project - items listed - gby2](https://user-images.githubusercontent.com/4275957/48530979-4bc14400-e868-11e8-9cbd-3916f92c96ba.png)
![project - items listed - monochrom](https://user-images.githubusercontent.com/4275957/48530980-4bc14400-e868-11e8-8dd3-9fbdc8701f9f.png)

jbresner:
The hex values for the color schemes:

Monochromatic
Completed - 006477
Submitted for Review - 009CBA
Open for Edit - 83CFDD

Muted GBY
Completed - A2C800
Submitted for Review - 4A90E2
Open for Edit - FAAC1D

GBY
Completed - 2E8540
[...]

PR #632, opened 2018-11-15
Continued progress indicator work

acdha:
* Adjust progress bar colors and ordering
* Adjust sizes for progress bars on cards
* Display the contributor count above the progress bar
* Bring the campaign detail page's project list inline with the lower-level listings
* Display progress bars on the project and campaign pages

See #626
See #566

### DIFF
--- a/concordia/static/css/base.css
+++ b/concordia/static/css/base.css
@@ -21,9 +21,9 @@
     --bg-divider: rgba(189, 189, 189, 1); /* #BDBDBD */
     --bg-card-title: rgba(51, 51, 51, 0.7); /* #333333 */
     --bg-loading: rgba(51, 51, 51, 0.5); /* #333333 */
-    --bg-completed: #006477;
-    --bg-submitted: #009cba;
-    --bg-edit: #83cfdd;
+    --bg-completed: #00618e;
+    --bg-submitted: #2e91bf;
+    --bg-edit: #80cef3;
 
     --shadow-color: rgba(51, 51, 51, 0.8); /* #333333 */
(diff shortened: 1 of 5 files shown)


acdha:
Campaign detail page:

screen shot 2018-11-15 at 11 14 12


Project detail page:

screen shot 2018-11-15 at 11 13 45


Item detail page:

screen shot 2018-11-15 at 11 14 02

rstorey:
Not crazy about the black border around the progress bar in the cards... is there an alternative to that? It especially looks weird since the cards themselves have a much lighter border.

acdha:
@rstorey that matches the comps. We can tone it down to one of the border colors we're using elsewhere in the UI (e.g. bg-medium-gray or bg-dark-gray) or just get rid of it entirely — feel free to play with the value in the developer tools and see how it feels.

rstorey:
I like it best with no border around the progress bar: 
screen shot 2018-11-15 at 11 43 00 am

colefemeno:
@acdha @rstorey this is looking good! 
Can we also apply the de-emphasizing gray (#626 ) on the Project detail page to an item if it is complete? So all of these would be de-emphasized 
projectdetail_itemscompletenotgray_20181115

rstorey:
From @colefemeno : let's add "Not Started" - and if not too difficult, change "Open for Edit" to "In Progress"

acdha:
@colefemeno The CSS to make entirely completed things is already present on the project and campaign detail pages but we need to update the source of that data after the status change updates.

PR #633, opened 2018-11-15
WIP: Unified transcription status filtering UI

acdha:
This changes the transcription status filter to be a group of radio-style buttons on the campaign, project, and item detail pages.

Campaign detail:

image

Project detail:

image

Item detail:

image

### DIFF
--- a/concordia/forms.py
+++ b/concordia/forms.py
@@ -4,8 +4,6 @@ from django import forms
 from django.contrib.auth import get_user_model
 from django_registration.forms import RegistrationForm
 
-from .models import TranscriptionStatus
-
 User = get_user_model()
 logger = getLogger(__name__)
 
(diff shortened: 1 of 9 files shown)


acdha:
Yeah, I like the numbers, too, so I'm trying a few variations like this:

image

rstorey:
Can the image count / page count be part of the tooltip for the progress bar sections?

rstorey:
Or maybe the whole table with the numbers is a giant fancy tooltip over the entire progress bar.

acdha:
We can definitely add them as a tooltip. I'm also wondering about returning them to the buttons since a horizontal layout gives us the room needed:

image

rstorey:
I really like the side-by-side layout with the filter buttons / legend to the left and the progress bar to the right. And I think you're right about how much vertical space it takes up. I think personally I would prefer to keep the table intact and display it in a popover thingie?

acdha:
I was just testing `<details>`:

image
image

rstorey:
I vote for that one.

acdha:
I was wondering about simply making it hidden / zero height until either a click or hover on the progress bar expands it.

acdha:
I think `<details>` may have MS Edge / IE 11 issues so we probably want this to be custom. Sigh…

rstorey:
Can we use the same popover we used on the asset detail page for the nothing to transcribe help?

acdha:
Using the popover looks like this:

screenshot 2018-11-15 16 00 23

PR #640, opened 2018-11-15
WIP: Use popover for progress statistics

acdha:
I'm parking this experiment here making the progress statistics a popover triggered on click/hover/focus:

screenshot 2018-11-15 16 00 23

### DIFF
--- a/concordia/static/js/base.js
+++ b/concordia/static/js/base.js
@@ -189,3 +189,16 @@ $.ajax({url: '/account/ajax-messages/', method: 'GET'}).done(function(data) {
         });
     }
 });
+
+$(function() {
+    var $progressBar = $('#progress-bar');
+    if ($progressBar.length) {
+        $progressBar.popover({
+            content: document.getElementById('progress-bar-popover-contents')
+                .innerHTML,
+            html: true,
+            trigger: 'click hover focus',
+            placement: 'bottom'
+        });
+    }
+});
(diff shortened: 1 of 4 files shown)


jbresner:
I get the issue of redundancy for the key and the button labels, but at first glance, I'm not sure the progress bar comes across as a progress bar without any labeling.

elainekamlley:
The context with the progress bar is v. helpful and I don't think we should do the hover state. In my opinion we should rethink the radio buttons instead. Curious to know where the drop down filter went? This was is a new feature that was implemented while I was away?

rstorey:
This was part of the release that went out on Thursday afternoon to support the Monday event. I'm not in the loop on every single UI decision that's being made, so I thought this was something already discussed without me. Unfortunately the branch containing certain UI changes that were wanted for the release in advance of the event also included changing the filter dropdown to radio buttons. The reason for this is because we didn't plan the release and work was being done with the thought that we had until the end of the sprint to change things and finalize things.

acdha:
See #632, #626, and #566 for the other work which led to this. The general trend was the heavily focus in the meeting last Tuesday on exposing progress information as a navigational tool to help users find areas to contribute. That involved adding the displays on the campaign and project pages, turning the progress bar's text summary into links, and replacing the harder to use dropdown with the current buttons. This also involved changing the database schema slightly so different views didn't need to calculate the number of assets which had not been started independent from those which had some progress or vice versa, which added a fourth row to the summary table.

This pull request grew out of that combined work to reduce the amount of vertical space used by the progress bar + table, which is currently ~120 px and will grow further if we add a label (good idea) or any other field details. We might want to iterate on ways to make this display more compact since it's secondary information but the popover had the advantage of being very quick to implement for testing.

rstorey:
This issue needs attention from the UI/UX team. Closing until then.

PR #650, opened 2018-11-19
Make rate limit blocking configurable

rstorey:


concordia/settings_prod.py
```diff
@@ -78,3 +78,5 @@
 REGISTRATION_SALT = "django_registration"  # doesn't need to be secret
 
 SESSION_ENGINE = "django.contrib.sessions.backends.cached_db"
+
+RATELIMIT_BLOCK = os.getenv("RATELIMIT_BLOCK", True)
```
acdha: Doesn’t this need to evaluate that to a Boolean? Maybe just have something like `getenv(…) != “false”`?

rstorey: I was thinking we'd set it to 0 or 1 in the env vars

acdha: Okay, but don’t we still need to convert it from a string so it won’t always be true?

rstorey: Like this? `   RATELIMIT_BLOCK = bool(strtobool(str(os.getenv("RATELIMIT_BLOCK"))))` or you have a different preferred method?

rstorey:
> This entire block can be:
> 
> ```
> RATELIMIT_BLOCK = os.getenv("RATELIMIT_BLOCK", “”).lower() not in (“false”, “0”)
> ```

Thank you - hard to think straight when adrenaline is going.

PR #651, opened 2018-11-19
Make rate limit blocking configurable through env var

### LINKED ISSUE #625: Update Message after failed login attempts
Update message to be friendlier. Instead of a raised HTTP error, I expect to see a message box on the login page with a friendlier message. 

`We're sorry. Due to too many failed log in attempts we have locked your account. Please try again later or reset your password`

rstorey:
Closes #625 and #654


concordia/settings_prod.py
```diff
@@ -78,3 +78,5 @@
 REGISTRATION_SALT = "django_registration"  # doesn't need to be secret
 
 SESSION_ENGINE = "django.contrib.sessions.backends.cached_db"
+
+RATELIMIT_BLOCK = os.getenv("RATELIMIT_BLOCK", True).lower() not in ("false", "0")
```
acdha: The default value needs to be a type which implements `.lower()`:

```suggestion
RATELIMIT_BLOCK = os.getenv("RATELIMIT_BLOCK", "").lower() not in ("false", "0")
```

concordia/settings_template.py
```diff
@@ -99,6 +99,7 @@
 ]
 
 RATELIMIT_VIEW = "concordia.views.ratelimit_view"
+RATELIMIT_BLOCK = False
```
acdha: I'm wondering whether we should move the `RATELIMIT_BLOCK` down here so we can test it in staging, etc.

rstorey: `settings_prod` is used in dev, test, stage and prod environments.

concordia/views.py
```diff
@@ -209,25 +211,54 @@ def registration_rate(self, group, request):
...
-    ratelimit_block = True
+    ratelimit_block = False
+
+    def post(self, request, *args, **kwargs):
+        """
+        Handle POST requests: instantiate a form instance with the passed
+        POST variables and then check if it's valid.
+        """
+        form = self.get_form()
+        if form.is_valid():
```
acdha: I'm wondering how much we should care about the ability to guess using parallel requests here since the success path isn't limited. I'm wondering whether this process should always test the rate-limit but instead use a CAPTCHA for clients which are rate-limited:

```python
blocked = is_ratelimited()
recent_captcha = time.time() - request.session.get("captcha_validation_time", 0) > 86400

if blocked:
    if not recent_captcha:
        # redirect to a captcha or add a captcha field to the form so the next attempt will have the session value
    else:
        raise RateLimited
else:
    return super().post(request, *args, **kwargs)
```

concordia/views.py
```diff
@@ -209,25 +211,62 @@ def registration_rate(self, group, request):
...
+                request,
+                group=self.ratelimit_group,
+                key=self.ratelimit_key,
+                method=self.ratelimit_method,
+                rate=self.ratelimit_rate,
+            ):
+                raise Ratelimited(
+                    "We're sorry. Due to too many failed log in attempts "
+                    "we have locked your account. Please try again later "
+                    "or reset your password."
```
acdha: Does the password reset actually bypass this limit?

rstorey: No it doesn't @elainekamlley supplied that text

concordia/settings_template.py
rstorey: Jenkins builds the docker images which use `settings_prod` - Travis uses `settings_test` and local development runserver uses `settings_dev`

acdha: I guess where I was going with that is that as our logic gets more complicated we really need to nail it down with tests but the code-coverage tracking should help with that, too.

concordia/views.py
rstorey: I'd like to address this in a separate ticket.

rstorey: #654 created for this

concordia/settings_template.py
rstorey: So our unit test execution runs with rate limiting totally disabled - https://github.com/LibraryOfCongress/concordia/blob/master/concordia/settings_test.py#L25

It seems to me that running unit tests with rate limiting enabled introduces a bunch of problems. We would need to change the configuration of the rate for test purposes.

concordia/views.py
acdha: We can defer this but then I'm wondering whether we should either leave `RATELIMIT_BLOCK = True` or increase the limits because this change is effectively removing the reason for rate-limiting this view since an attacker guessing passwords still learns when they have the correct password.

rstorey: I see what you mean. I guess I should take care of this today then

rstorey: Okay, so digging into this a bit more, I don't follow the logic. If the login request is rate limited (blocked) and the user has completed a captcha, it shouldn't raise a RateLimited exception. In fact if we allow people to complete a captcha for rate limited logins then the RateLimited exception would never come up, would it? So I think it should be more like the following. Also, the test for recent_captcha should be that the difference is less than 86400, not greater than

```
blocked = is_ratelimited()
recent_captcha = time.time() - request.session.get("captcha_validation_time", 0) < 86400

if blocked:
    if not recent_captcha:
        # redirect to a captcha or add a captcha field to the form so the next attempt will have the session value
else:
    return super().post(request, *args, **kwargs)
```

acdha: Hmmm, yes, that needed more time. The idea was that we never actually test the password before checking the CAPTCHA to avoid leaking that information but your simplified version does that.

concordia/views.py
```diff
@@ -221,34 +221,27 @@ class ConcordiaRegistrationView(RatelimitMixin, RegistrationView):
...
+            key=self.ratelimit_key,
+            method=self.ratelimit_method,
+            rate=self.ratelimit_rate,
+        )
+        recent_captcha = (
+            time.time() - request.session.get("captcha_validation_time", 0) < 86400
+        )
 
-            return self.form_invalid(form)
+        if blocked and not recent_captcha:
```
acdha: This looks like a good start on this side – should we attempt to move the login processing into JavaScript and reuse the existing AJAX captcha view, add a separate view, or just add it to this form dynamically?

concordia/views.py
```diff
@@ -221,34 +221,27 @@ class ConcordiaRegistrationView(RatelimitMixin, RegistrationView):
...
-                )
+        blocked = is_ratelimited(
+            request,
+            group=self.ratelimit_group,
+            key=self.ratelimit_key,
+            method=self.ratelimit_method,
+            rate=self.ratelimit_rate,
+        )
+        recent_captcha = (
+            time.time() - request.session.get("captcha_validation_time", 0) < 86400
```
acdha: I'm tempted to parenthesize the subtraction here just to make it clear to humans what's being compared

concordia/settings_template.py
acdha: I think we should use the settings override decorator / context manager to selectively enable it only for the cases where we're actually testing this logic so we don't have random collateral damage.

concordia/views.py
```diff
@@ -209,25 +211,55 @@ def registration_rate(self, group, request):
...
+            raise Ratelimited()
+        else:
+            return super().post(request, *args, **kwargs)
 
 
 def ratelimit_view(request, exception=None):
-    template_name = "429.html"
     status_code = 429
+
+    # TODO: return JSON response for JSON requests
```
acdha: I'm thinking this is as simple as building the context dictionary below and then doing the Accept check to return either `JsonResponse` or `HttpResponse` as below

concordia/views.py
rstorey: Black took 'em out.

concordia/views.py
rstorey: I was thinking add it to the form dynamically - do you have a preference?

concordia/views.py
rstorey: This is going to be moot since the logic with captcha won't raise any Ratelimited exception.

concordia/views.py
```diff
@@ -209,25 +211,58 @@ def registration_rate(self, group, request):
...
+        response = JsonResponse({"exception": str(exception), "status": status_code})
+    else:
+        template_name = "429.html"
+        template = loader.get_template(template_name)
+
+        response = HttpResponse(
+            template.render({"exception": exception}), status=status_code
+        )
+
+    response["Retry-After"] = 15 * 60
```
acdha: Does the ratelimiter pass information on the request about the triggered rate limit period? It'd be nice if `Retry-After` was based on the actual value, although there's not too much harm from a few extra retries.

concordia/views.py
acdha: 09a4009a52491a1eba0caf5c9b4ae4f3affe8d49 implements this so I'm marking this as resolved.

concordia/templates/429.html
```diff
@@ -28,21 +28,13 @@
         <h1>HTTP 429 Error</h1>
 
         <p>
-            Please wait a bit, then try again.
+            {% if exception %}
+                {{ exception }}
+            {% else %}
```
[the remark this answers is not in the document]
rstorey: Yes - added.

concordia/views.py
rstorey: It doesn't as far as I can tell. You actually need to know the rate being used in order to even call the helper function to get a True/False of whether the request was ratelimited.

concordia/views.py
```diff
@@ -239,11 +239,9 @@ def post(self, request, *args, **kwargs):
         ) < 86400
 
         if blocked and not recent_captcha:
-            raise Ratelimited(
-                "We're sorry. Due to too many failed log in attempts "
-                "we have locked your account. Please try again later "
-                "or reset your password."
-            )
+            form = self.get_form()
+            form.fields["captcha"] = CaptchaField()
```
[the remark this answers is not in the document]
rstorey: I don't follow. I tested this and the default form field handling seemed to work fine.

concordia/views.py
acdha: Looking at the code confirms that it only sets the boolean `limited` value. I'm inclined not to spend more time exposing this now.

concordia/views.py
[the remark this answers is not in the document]
rstorey: The django-simple-captcha app does that. https://django-simple-captcha.readthedocs.io/en/latest/usage.html#adding-to-a-form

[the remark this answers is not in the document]
rstorey: All I can say is that it works the way it is now. Please feel free to check out the branch and test it yourself.

elainekamlley:
@rstorey @acdha review the contact us link to 429 page?

concordia/views.py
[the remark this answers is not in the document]
acdha: d3c442227be7c298ac3f70510d00eb560920b39a works in my testing: the captcha shows up when the rate-limit is hit and is then required but the form can successfully be submitted.

rstorey:
> @rstorey @acdha review the contact us link to 429 page?

Yes - it reads "Seeing this page a lot? Contact Us" - would welcome suggestions for the text.

concordia/views.py
```diff
@@ -239,11 +241,16 @@ def post(self, request, *args, **kwargs):
...
-            return self.form_invalid(form)
+
+        if form.is_valid():
+            return self.form_valid(form)
         else:
-            return super().post(request, *args, **kwargs)
+            return self.form_invalid(form)
+
+    def form_valid(self, form):
+        self.request.session["captcha_validation_time"] = time.time()
```
rstorey: Ah now I see what you mean.

PR #653, opened 2018-11-19
Admin: add related object navigation links

acdha:
Some features requested in #203:

* The user change form now has a link to the transcription changelist filtered by that user
* The campaign, project, item, and asset forms now have links to the changelists for the lower-level related objects
* The project, item, asset, and transcription forms now have links up to the related parent objects


acdha:
@rstorey Sorry about that – I just rebased this against master

acdha:
@rstorey https://github.com/LibraryOfCongress/concordia/pull/653/commits/3d63ed8a4ca624486deaf1b8d4a348986ef7b58a moves a bit of code around but I like the idea of avoiding one giant file here. It wouldn't be much work to change if we don't want to take that on now, however.

elainekamlley:
@acdha I'm unclear on what use cases these cover in #203. Is this related to @lalgee's request of 

```As a community manager I need to be able to find the assets/transcripts a specific user has worked on. Suggested solution: In admin page for a user, link to their actions (like on user profile page) AND be able to search and sort by user in the admin transcriptions page```?

elainekamlley:
Ah got it. Thanks, @acdha

PR #667, opened 2018-11-27
Greater opacity for disabled buttons

### LINKED ISSUE #657: Review disabled button state
This is a bookmark to review button disabled states with @jbresner along with other UI cues for things like when rate-limiting is active or indicating why controls are disabled.

acdha:
This increases the distinction between enabled and disabled buttons (closes #657) to be more obvious on lower-contrast displays.

image

### DIFF
--- a/concordia/static/css/base.css
+++ b/concordia/static/css/base.css
@@ -352,6 +352,11 @@ ul.nav-secondary {
     left: -5rem;
 }
 
+.btn.disabled,
+.btn:disabled {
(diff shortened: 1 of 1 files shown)


jbresner:
@acdha - I know it's very subjective (and very subtle) but can you bump it to 30%? It makes it slightly more noticeable on the gray background.

acdha:
Done — looks like this:

image

jbresner:
@acdha - Thanks. Let's go with this for now.

PR #675, opened 2018-11-28
WIP: Enable unicode support for slug fields

### LINKED ISSUE #214: Validate slug characters
Steps to reproduce:
1. Create a campaign and/or a project that has accented non-ASCII characters.
2. Create a campaign and/or a project that has punctuation in it such as parentheses.
3. Browse to those campaigns and projects in the site and check the URLs.

Observed behavior:
1. Unknown

Expected behavior:
1. non-ASCII characters should be fine for URL slugs
2. Punctuation should not be allowed in URL slugs

acdha:
This sets allow_unicode=True on all slug creation/validation calls. It still needs to work around the lack of a Unicode-compatible URL path converter (https://docs.djangoproject.com/en/2.1/_modules/django/urls/converters/) which breaks URL routing.

Closes #214

### DIFF
--- a/cloudformation/infrastructure/elasticache-feature.yaml
+++ b/cloudformation/infrastructure/elasticache-feature.yaml
@@ -1,6 +1,6 @@
 Description: >
     This template deploys an elasticache cluster to the provided VPC and subnets
-    
+
 Parameters:
 
     EnvironmentName:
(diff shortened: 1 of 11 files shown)


acdha:
I think merging it at this point is safe — it's more code than I'd like until Django has the equivalent unicode slug pattern converter but it's not that much more code

acdha:
Oh, we should rebase this branch before merging it since those changes weren't part of this work

PR #677, opened 2018-11-29
Bagit export updates

### LINKED ISSUE #664: BagIt Export updates
Update the` bag-info.txt` to meet the routine acquisitions needs in DCM. These include the following fields: 

- bag_id
- LC-Bag-Id 
- LC-Project 
- LCCN-Permalink 
- License-Information 
- Content-Access 
- Content-Custodian 
- Work-Requestor 
- External-Indentifier

rstorey:
Closes #664

### DIFF
--- a/exporter/views.py
+++ b/exporter/views.py
@@ -2,6 +2,7 @@ import os
 import re
 import shutil
 import tempfile
+from logging import getLogger
 
 import bagit
 import boto3
@@ -14,6 +15,8 @@ from django.views.generic import TemplateView
 from tabular_export.core import export_to_csv_response, flatten_queryset
 
 from concordia.models import Asset, Transcription
+
+logger = getLogger(__name__)
 
 
 def get_latest_transcription_data(asset_qs):
@@ -34,32 +37,36 @@ def get_original_asset_id(download_url):
     """
     if download_url.startswith("http://tile.loc.gov/"):
         pattern = r"/service:([A-Za-z0-9:\-]*)/"
-        asset_id = re.search(pattern, download_url)
+        asset_id = re.search(pattern, download_url).group(1)
         assert asset_id
-        return asset_id.group(1)
+        logger.debug("Found asset ID %s in download URL %s" % (asset_id, download_url))
+        return asset_id
     else:
+        logger.warning(
+            "Download URL doesn't start with tile.loc.gov: %s" % download_url
+        )
         return download_url
 
 
 def do_bagit_export(assets, export_base_dir, export_filename_base):
     """
-    Creates temp directory structure for source data.
-    Executes bagit.py to turn temp directory into bagit strucutre.
+    Executes bagit.py to turn temp directory into LC-specific bagit strucutre.
     Builds and exports bagit structure as zip.
     Uploads zip to S3 if configured.
-    Removes all temporary directories and files.
     """
-    os.makedirs(export_base_dir, exist_ok=True)
 
     for asset in assets:
-        dest_folder = os.path.join(export_base_dir, asset.item.item_id)
-        os.makedirs(dest_folder, exist_ok=True)
+        asset_id = get_original_asset_id(asset.download_url)
+        logger.debug("Exporting asset %s into %s" % (asset_id, export_base_dir))
+
+        asset_id = asset_id.replace(":", "/")
+        asset_path, asset_filename = os.path.split(asset_id)
+
+        dest_path = os.path.join(export_base_dir, asset_path)
+        os.makedirs(dest_path, exist_ok=True)
 
         # Build transcription output text file
-        text_output_path = os.path.join(
-            dest_folder,
-            "%s.txt" % os.path.basename(get_original_asset_id(asset.download_url)),
-        )
+        text_output_path = os.path.join(dest_path, "%s.txt" % asset_filename)
         with open(text_output_path, "w") as f:
             f.write(asset.latest_transcription or "")
 
@@ -88,6 +95,7 @@ def do_bagit_export(assets, export_base_dir, export_filename_base):
     s3_bucket = getattr(settings, "EXPORT_S3_BUCKET_NAME", None)
 
     if s3_bucket:
+        logger.debug("Uploading exported bag to S3 bucket %s" % s3_bucket)
         s3 = boto3.resource("s3")
         s3.Bucket(s3_bucket).upload_file(
             "%s.zip" % export_base_dir, "%s" % export_filename
--- a/exporter/views.py
+++ b/exporter/views.py
@@ -39,7 +39,7 @@ def get_original_asset_id(download_url):
         pattern = r"/service:([A-Za-z0-9:\-]*)/"
         asset_id = re.search(pattern, download_url).group(1)
         assert asset_id
-        logger.debug("Found asset ID %s in download URL %s" % (asset_id, download_url))
+        logger.debug("Found asset ID %s in download URL %s", asset_id, download_url)
         return asset_id
     else:
         logger.warning(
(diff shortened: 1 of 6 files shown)


exporter/views.py
```diff
@@ -35,13 +37,81 @@ def get_original_asset_id(download_url):
     """
     if download_url.startswith("http://tile.loc.gov/"):
         pattern = r"/service:([A-Za-z0-9:\-]*)/"
-        asset_id = re.search(pattern, download_url)
+        asset_id = re.search(pattern, download_url).group(1)
```
acdha: This will fail if it gets a value which doesn't match the pattern. We should probably do the `if not match: …` check even if that's unlikely.

exporter/views.py
```diff
@@ -35,13 +37,81 @@ def get_original_asset_id(download_url):
     """
     if download_url.startswith("http://tile.loc.gov/"):
         pattern = r"/service:([A-Za-z0-9:\-]*)/"
```
acdha: At the very least our service identifiers should be at least one character long, right?

```suggestion
        pattern = r"/service:([A-Za-z0-9:\-]+)/"
```

exporter/views.py
```diff
@@ -35,13 +37,81 @@ def get_original_asset_id(download_url):
...
-        asset_id = re.search(pattern, download_url)
+        asset_id = re.search(pattern, download_url).group(1)
         assert asset_id
-        return asset_id.group(1)
+        logger.debug("Found asset ID %s in download URL %s" % (asset_id, download_url))
```
acdha: Since this will be called potentially very many times, I would not force string interpolation:

```suggestion
        logger.debug("Found asset ID %s in download URL %s", asset_id, download_url)
```

exporter/views.py
```diff
@@ -35,13 +37,81 @@ def get_original_asset_id(download_url):
...
+        return asset_id
     else:
+        logger.warning(
+            "Download URL doesn't start with tile.loc.gov: %s" % download_url
```
acdha:
```suggestion
            "Download URL doesn't start with tile.loc.gov: %s", download_url
```

exporter/views.py
```diff
@@ -35,13 +37,81 @@ def get_original_asset_id(download_url):
...
+def do_bagit_export(assets, export_base_dir, export_filename_base):
+    """
+    Executes bagit.py to turn temp directory into LC-specific bagit strucutre.
+    Builds and exports bagit structure as zip.
+    Uploads zip to S3 if configured.
+    """
+
+    for asset in assets:
+        asset_id = get_original_asset_id(asset.download_url)
+        logger.debug("Exporting asset %s into %s" % (asset_id, export_base_dir))
```
acdha: More deferred string interpolation:

```suggestion
        logger.debug("Exporting asset %s into %s", asset_id, export_base_dir)
```

exporter/views.py
```diff
@@ -35,13 +37,81 @@ def get_original_asset_id(download_url):
...
+    archive_name = export_base_dir
+    shutil.make_archive(archive_name, "zip", export_base_dir)
+
+    export_filename = "%s.zip" % export_filename_base
+
+    # Upload zip to S3 bucket
+    s3_bucket = getattr(settings, "EXPORT_S3_BUCKET_NAME", None)
+
+    if s3_bucket:
+        logger.debug("Uploading exported bag to S3 bucket %s" % s3_bucket)
```
acdha:
```suggestion
        logger.debug("Uploading exported bag to S3 bucket %s", s3_bucket)
```

exporter/views.py
```diff
@@ -84,74 +154,65 @@ def get(self, request, *args, **kwargs):
...
+    def get(self, request, *args, **kwargs):
+        campaign_slug = self.kwargs["campaign_slug"]
+        project_slug = self.kwargs["project_slug"]
+        item_id = self.kwargs["item_id"]
+
+        asset_qs = Asset.objects.filter(
+            item__project__campaign__slug=campaign_slug,
+            item__project__slug=project_slug,
+            item__item_id=item_id,
+            transcription_status="completed",
```
acdha: This should probably use the same constants we're trying to use elsewhere:

```suggestion
            transcription_status=TranscriptionStatus.COMPLETED,
```

exporter/views.py
```diff
@@ -84,74 +154,65 @@ def get(self, request, *args, **kwargs):
...
+
+class ExportProjectToBagIt(TemplateView):
     @method_decorator(staff_member_required)
     def get(self, request, *args, **kwargs):
         campaign_slug = self.kwargs["campaign_slug"]
+        project_slug = self.kwargs["project_slug"]
         asset_qs = Asset.objects.filter(
             item__project__campaign__slug=campaign_slug,
+            item__project__slug=project_slug,
             transcription_status="completed",
```
acdha:
```suggestion
            transcription_status=TranscriptionStatus.COMPLETED,
```

exporter/views.py
```diff
@@ -84,74 +154,65 @@ def get(self, request, *args, **kwargs):
...
+            return do_bagit_export(assets, export_base_dir, export_filename_base)
+
+
+class ExportCampaignToBagit(TemplateView):
+    @method_decorator(staff_member_required)
+    def get(self, request, *args, **kwargs):
+        campaign_slug = self.kwargs["campaign_slug"]
+        asset_qs = Asset.objects.filter(
+            item__project__campaign__slug=campaign_slug,
+            transcription_status="completed",
```
acdha:
```suggestion
            transcription_status=TranscriptionStatus.COMPLETED,
```

acdha:
> When we are ready to talk about/implement public exports, I'm not sure that BagIt 1.0 bags are what we want to use. I think the CSV export is better, not the least of which is because it's a streaming response that won't OOM on large exports.

I suspect that the best answer here is going to be Celery and the work on bagit-python to make it upload directly to S3 (and that could be safely made public with some sort of throttle). That's definitely a separate project but we might want to see whether there's something we can do here to avoid OOMs — do you know where in the code it's hitting that limit?

rstorey:
> > When we are ready to talk about/implement public exports, I'm not sure that BagIt 1.0 bags are what we want to use. I think the CSV export is better, not the least of which is because it's a streaming response that won't OOM on large exports.
> 
> I suspect that the best answer here is going to be Celery and the work on bagit-python to make it upload directly to S3 (and that could be safely made public with some sort of throttle). That's definitely a separate project but we might want to see whether there's something we can do here to avoid OOMs — do you know where in the code it's hitting that limit?

I haven't actually personally witnessed an OOM, so not sure.

rstorey:
SInce we're only getting completed transcriptions in the BagIt export, either the assets are published or were published. I think the scenario where something is published, completed, unpublished, and then exported might be a valid use case, so I'm not going to add a filter for publish status at this time. If @elainekamlley wants to weigh in, I defer to them.

exporter/views.py
rstorey: I changed it - let me know what you think.

exporter/views.py
```diff
@@ -34,14 +36,88 @@ def get_original_asset_id(download_url):
...
-        pattern = r"/service:([A-Za-z0-9:\-]*)/"
+        pattern = r"/service:([A-Za-z0-9:\-]+)/"
         asset_id = re.search(pattern, download_url)
-        assert asset_id
-        return asset_id.group(1)
+        if not asset_id:
+            logger.error(
+                "Couldn't find a matching asset ID in download URL %s", download_url
+            )
+            raise AssertionError
```
acdha: This could be some other exception class but I care less about that than that it’s checked since it seems like this condition will usually be telling us that we got some bogus data from the API, and that edge case probably means we need to get some records fixed

PR #678, opened 2018-11-29
Improved user profile contribution history

### LINKED ISSUE #532: Add total number of contributions to profile page
Users can't quick find out how many contributions they have made to campaigns on the platform. 

Show the total number of contributions next to or under the My Contributions section header on the profile page.

Ideas:

> My Contributions (10)

>**My Contributions**
>Contributions: 10

**Describe alternatives you've considered**
Show total number of contributions and number per campaign in a table or list. 

| Campaign           | Contributions |
|--------------------|---------------|
| Letters to Lincoln | 10            |
| Scouting Reports   | 20            |

### LINKED ISSUE #676: Item statuses are not updating on user profile
User story shared by a volunteer via email:
On my profile page there is a list of the documents I have worked on. Many of them say "In Progress". However, when I click on the link to the document, it shows as complete. 
I experienced the bug on my account page as well.

1. As a logged in user, visit your account page.
2. Scroll to an item you have worked on that says "In Progress" and click the item link.
3. In nearly every case the item is now complete.

The item status on the user profile page should accurately reflect the worked-on-item's current status.

 - OS: Windows
 - Browser: Chrome

**Additional context**
Possibly linked to work of #621 and #532

acdha:
This makes the contribution section of the profile page work more like how people believed it was working all along. Now it takes the transcription and review activities which the user has participated in and displays the user’s last action and the asset’s current status.

screenshot_2018-11-30 crowd user profile 1

Closes #532 
Closes #676

### DIFF
--- a/concordia/views.py
+++ b/concordia/views.py
@@ -283,6 +283,10 @@ class AccountProfileView(LoginRequiredMixin, FormView, ListView):
     ordering = ("-updated_on",)
     paginate_by = 12
 
+    def post(self, *args, **kwargs):
+        self.object_list = self.get_queryset()
+        return super().post(*args, **kwargs)
+
     def get_queryset(self):
         transcriptions = Transcription.objects.filter(
             Q(user=self.request.user) | Q(reviewed_by=self.request.user)
--- a/concordia/views.py
+++ b/concordia/views.py
@@ -23,8 +23,7 @@ from django.core.exceptions import ValidationError
 from django.core.mail import send_mail
 from django.core.paginator import Paginator
 from django.db import connection
-from django.db.models import Case, Count, F, IntegerField, Max, Q, When
-from django.db.models.functions import Cast
+from django.db.models import Case, Count, IntegerField, Max, Q, When
 from django.db.transaction import atomic
 from django.http import HttpResponse, JsonResponse
 from django.shortcuts import get_object_or_404, redirect, render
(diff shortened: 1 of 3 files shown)


concordia/views.py
```diff
@@ -276,12 +277,33 @@ class AccountProfileView(LoginRequiredMixin, FormView):
...
-        ctx["transcriptions"] = (
-            Transcription.objects.filter(user=self.request.user)
-            .select_related("asset__item__project__campaign")
-            .order_by("asset__pk", "-pk")
-            .distinct("asset")
+
+        transcriptions = Transcription.objects.filter(user=self.request.user)
+        items = Item.objects.filter(pk__in=transcriptions.values("asset__item__pk"))
+        items = items.select_related("project", "project__campaign")
+        items = items.order_by("project__campaign__title", "project__title", "item_id")
```
rstorey: I think we're missing another order by field here, for `transcription.created_on desc`

elainekamlley:
I think preserving the same headers in the current profile page is best for now. We want to know more about what our users want and what information would be useful. I think the bug that @lalgee pointed out was that the status on the user profile wasn't updating to reflective the actual status of the asset. Could we just hone in on adding the accurate count of a user's contributions and fixing `Current Status` and then engage our users to give us more guidance?

acdha:
The counts and status values have always been accurate but they're not what you think: they're the transcription records, not the corresponding asset: if you hit the save button 5 times, you'll see 5 rows and the status will be what ever each entry's status was at that time. 

The simplest change would be renaming “Current Status” to “Status” since that reflects the historic value but based on discussion and #532 it doesn't seem like that's really what people want on this page. If we changed it to report the assets rather than rolling up to the item level we will still need to do something to handle the length of the campaign, project, item and then asset titles and the need for pagination will be even more pronounced.

rstorey:
I think asset level is what users would expect to see in this listing, rather than item. It seems like displaying the history of a user's contributions has different goals than showing a user the current status of assets to which they have contributed. Perhaps there are two separate columns, one to show what the user contributed (whether they edited, submitted, or accepted) and another to indicate the current status. It's nice for users to see that they made contributions to assets that become completed.

rstorey:
I also think this listing is a place where users might want to go back to see exactly what they did or to find an interesting asset that they knew they worked on before. So that's why I think assets, rather than items, are what we should list.

acdha:
My main concern is the size of that list and the query complexity need to pull in the full history. If we're paginating that might not be too bad but it makes me wonder whether we want to have a summary view which allows you to drill down into your history since [hopefully] many of our users will have hundreds or thousands of transcription records.

lalgee:
Since profile page refinement is planned for an upcoming sprint, I think we should leave it at asset info for now and can address item views, pagination, etc. at that time.  The most straightforward fix for now would be @acdha's suggestion to change the status column name to something more descriptive - would "Contribution Status" fit? (and does it make sense? @VVH @elainekamlley @colefemeno)

acdha:
Leaving it means showing the full log of every transcription record (potentially many per asset — a quick database query shows that there are individual users who would see the same asset listed as many as 58 times). That definitely puts weight on the need for pagination — we have users who'd have hundreds of rows in that table — and it's also hard to read with the wrapped campaign, project, and item titles.

I don't think “Contribution Status” is going to clarify enough since I think relatively few people want to know each stage of an asset compared to a final state, whether that was the way they left it or the current status.

acdha:
The current screenshot shows the new activity summary broken down by campaign (#532) and the updated asset display which shows each asset you've contributed to (either by transcribing or reviewing) with your most recent date & action and the current status of the asset to avoid the problem in #676 where someone who edited a transcription would see the status as edit (i.e. what their transcription record was at the time they made it) but not that the asset had subsequently been completed.

This also enables pagination to avoid the significant performance and usability problems of displaying a list with more than a few contributions — the time to generate the page should consistently stay in the range of tens of milliseconds rather than seconds for prolific contributors.

lalgee:
This looks good @acdha.  The only change I'd make is to remove the project names in the listing.  They are confusing to me.  For now can you take them out and just sort by what the user worked on most recently.  I think further faceting/sublisting would be great when we really dig into profile refinement.

rstorey:
I think the main thing left to work on here is the grouping / ordering of the listing of contributions. I don't know that it's helpful to have them grouped by project. I don't know that we actually need to include campaign and project info here at all - the item titles make it clear which bucket of stuff this page is in. I think it might be a good idea to take out the campaign and project headings along with their groupings and order all the contributions in reverse chronological order across projects. I also prefer having the actual date and time instead of the naturaltime thing.

acdha:
The item titles don’t consistently have info and I wanted the sections to help provide navigational cues for the overall status

acdha:
There is no grouping: it’s strict chronological ordering but it displays the campaign/project info when it changes

Issue #550, opened 2018-10-30
Add Welcome Guide and Challenge to Carousel 


elainekamlley:
User story: 

- As a community manager, I want to highlight a challenge so that I we can signal to users what campaign to go to. 
- As a community manager, I want to add to the carousel
- As user, I am not quite sure where to start. I'd like to read instructions or a tutorial as a way to onboard. 

Acceptance criteria: 
- write CTA language for Welcome guide and Challenge
- Get images to the carousel image dimensions
- Add right side CTA as an option like in the loc.gov carousel below:  
<img width="1220" alt="screen shot 2018-10-30 at 12 24 36 pm" src="https://user-images.githubusercontent.com/7362915/47733546-df76fb80-dc3e-11e8-9acc-89b1027ae4a3.png">

jbresner:
Here are two of the images we may consider using for the home page carousel. The Lincoln letter gets cropped a bit awkwardly so let me know if you think we need to replace it with something else or try another layout with it.

![branch-rickey](https://user-images.githubusercontent.com/4275957/47931536-6decb680-dea5-11e8-9ffb-fb37e83c3198.jpg)
![lincoln-letter-2](https://user-images.githubusercontent.com/4275957/47931541-734a0100-dea5-11e8-9672-368f7427aa8f.jpg)

elainekamlley:
@jbresner mind adding your recommendations on carousel rules?

jbresner:
@elainekamlley - Here are some general guidelines we should consider for the carousel if we use auto-rotation:

1. Slides should be displayed about 5 seconds before advancing to the next image (if there's a lot of text on the image, which there shouldn't be, then we could bump up the delay a second or two).
2. On hover, the carousel should pause and no longer auto-rotate. If the cursor is no longer over the carousel, the auto-rotate sequence may continue.
3. If the user is interacting with the carousel controls (i.e. the left/right navigation and/or the slide indicator buttons) then the system should also pause the auto-rotate.
4. At the mobile breakpoint, the auto-rotate should be disabled.
5. Also for mobile, the user should be able to swipe to advance slides in the carousel.

We can use the same UI elements that we have on the LOC.gov home page if that expedites building this out. I'll add a mockup of the UI for reference.

jbresner:
[...]

PR #680, opened 2018-11-30
Update homepage carousel

acdha:
This is collecting the changes from #550:

* [x] pause rotation when the pointer is over the carousel
* [x] add navigation arrows
* [x] add carousel indicators
* [x] add new carousel images
* [x] add new carousel text & link targets
* [x] Adjust spacing for the text box to match [@jbresner's comps](https://github.com/LibraryOfCongress/concordia/issues/550#issuecomment-439121557)


acdha:
Here's an example screenshot of the current design, which is basically Bootstrap with larger navigation arrows:

image
image
image

We probably need to think about something like a box-shadow to make the arrows and indicators more consistently visible across a range of background colors, especially since they're white and our default homepage image has lots of the same color.

acdha:
Here's what it looks like with the latest commits:

image

image

image

I adjusted some of the spacing slightly from @jbresner's comps to ensure that the text always fits and that the overlay is always distinct from the carousel navigation buttons.

acdha:
One other question, which could be deferred along with updating the alt text: do we ever want the button text to be something other than “LET's GO”?

elainekamlley:
@acdha I think so eventually. For the Welcome Guide I think `Learn more` would be more appropriate

acdha:
@elainekamlley let's put a ticket in the next sprint. It'll be easy to add but I don't want to add another review cycle here.

Issue #507, opened 2018-10-24
Reduce friction on registration/login


benwbrum:
When I first tried to register for an account, I encountered these difficulties:

* Password requirements are discouragingly strict; consider requirements that are both easier and more secure than arbitrary special character requirements.  cf. [XKCD on password strength](https://xkcd.com/936/)
* When I created an account, it was not at all obvious that I needed to check my email and click a validation link.
* When a user who has not clicked a validation link attempts to log in, the error message is a standard one of the form _bad username or password_, and does not hint that the account was not validated.  
* Generally, when a user registers for their account, they are not automatically logged into the system.  This means additional clicking, emailing, etc before they can actually contribute to login-required features.

I suggest reviewing the process of new user registration and creating specific issues addressing the friction points.

VVH:
Any issue that references an XKCD cartoon gets an A+ in my book. Thanks for this thoughtful feedback @benwbrum. We'll circle up about this and see which aspects we can address when. Much appreciated.

acdha:
See #168 for the plan to switch to zxcvbn to provide better password strength checks which are also less onerous.

lalgee:
Remove slashes from username instructions.  Currently underneath the field for creating a username it says "Required. 150 characters or fewer. Letters, digits and @/./+/-/_ only." This leads some users to believe that slashes are allowed.

elainekamlley:
@lalgee Let's take your comment and create a separate ticket for that. We can include that in this next sprint

acdha:
Specifically, we should note that this is overriding this value: https://github.com/django/django/blob/e81955401885a93a459bcc130642b6ea5bf4ba4d/django/contrib/auth/models.py#L301

elainekamlley:
Happy belated anniversary of this issue! Over the last year and especially in the latest release, we've many improvements to the registration process. Thanks to you, @benwbrum for creating this issue as it was a nice outline of the work. To learn more about what else we've been up to, check out [releases](https://github.com/LibraryOfCongress/concordia/releases. Thanks again!

PR #701, opened 2018-12-07
Customize username help text on registration form

acdha:
This makes the list of allowed punctuation easier to parse — see
https://github.com/LibraryOfCongress/concordia/issues/507#issuecomment-443840209 for the motivation.

Using the default Bootstrap styles for `<kbd>` here's what this looks like: 

image

### DIFF
--- a/concordia/forms.py
+++ b/concordia/forms.py
@@ -19,7 +19,7 @@ class UserRegistrationForm(RegistrationForm):
         help_texts = {
             "username": (
                 "Required. 150 characters or fewer. Usernames can contain"
-                " letters, digits and any of these punctuation symbols:"
+                " letters, numbers, and any of these punctuation symbols:"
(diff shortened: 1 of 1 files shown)


acdha:
I don't think we can fit it on one line without removing too much detail, although we can safely omit “Required.” since the browser will tell them that if they try to submit the form without it.

acdha:
With the last two commits it looks like this:

image

I don't think it's worth worrying about it wrapping to a second line but if we did it would probably be better to make the form less narrow rather than try to trim this text too aggressively (even removing all of the spacing and description around the symbols isn't enough to fit it currently).

lalgee:
I'm ok with the 2 lines.   But at mefe's excellent suggestion, can you swap the order of the sentences?  Then I think we'll be good to go.

acdha:
I haven't seen a suggestion from @colefemeno but assuming that it is just as simple as reversing the order:

image

PR #707, opened 2018-12-11
Update find-next-page ranking logic

acdha:
This changes the rules slightly but was mostly intended to add test coverage for  the selection rules for the find-next-page feature: favoring an asset which is ahead of the current one, unstarted, in the same item, or in the same project, in that order.

The ranking changes are relative weights: staying within the same item counts more than just same project and moving forward counts more than finding an unstarted item so it will try to move you to a different page even if you happen to be on the only remaining unstarted page.

### DIFF
--- a/concordia/tests/test_view.py
+++ b/concordia/tests/test_view.py
@@ -895,3 +895,94 @@ class TransactionalViewTests(JSONAssertMixin, TransactionTestCase):
                 "transcriptions:project-detail", args=(campaign.slug, project.slug)
             ),
         )
+
+    def test_find_next_transcribable_hierarchy(self):
+        """Confirm that find-next-page selects assets in the expected order"""
+
+        asset = create_asset()
+        item = asset.item
+        project = item.project
+        campaign = project.campaign
+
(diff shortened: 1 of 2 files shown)


concordia/tests/test_view.py
```diff
@@ -895,3 +895,94 @@ def test_find_next_transcribable_project_redirect(self):
...
+            ),
+            slug="test-asset-in-same-campaign",
+        )
+
+        # Now that we have test assets we'll see what find-next-page gives us as
+        # successive test records are marked as submitted and thus ineligible.
+        # The expected ordering is that it will favor moving forward (i.e. not
+        # landing you on the same asset unless that's the only one available),
+        # and will keep you closer to the asset you started from (i.e. within
+        # the same item or project in that order).
```
rstorey: Is this comment accurate?

acdha: I believed so when I wrote it but if you have time to review the code that can't hurt

rstorey: It looks like it's the other way around in the implementation - same project or item (the comment indicates it's same item or project).

PR #708, opened 2018-12-11
Update homepage carousel control design

acdha:
This builds on #680 to improve contrast with the black/white-heavy backgrounds by adding a border and colored hover state.

Screenshots:

screenshot_2018-12-11 crowd home
screen shot 2018-12-11 at 12 34 06
screen shot 2018-12-11 at 12 34 27
screen shot 2018-12-11 at 12 34 15

### DIFF
--- a/concordia/static/css/base.css
+++ b/concordia/static/css/base.css
@@ -674,6 +674,22 @@ ul.nav-secondary {
     max-width: 360px;
 }
 
+.view-homepage .carousel-control-next-icon {
+    background-image: url("data:image/svg+xml;charset=utf8,%3Csvg xmlns='http://www.w3.org/2000/svg' fill='%23fff' viewBox='0 0 8 8'%3E%3Cpath stroke='%23333333' stroke-width='.25' d='M2.75 0l-1.5 1.5 2.5 2.5-2.5 2.5 1.5 1.5 4-4-4-4z'/%3E%3C/svg%3E");
+}
+
+.view-homepage .carousel-control-next:hover .carousel-control-next-icon {
+    background-image: url("data:image/svg+xml;charset=utf8,%3Csvg xmlns='http://www.w3.org/2000/svg' fill='%23fc4c02' viewBox='0 0 8 8'%3E%3Cpath stroke='%23ffffff' stroke-width='.25' d='M2.75 0l-1.5 1.5 2.5 2.5-2.5 2.5 1.5 1.5 4-4-4-4z'/%3E%3C/svg%3E");
+}
+
+.carousel-control-prev-icon {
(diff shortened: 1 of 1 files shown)


jbresner:
Can we change the arrows to solid white instead of the slightly transparent state when they're not active? Also, at least in the screen shots, the dot/circle navigation looks odd. Is that getting updated as well?

acdha:
The page indicators are currently just a div with a border-radius. Do you want them to use the same colors as the arrows?

acdha:
Removing the opacity from the next/previous controls looks like this:
screen shot 2018-12-11 at 13 35 53

acdha:
@jbresner removing the opacity and adding the same border / fill colors for the indicators looks like this:

screen shot 2018-12-11 at 13 42 32

jbresner:
@acdha - I like the .5 stroke on the arrows but they're kinda beefy looking. Can we size them similar to what they are on the LOC.gov page? I've attached SVGs for left/right and active/inactive states. Also, instead of ovals for the slide indicator, can we use circles instead (I'm still using the 1-px stroke for those)? I've attached SVGs for those (active/inactive) as well.

elainekamlley:
@jbresner Not seeing the svgs did you mean to attach them to the comment?

acdha:
“beefy” meaning thicker? The loc.gov arrows are slightly taller but a bit narrower. The SVG should be small enough just to put in as a code block since it's just text.

jbresner:
Yeah, sorry, in a meeting while trying to get the SVG's uploaded but beefy meant thicker.

acdha:
As of df6258c the indicators are circles but use the same colors:
screen shot 2018-12-11 at 14 50 17

jbresner:
Here are the SVGs (in a zipped file since GitHub doesn't support SVG upload)
[crowd-carousel-graphics.zip](https://github.com/LibraryOfCongress/concordia/files/2669143/crowd-carousel-graphics.zip)

acdha:
Cool - let me run them through SVGO and I'll put them into the CSS

acdha:
@jbresner Here's what it looks like after https://github.com/LibraryOfCongress/concordia/pull/708/commits/15a8451d15bc35269fe33a862dc0b77a6fd3ee92 and https://github.com/LibraryOfCongress/concordia/pull/708/commits/660a05542319fe1fa5e55349cd5cebf894540d4f:

screen shot 2018-12-11 at 15 15 26
screen shot 2018-12-11 at 15 15 38

jbresner:
@acdha - per our discussion, here are the SVGs with the correct hex value for orange, which is #F05129 (it's also not web accessible from a contrast perspective but we'll deal with that later I suspect):

[crowd-carousel-graphics.zip](https://github.com/LibraryOfCongress/concordia/files/2669716/crowd-carousel-graphics.zip)

acdha:
With the updated SVGs and the color changes from #713 here's what it looks like, with a regular and hover state:

screenshot_2018-12-12 crowd home

screen shot 2018-12-12 at 09 50 00-fullpage

PR #709, opened 2018-12-11
Campaign sort ordering from lowest to highest

rstorey:
This makes campaign ordering field go in ascending order (same as related links).


rstorey:
This could also be merged against release if we want it to be part of that set of changes.

PR #715, opened 2018-12-13
Scale homepage carousel controls down for mobile

acdha:
Two of the titles added in #680 were wide enough that the homepage
carousel controls could overlap the text. This commit changes the sizing
rules for screens below the small breakpoint to keep the arrows centered
over the image rather than the combined image + overlay.

## Before

“By the People” is fine on most devices:
simulator screen shot - iphone xr - 2018-12-13 at 10 35 30

… but “Letters to Lincoln” is not:

simulator screen shot - iphone xr - 2018-12-13 at 10 35 32

## After

simulator screen shot - iphone xr - 2018-12-13 at 11 12 16

### DIFF
--- a/concordia/static/css/base.css
+++ b/concordia/static/css/base.css
@@ -719,6 +719,23 @@ ul.nav-secondary {
     .homepage-activity p {
         font-size: 85%;
     }
+
+    .view-homepage .carousel-control-next,
+    .view-homepage .carousel-control-prev {
+        /*
+            the background images are all 1200x480 sized with width: 100% so
+            we'll set our height to maintain that aspect ratio so the arrows
+            will be centered on the image rather than the total carousel height
+            which includes the overlay
+        */
+        height: calc(100vw / (1200 / 480));
+    }
+
+    .view-homepage .carousel-control-next-icon,
+    .view-homepage .carousel-control-prev-icon {
+        height: 36.5px;
+        width: 22px;
+    }
 }
 
 @media screen and (min-width: 720px) {
--- a/concordia/static/css/base.css
+++ b/concordia/static/css/base.css
@@ -625,10 +625,6 @@ ul.nav-secondary {
     line-height: 125%;
 }
 
-.view-homepage .carousel-overlay h1 {
-    text-transform: uppercase;
-}
-
 .view-homepage .carousel-indicators li {
     width: 12px;
     height: 12px;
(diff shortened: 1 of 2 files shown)


jbresner:
@acdha - here's a quick mockup for the iPhone (375px) with some of the specs I used. Also, change the title text from uppercase to title case (that should save some space as well).
crowd-home-375px

jbresner:
At the smaller breakpoints, I'd recommend keeping the arrows at a fixed distance from the image and NOT centering it horizontally with all the text (Title, Description, CTA button) because as those vary in overall height, the left/right arrows end up bouncing up and down sometimes, which would be less than desirable for a user navigating.

acdha:
Currently it's not aligned at all with the text — it's centered vertically over the image. I should be able to change that to stay consistently below the image but that's complicated by the viewport-specific sizing.

acdha:
@jbresner Here are some screenshots at various sizes from a Samsung S9 screen through a tablet:

screen shot 2018-12-13 at 13 16 20
screen shot 2018-12-13 at 13 16 27
screen shot 2018-12-13 at 13 16 31
screen shot 2018-12-13 at 13 16 35
screen shot 2018-12-13 at 13 18 18
screen shot 2018-12-13 at 13 18 29

jbresner:
How much effort does it take to create a media query for the left/right arrows so the mobile versions (where the arrows aren't over the image) are solid black instead of outline? I know it's not an ideal way to have the two sets of arrows but because they're moving from images to a known solid color background, I'd rather they be as noticeable as possible in both cases.

acdha:
We already have one so it's really just a question of the size of the extra SVG image(s)

elainekamlley:
@acdha the last item you need is the smaller svg arrow correct?

acdha:
The images scale down already (they're SVG so this looks fine) so it's just a question of whether we want separate images to flip the color or do something like put a CSS `filter: invert(1)` at the small breakpoint (IE11 doesn't support it but no phones run it either).

acdha:
With https://github.com/LibraryOfCongress/concordia/pull/715/commits/a2c4748e492f0eb9da5c5d80d3c6c956f564976e it looks like this on mobile:

simulator screen shot - iphone xr - 2018-12-14 at 12 54 55
simulator screen shot - iphone xr - 2018-12-14 at 12 55 01

PR #716, opened 2018-12-13
Dependency updates: Django 2.1 and pending improvements

acdha:
Time for some early spring cleaning:

* [x] Upgrade to Django 2.1
* [x] Add attribution for `EmailOrUsernameModelBackend` from https://stackoverflow.com/a/35836674/59984
* [x] Update admin customizations to use the new extension mechanism referenced in https://github.com/LibraryOfCongress/concordia/blob/c5ec4ba43920c18c43cbdcaf340fd8862092bd9b/concordia/admin/views.py#L30-L34


acdha:
@rstorey https://github.com/LibraryOfCongress/concordia/pull/716/commits/b45faa1ee4732d985151fc2b4e37e0aab5e8e543 suggests we should have some test data for that with a known input:output sanity check

acdha:
@rstorey I've been testing this branch more and it seems to behave identically to the existing site. Is there anything else you'd like to confirm works first?

acdha:
I updated with the boto updates which came out recently

PR #731, opened 2018-12-27
Library-specific attribution for text destined for loc.gov

### LINKED ISSUE #717: Include attribution in exported text files
The plain text files, one per page / asset, which are exported in BagIt form need to have attributions at the bottom of the text.

Each text file should contain the attribution "Transcribed and reviewed by volunteers participating in the By The People project at crowd.loc.gov." at the bottom.

rstorey:
This isn't the ideal way I'd like to achieve this goal, but given the constraints it's the path of least resistance.

Closes #717

### DIFF
--- a/exporter/views.py
+++ b/exporter/views.py
@@ -75,8 +75,9 @@ def do_bagit_export(assets, export_base_dir, export_filename_base):
         text_output_path = os.path.join(dest_path, "%s.txt" % asset_filename)
         with open(text_output_path, "w") as f:
             f.write(asset.latest_transcription or "")
-            f.write("\n\n")
-            f.write(settings.ATTRIBUTION_TEXT)
+            if settings.ATTRIBUTION_TEXT:
+                f.write("\n\n")
+                f.write(settings.ATTRIBUTION_TEXT)
 
     # Turn Structure into bagit format
     bagit.make_bag(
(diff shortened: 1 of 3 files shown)


exporter/views.py
```diff
@@ -75,6 +75,9 @@ def do_bagit_export(assets, export_base_dir, export_filename_base):
         text_output_path = os.path.join(dest_path, "%s.txt" % asset_filename)
         with open(text_output_path, "w") as f:
             f.write(asset.latest_transcription or "")
```
acdha: Is the idea that we would want an empty file for the this-page-is-blank transcriptions?

exporter/views.py
```diff
@@ -75,6 +75,9 @@ def do_bagit_export(assets, export_base_dir, export_filename_base):
...
+            if settings.ATTRIBUTION_TEXT:
```
acdha: I'm wondering whether the setting location should be tweaked to make this opt-in rather than opt-out:

```suggestion
            if getattr(settings, "ATTRIBUTION_TEXT"):
```

… and then moving our value to `settings_prod.py`

exporter/views.py
rstorey: Good question. I'm not sure what the desired behavior is.

exporter/views.py
rstorey: Yeah it should be this way.

acdha: Please note that I left out the third getattr parameter – since it's not `dict.get` that should have been:

```suggestion
            if getattr(settings, "ATTRIBUTION_TEXT", None):
```

rstorey: I used hasattr instead of getattr - since we're not saving the result, it seems more appropriate.

PR #732, opened 2018-12-27
Export one page per item

### LINKED ISSUE #718: Include one-page text per item in BagIt export
In the BagIt export which feeds transcriptions into loc.gov, we want to provide one plain text file per item which has all the transcription text from all pages.

The text file should contain all pages in the item and be named with the item ID. There should be a single attribution at the bottom of all the text.

rstorey:
This version of BagIt export includes one text file per item that contains all assets in order for that item.

There are also (very basic) PDFs made of every text file in the bag.

Closes #718

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -41,6 +41,7 @@ pylibmc = "*"
 cfn-lint = "*"
 django-celery-beat = "*"
 django-passwords = "*"
+kombu = ">=4.2.2<4.3.0"
 
 [dev-packages]
 invoke = "*"
--- a/Pipfile
+++ b/Pipfile
@@ -41,7 +41,6 @@ pylibmc = "*"
 cfn-lint = "*"
 django-celery-beat = "*"
 django-passwords = "*"
-pdfkit = "*"
 
 [dev-packages]
 invoke = "*"
--- a/Pipfile
+++ b/Pipfile
@@ -41,6 +41,7 @@ pylibmc = "*"
 cfn-lint = "*"
 django-celery-beat = "*"
 django-passwords = "*"
+pdfkit = "*"
 
 [dev-packages]
 invoke = "*"
--- a/exporter/views.py
+++ b/exporter/views.py
@@ -67,14 +67,22 @@ def do_bagit_export(assets, export_base_dir, export_filename_base):
 
         asset_id = asset_id.replace(":", "/")
         asset_path, asset_filename = os.path.split(asset_id)
+        item_path, item_filename = os.path.split(asset_path)
 
-        dest_path = os.path.join(export_base_dir, asset_path)
-        os.makedirs(dest_path, exist_ok=True)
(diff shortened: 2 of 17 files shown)


acdha:
I'm hesitant to do this: it's adding a pretty big dependency which will get frequent updates, including security patches, just to transform content for a particular downstream service. I'd prefer either not to be in this business at all or to use something like Lambda to do it on-demand so we can keep that code separate, especially for the inevitable customization requests.

exporter/views.py
```diff
@@ -67,14 +69,40 @@ def do_bagit_export(assets, export_base_dir, export_filename_base):
...
+                this_text_file = os.path.join(dirpath, each_text_file)
+                with open(this_text_file, "a") as f:
+                    f.write("\n\n")
+                    f.write(settings.ATTRIBUTION_TEXT)
+
+    # Make PDFs of all pages
+    for dirpath, dirnames, filenames in os.walk(export_base_dir, topdown=False):
+        for each_text_file in filenames:
+            this_pdf_file = each_text_file.replace(".txt", ".pdf")
+            pdfkit.from_file(
```
acdha: Since the attribution text isn't part of the content, should we be using the `.txt` files or only including the actual transcription text and having pdfkit store it as document-level metadata or a footer?

rstorey:
I removed the PDF stuff. Now this PR adds a text file for each item with all the asset pages appended in sequence order.

exporter/views.py
rstorey: I think this is what we'll want to do eventually when we can implement the PDF creation bit as a separate service.

Pipfile
```diff
@@ -41,6 +41,7 @@ pylibmc = "*"
 cfn-lint = "*"
 django-celery-beat = "*"
 django-passwords = "*"
+pdfkit = "*"
```
acdha: We can revert this now?

exporter/views.py
```diff
@@ -67,17 +67,31 @@ def do_bagit_export(assets, export_base_dir, export_filename_base):
...
             f.write(asset.latest_transcription or "")
-            if hasattr(settings, "ATTRIBUTION_TEXT"):
-                f.write("\n\n")
-                f.write(settings.ATTRIBUTION_TEXT)
+            f.write("\n\n")
+
+    # Add attributions to the end of all files found under asset_dest_path
+    if hasattr(settings, "ATTRIBUTION_TEXT"):
+        for dirpath, dirnames, filenames in os.walk(export_base_dir, topdown=False):
+            for each_text_file in filenames:
```
acdha: Should this be filtered to avoid potential conflicts in the future if we ever start adding something which is not a text file?

```suggestion
            for each_text_file in (i for i in filenames if i.endswith(".txt")):
```

rstorey: Yes

PR #756, opened 2019-01-10
Add required env variables to app in docker-compose development setup

rstorey:


concordia/settings_template.py
```diff
@@ -127,12 +127,22 @@
...
     }
-}
+
+    SESSION_ENGINE = "django.contrib.sessions.backends.cached_db"
+
+else:
+
+    CACHES = {"default": {"BACKEND": "django.core.cache.backends.locmem.LocMemCache"}}
+
+    SESSION_ENGINE = "django.contrib.sessions.backends.cache"
```
acdha: I'm wondering whether this should be the database backend so you don't need to login every time you restart the dev server

docker-compose.yml
```diff
@@ -30,6 +30,10 @@ services:
             WAIT_HOSTS: db:5432
             POSTGRESQL_HOST: db
             POSTGRESQL_PW: ${POSTGRESQL_PW}
+            SENTRY_BACKEND_DSN: https://34db819263f34c28809da045f841f045@crowd-sentry.loc.gov/2
+            SENTRY_FRONTEND_DSN: https://48ec47f15e484502a29879e40ed2e0c3@crowd-sentry.loc.gov/3
```
acdha: I'm wondering whether these should be set at all or simply blank so we don't log every local typo to the Sentry server

rstorey: Well, if we want to keep the `raven test` step in the dockerfile... then we gotta make that conditional on these being set, I suppose. Otherwise the app container exits.

concordia/settings_template.py
rstorey: Yes, let's do it that way

docker-compose.yml
rstorey: This should be resolved now

concordia/settings_template.py
rstorey: This is done

PR #758, opened 2019-01-11
WIP: Deploy feature branch to crowd-test

rstorey:


rstorey:
> This generally looks good to me. The one question I had is whether we should be taking the time now to convert the hard-coded names for security groups, VPC IDs, etc. into output references or not. I'm slightly hesitant to do so since that adds more exposure to CloudFormation bugs but it would reduce the number of updates.

I was sorta thinking that too. Like, what if we decide we want to push feature branches to crowd-dev or something, what would the changes need to be, what should be parameterized, etc. But, I figured I'd address that if / when we decide to do this sort of thing again in a different VPC.

PR #760, opened 2019-01-15
Container build script changes

rstorey:


build_containers.sh
```diff
@@ -1,8 +1,11 @@
 #!/bin/bash
 
-
 set -euox pipefail
 
+BUILD_ALL=${BUILD_ALL:=0}
+BUILD_NUMBER=${BUILD_NUMBER:=1}
```
acdha: Is it reasonable to have a default here or should we fail if it's not defined? I was wondering whether that would be something where we'd want to try to avoid conflicts — e.g. maybe using the current Unix timestamp if it's not set:

```suggestion
BUILD_NUMBER=${BUILD_NUMBER:=$(date +%s)}
```

build_containers.sh
```diff
@@ -1,8 +1,11 @@
...
+TAG=${TAG:-test}
```
acdha: Similarly, I was wondering whether this should default to `test` or something like `$(git branch)` to avoid conflicts with the default values.

build_containers.sh
rstorey: This is normally set by Jenkins, but if you try to use the script locally it needs to be something. I'd prefer it to be an integer across the board since it gets used in things like cloudformation resource names that have character limits. I'll put a comment in there.

build_containers.sh
rstorey: Again, normally this would be set by Jenkins and would either be latest, release, or the git branch name. Perhaps latest would be a more reasonable default - there's always going to be a latest docker image in the ECR.

build_containers.sh
acdha: A timestamp will satisfy the integer goal but we could have conflicts if it's started multiple times at the same second which is unlikely except in a CI parallel build scenario.

build_containers.sh
acdha: I guess I'm wondering whether this means we should have defaults or just fail outright if those aren't set since we're only supporting the Jenkins-managed environment.

PR #763, opened 2019-01-15
Automate data load for feature branches

rstorey:
Refs #765

### DIFF
--- a/cloudformation/create_bastion_hosts.sh
+++ b/cloudformation/create_bastion_hosts.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
 
-export ENV_NAME=dev
+export ENV_NAME=${ENV_NAME:-dev}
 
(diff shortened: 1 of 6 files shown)


rstorey:
I'm done with this for now, unless there are changes you'd like made @acdha The most recent test was a success

rstorey:
> Are the changes for bastion 1/2 necessary now that there's a the dataloadhost declaration? I'm wondering whether we should just have that confined to that purpose, set `InstanceInitiatedShutdownBehavior=terminate`, and add a `shutdown -h now` call to the end of the user-data script.

I still use the bastion host template for loading DB dumps to dev / test / stage and get the DB dump from prod. It's not needed for loading data to feature branch deployments. But the changes I made there make it much simpler to dump or restore in the four main environments. I will add the shutdown stuff.

rstorey:
I need to merge this one before we can use the feature branch deployment. I'm happy to make changes even after the stuff is merged.

PR #766, opened 2019-01-17
Reopen asset for edit after submitting

rstorey:
Refs #690


concordia/tests/test_view.py
```diff
@@ -730,7 +730,25 @@ def test_transcription_disallow_self_review(self):
...
+
+        t1 = Transcription(asset=asset, user=self.user, text="test", submitted=now())
+        t1.full_clean()
+        t1.save()
+
+        resp = self.client.post(
+            reverse("review-transcription", args=(t1.pk,)), data={"action": "reject"}
+        )
+        self.assertValidJSON(resp, expected_status=200)
+        self.assertEqual(
```
acdha: This looks pretty good, do you think it's worth adding a check that rejected-by is the user?

Issue #749, opened 2019-01-08
As a reviewer, when I accept a completed transcription I want the option to keep reviewing


elainekamlley:
**Is your feature request related to a problem? Please describe.**
The low rate of acceptance numbers could be a sign of bottle neck in the process. When a user clicks accept the only option is to transcribe a new page or tag, essentially pulling the user out of review.

**Additional context**
By creating a review track, we want to improve the rate of review to completion and overall awareness of the ability to review. The health of the program relies on the volunteers completing all tasks. 

**Acceptance criteria:** 
- [x] As a reviewer, when I accept a completed transcription and see the message, I want to continue doing a similar task when I click `Find a new page` 
- [x] As a reviewer, I am brought to a page that I did not submit for review
- [x] Create a new message that aligns with the review track

acdha:
Just to record the standup conversation, I think we should do #753 now and start working with @jbresner to think about how incorporating the concept of a mode preference will ripple through the UI.

rstorey:
How should the buttons in this dialog be arranged? What should they say? Here's a screenshot of what it currently looks like in my branch with placeholder text:

<img width="638" alt="screen shot 2019-01-17 at 2 21 20 pm" src="https://user-images.githubusercontent.com/5270415/51343034-662fbd80-1a63-11e9-97ee-b1d6c139a80f.png">

@lalgee @jbresner @VVH

VVH:
Based on discussion in backlog: 'Review another page' 'Transcribe another page' 'Add tags' in that order

rstorey:
![screen shot 2019-01-18 at 5 09 46 pm](https://user-images.githubusercontent.com/5270415/51415393-ab2b2100-1b43-11e9-93f7-6a958fea88bf.png)

rstorey:
![screen shot 2019-01-18 at 5 14 54 pm](https://user-images.githubusercontent.com/5270415/51415602-5a67f800-1b44-11e9-8dbe-4e877adff6b1.png)

colefemeno:
@rstorey I prefer the stacked buttons! The button language looks good. 
For the rest of the dialogue, can we try: 
"Nice Job!"
Thanks for your help - we've saved your decision. 
You can stay on this page to add tags or select your next page to review or transcribe."
 
@lalgee @VVH @jbresner  I've specifically left out the word "review" in the second line so we can use this message more flexibly for both saves of edits and reviews

rstorey:
[...]

PR #767, opened 2019-01-17
Continue reviewing after accept

rstorey:
Refs #749


concordia/views.py
```diff
@@ -1118,6 +1128,73 @@ def reserve_asset_transcription(request, *, asset_pk):
...
+    potential_assets = Asset.objects.select_for_update(skip_locked=True, of=("self",))
+    potential_assets = potential_assets.filter(
+        item__project__campaign=campaign,
+        item__project__published=True,
+        item__published=True,
+        published=True,
+    )
+    potential_assets = potential_assets.filter(
+        transcription_status=TranscriptionStatus.SUBMITTED,
+        assettranscriptionreservation=None,
```
acdha: The `assettranscriptionreservation` shouldn't be relevant for review since that's only used when editing.

concordia/views.py
```diff
@@ -1164,7 +1241,7 @@ def redirect_to_next_transcribable_asset(request, *, campaign_slug):
         next_asset=Case(
             When(pk__gt=asset_id, then=1), default=0, output_field=IntegerField()
         ),
-    ).order_by("-next_asset", "-unstarted", "-same_project", "-same_item", "sequence")
+    ).order_by("-next_asset", "-unstarted", "-same_item", "-same_project", "sequence")
```
acdha: This is causing the test to fail because it no longer prioritizes staying in the same project ahead of staying in the same campaign.

rstorey:
My effort to de-dupe the code was fruitless. It caused the never_cache decorator to fail for a reason that I'm not quite sure of. In any case, although the two methods are very similar, they aren't exactly the same and there seems to be some kind of reason why never_cache wants the redirect call to be in the main view method. It was worth a try.

acdha:
`never_cache` shouldn't be preventing that – it just adds a couple of headers on the returned `Response` object so it should be fine as long as we consistently do so.

rstorey:
But the returned `Response` object isn't available in the secondary method for some reason. It seems like it should be, but it isn't.

concordia/views.py
```diff
@@ -1198,9 +1164,26 @@ def redirect_to_next_reviewable_asset(request, *, campaign_slug):
         ),
     ).order_by("-next_asset", "-same_project", "-same_item", "sequence")
 
-    redirect_to_next_asset(
```
acdha: It looks like this should be `return redirect_to_next_asset(`

concordia/views.py
```diff
@@ -1251,6 +1234,27 @@ def redirect_to_next_transcribable_asset(request, *, campaign_slug):
         ),
     ).order_by("-next_asset", "-unstarted", "-same_project", "-same_item", "sequence")
 
-    redirect_to_next_asset(
```
acdha: Also `return` here

rstorey: D'oh

PR #773, opened 2019-01-18
ARIA updates

acdha:
This pull request makes several changes which improved navigation in Safari using VoiceOver. This might be a good chance to test the feature-branch deployment option, too, since that would allow real-world testing using tools such as JAWS.

## Summary of changes

* Pagination adds the `aria-current=page` attribute for the current page number element.
* Pagination sets `aria-hidden=true` on the disabled elements used to visually represent the range of omitted pages between the displayed page window and the first/last page links.
* The login form set `aria-role=dialog` on the form itself but the [HTML spec only allows a couple of roles for `form` elements](https://w3c.github.io/html/sec-forms.html#the-form-element) so that has been hoisted to the container.

### DIFF
--- a/concordia/templates/base.html
+++ b/concordia/templates/base.html
@@ -90,16 +90,18 @@
             </nav>
         </header>
     {% endblock site-header %}
+
+    {% block breadcrumbs-container %}
+        <nav aria-label="breadcrumb">
+            <ol class="breadcrumb bg-offwhite my-0">
+                <li class="breadcrumb-item"><a class="primary-text" href="/">Home</a></li>
+                {% block breadcrumbs %}{% endblock breadcrumbs %}
+            </ol>
+        </nav>
+    {% endblock breadcrumbs-container %}
+
     {% block site-main %}
         <main class="{% block extra_main_classes %}{% endblock %} d-print-block">
-            {% block breadcrumbs-container %}
-                <nav aria-label="breadcrumb">
-                    <ol class="breadcrumb bg-offwhite my-0">
-                        <li class="breadcrumb-item"><a class="primary-text" href="/">Home</a></li>
-                        {% block breadcrumbs %}{% endblock breadcrumbs %}
-                    </ol>
-                </nav>
-            {% endblock breadcrumbs-container %}
 
             {% block messages-container %}
                 <div id="messages" class="row" hidden>
--- a/concordia/templates/transcriptions/asset_detail.html
+++ b/concordia/templates/transcriptions/asset_detail.html
@@ -457,23 +457,29 @@
             </div>
         </div>
     </div>
-    <div id="captcha-modal" class="modal" tabindex="-1" role="dialog">
+    <div id="captcha-modal" class="modal" tabindex="-1" role="alertdialog" aria-labeledby="captcha-modal-title" aria-describedby="captcha-modal-description">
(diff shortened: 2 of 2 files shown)


rstorey:
Spoke too soon on using the feature branch deployment stuff. When this branch has the stuff that was just now merged into master, then we'll be able to deploy it and test on a LIB workstation.

zuhairmahd:
I look forward to trying it out when it’s ready.

rstorey:
https://aria-updates.crowd-test.loc.gov/ is ready

acdha:
@zuhairmahd I made one change in https://github.com/LibraryOfCongress/concordia/pull/773/commits/a6e5aec851ffa8b01ff3728757311766bb2790fb which moves the breadcrumbs out of the main element to improve the initial experience landing on the main landmark. This will require a slight tweak to the CSS (we currently set the minimum height for the page based on a formula which will need to be adjusted) if you think this is worth doing.

acdha:
Based on our conversation earlier, I'm also wondering whether we should set `autofocus` on the text field for the transcription page so it would always load that page with the focus set to the transcribed text, if any. It seems like that would be good for completed pages but not other pages since there wouldn't be an explanation for why the value is empty or incomplete which seems like this should be a separate ticket.

zuhairmahd:
@acdha Yes, I think it would be worth trying setting up the autofocus to go to the transcription box, though I'd like to test it first to make sure it doesn't cause unintended behaviours in NVDA and JAWS.

acdha:
Since this is an area of ongoing work, @rstorey, how does it sound to merge this to master at https://github.com/LibraryOfCongress/concordia/pull/773/commits/cd6c00e0a115a00c7d9392e81bebdbf956802c1d (i.e. excluding my most recent commit until I finish the CSS tweaks) but leave the branch & fargate target open so @zuhairmahd and I/you/@klee72 can make additional improvements with as much testing as needed.

acdha:
@zuhairmahd I was initially just going to make the change to [remove an invalid ARIA role on a form](https://github.com/LibraryOfCongress/concordia/pull/773/commits/a74d6293cf4e3ef6bf5dd20ce9b944cf9cc8a79c) but figured that was as good an opportunity as any to practice making the dialog more accessible. [The second commit](https://github.com/LibraryOfCongress/concordia/pull/773/commits/7a474d777cd0ec0580097da0d06a5d1bca9a22ca) does that. It's somewhat moot because someone who can't see the CAPTCHA challenge also won't be able to transcribe the image but I figured it's good to confirm that this is a good convention for improving the alert dialog experience and it is a good reminder that we will need to add an audio version of the CAPTCHA the first time we have an activity which someone could participate in using a screen reader (this particular dialog is only displayed when you attempt to save a transcription without being logged in).

rstorey:
> Since this is an area of ongoing work, @rstorey, how does it sound to merge this to master at [cd6c00e](https://github.com/LibraryOfCongress/concordia/commit/cd6c00e0a115a00c7d9392e81bebdbf956802c1d) (i.e. excluding my most recent commit until I finish the CSS tweaks) but leave the branch & fargate target open so @zuhairmahd and I/you/@klee72 can make additional improvements with as much testing as needed.

Yes, let's merge it at the point where I merged master into the branch, and keep working in the same branch.

acdha:
I just re-based this against master. I'm thinking we can go ahead and merge it now so there's nothing pending & pick up another round when we have time to focus on it.

PR #774, opened 2019-01-18
pre-commit: enable more hooks

acdha:
This enables more of the built-in pre-commit hooks and enables auto-fixing for JSON & mixed line 
endings.


asottile:
looking at the travis output you probably want:

```yaml
        args: [--allow-missing-credentials]
```

PR #775, opened 2019-01-18
Simplify running Docker with DEBUG=true

acdha:
This changes `settings_docker` to set the `DEBUG` setting based on the environment and `docker-compose` will now pass that value through.

### DIFF
--- a/concordia/settings_docker.py
+++ b/concordia/settings_docker.py
@@ -13,6 +13,7 @@ LOGGING["handlers"]["celery"]["filename"] = "./logs/concordia-celery.log"
 LOGGING["loggers"]["django"]["level"] = "INFO"
 LOGGING["loggers"]["celery"]["level"] = "INFO"
 
+DEBUG = os.getenv("DEBUG", "").lower() == "true"
 
 DJANGO_SECRET_KEY = os.getenv("DJANGO_SECRET_KEY", get_random_secret_key())
 
--- a/concordia/views.py
+++ b/concordia/views.py
@@ -88,7 +88,11 @@ def default_cache_control(view_function):
 
 @never_cache
 def healthz(request):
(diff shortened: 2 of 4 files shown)


acdha:
The change I'm wondering about is this:

```diff
diff --git a/concordia/views.py b/concordia/views.py
index 9d156ab..f390ac9 100644
--- a/concordia/views.py
+++ b/concordia/views.py
@@ -88,7 +88,11 @@ def default_cache_control(view_function):
 
 @never_cache
 def healthz(request):
-    status = {"current_time": time.time(), "load_average": os.getloadavg()}
+    status = {
+        "current_time": time.time(),
+        "load_average": os.getloadavg(),
+        "debug": settings.DEBUG,
+    }
 
     # We don't want to query a large table but we do want to hit the database
     # at last once:
```

It's not a big difference but I'm not sure we normally need to expose this.

rstorey:
> The change I'm wondering about is this:
> 
> ```diff
> diff --git a/concordia/views.py b/concordia/views.py
> index 9d156ab..f390ac9 100644
> --- a/concordia/views.py
> +++ b/concordia/views.py
> @@ -88,7 +88,11 @@ def default_cache_control(view_function):
>  
>  @never_cache
>  def healthz(request):
> -    status = {"current_time": time.time(), "load_average": os.getloadavg()}
> +    status = {
> +        "current_time": time.time(),
> +        "load_average": os.getloadavg(),
> +        "debug": settings.DEBUG,
> +    }
>  
>      # We don't want to query a large table but we do want to hit the database
>      # at last once:
> ```
> 
> It's not a big difference but I'm not sure we normally need to expose this, although that could be used for continuous monitoring.

I think we should include this in the changes for monitoring purposes. And as a sanity check too.

PR #778, opened 2019-01-23
Campaign UI updates

klee72:
concordia/static/css/base.css
* Added class .h1 to match the font styling of a h1. this was needed so that we can use correct heading level `<h2>` in the markup for campaigns list. 
* Since both .btn-primary and .btn-secondary are already reserved, I temporarily created .btn-primary-alternative for orange button. I will work with Jaime for customizing global styles once we start using sass version of bootstrap.

concordia/templates/base.html
* Updated google font customize option for Roboto Slab so it displays the correct font weight (700) when using bold.

concordia/templates/transcriptions/campaign_list.html
* Updated the markup so it’s more meaningful, mainly to convert campaigns list into unordered list. And replaced `<h1>` heading level with `<h2>` for title of campaign list, and added `<h1>Campaigns</h1>` for page heading. 
* Also fine-tuned responsive design by using appropriate bootstrap classes.

### DIFF
--- a/concordia/static/css/base.css
+++ b/concordia/static/css/base.css
@@ -357,17 +357,6 @@ ul.nav-secondary {
     opacity: 0.3 !important;
 }
 
-.btn-primary-alternative {
-    background-color: var(--primary-color);
-    border-color: var(--primary-color);
-    color: var(--text-color-offwhite);
-}
-.btn-primary-alternative:hover {
-    background-color: var(--primary-color-hover);
-    border-color: var(--primary-color-hover);
-    color: var(--text-color-offwhite);
-}
-
 .btn-secondary {
     background-color: var(--secondary-color);
     border-color: var(--secondary-color);
--- a/concordia/static/css/base.css
+++ b/concordia/static/css/base.css
@@ -1,317 +1,319 @@
 :root {
-  --primary-color: #f05129;
-  --primary-color-hover: rgba(188, 57, 1, 1); /* #BC3901 */
-  --secondary-color: rgba(0, 167, 228, 1); /* #00A7E4 */
-  --secondary-color-hover: rgba(0, 35, 71, 1); /* #002347 */
-  --color-success: rgba(22, 201, 74, 1); /* #16C94A */
-  --color-danger: rgba(254, 12, 11, 1); /* #FE0C0B */
-  --color-warning: rgba(254, 153, 11, 1); /* #FE990B */
-  --color-info: rgba(2, 252, 216, 1); /* #02FCD8 */
-  --text-color-primary: rgba(38, 50, 56, 1); /* #263238 */
-  --text-color-offwhite: rgba(250, 250, 250, 1); /* #FAFAFA */
-  --text-nav-current: #f05129;
+    --primary-color: #f05129;
+    --primary-color-hover: rgba(188, 57, 1, 1); /* #BC3901 */
+    --secondary-color: rgba(0, 167, 228, 1); /* #00A7E4 */
+    --secondary-color-hover: rgba(0, 35, 71, 1); /* #002347 */
+    --color-success: rgba(22, 201, 74, 1); /* #16C94A */
+    --color-danger: rgba(254, 12, 11, 1); /* #FE0C0B */
+    --color-warning: rgba(254, 153, 11, 1); /* #FE990B */
+    --color-info: rgba(2, 252, 216, 1); /* #02FCD8 */
+    --text-color-primary: rgba(38, 50, 56, 1); /* #263238 */
+    --text-color-offwhite: rgba(250, 250, 250, 1); /* #FAFAFA */
+    --text-nav-current: #f05129;
 
-  --bg-offwhite: rgba(250, 250, 250, 1); /* #FAFAFA */
-  --bg-lightest-gray: rgba(245, 245, 245, 1); /* #F5F5F5 */
-  --bg-darkest-gray: rgba(51, 51, 51, 1); /* #333333 */
-  --bg-offblack: rgba(38, 50, 56, 1); /* #263238 */
-  --bg-dark-gray: rgba(102, 102, 102, 1); /* #666666 */
-  --bg-medium-gray: rgba(153, 153, 153, 1); /* #999999 */
-  --bg-footer: rgba(238, 238, 238, 1); /* #EEEEEE */
-  --bg-divider: rgba(189, 189, 189, 1); /* #BDBDBD */
-  --bg-card-title: rgba(51, 51, 51, 0.7); /* #333333 */
-  --bg-loading: rgba(51, 51, 51, 0.5); /* #333333 */
-  --bg-completed: #00618e;
-  --bg-submitted: #2e91bf;
-  --bg-in_progress: #80cef3;
-  --bg-not_started: #e9ecef;
+    --bg-offwhite: rgba(250, 250, 250, 1); /* #FAFAFA */
+    --bg-lightest-gray: rgba(245, 245, 245, 1); /* #F5F5F5 */
+    --bg-darkest-gray: rgba(51, 51, 51, 1); /* #333333 */
+    --bg-offblack: rgba(38, 50, 56, 1); /* #263238 */
+    --bg-dark-gray: rgba(102, 102, 102, 1); /* #666666 */
+    --bg-medium-gray: rgba(153, 153, 153, 1); /* #999999 */
+    --bg-footer: rgba(238, 238, 238, 1); /* #EEEEEE */
+    --bg-divider: rgba(189, 189, 189, 1); /* #BDBDBD */
+    --bg-card-title: rgba(51, 51, 51, 0.7); /* #333333 */
+    --bg-loading: rgba(51, 51, 51, 0.5); /* #333333 */
+    --bg-completed: #00618e;
+    --bg-submitted: #2e91bf;
+    --bg-in_progress: #80cef3;
+    --bg-not_started: #e9ecef;
 
-  --shadow-color: rgba(51, 51, 51, 0.8); /* #333333 */
-  --light-shadow-color: rgba(51, 51, 51, 0.4); /* #333333 */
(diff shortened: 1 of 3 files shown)


acdha:
I’ve been trying to include before and after screenshots in tickets like this for anyone who doesn’t have a dev server handy. You can copy and paste them into github descriptions or comments.

acdha:
The tests are failing because the CSS should be formatted using Prettier. Most editors have integration for doing that automatically and there are instructions for setting up pre-commit to check & format files:

https://github.com/LibraryOfCongress/concordia/blob/56defe5a932470bcdaac74c7d45d9aacb2ed5b24/docs/how-we-work.md

(We should copy that into the developers doc, too.)

concordia/static/css/base.css
```diff
@@ -60,7 +60,7 @@ main {
     min-height: var(--height-main);
 }
 
-h1 {
+h1, .h1 {
```
acdha: I'm not in love with a class matching a tag name since that doesn't explain what fraction of the semantics we want to match. One way to solve this would be to use a class name like `campaign-title`. We could also use multiple `<h1>` tags but that seems to be an area for concern with screen readers which don't follow the HTML5 spec completely and probably isn't worth the trouble of setting the ARIA attributes & testing.

concordia/static/css/base.css
```diff
@@ -357,6 +357,17 @@ ul.nav-secondary {
     opacity: 0.3 !important;
 }
 
+.btn-primary-alternative {
```
acdha: I'm wondering whether we should explain anything more about the intended use of this class. Also, do we need to worry about having the same `:focus`, `:disabled`, and `:active` classes?

https://github.com/twbs/bootstrap/blob/dfba588d0d1ea020634f1a8f2911c7222702b125/scss/_buttons.scss#L7-L45

This is definitely a good argument for adopting Bootstrap SCSS where we could set the base theme colors and then make this just a modifier class.

concordia/templates/transcriptions/campaign_list.html
```diff
@@ -13,28 +13,24 @@
...
-                            {{ campaign.short_description|linebreaksbr }}
-                        </div>
-                        <div class="col-12 col-md-4">
-                            <a href="{% url 'transcriptions:campaign-detail' campaign.slug %}">
-                            <img src="{{ MEDIA_URL }}{{ campaign.thumbnail_image }}" class="img-fluid" alt="{{ campaign.title }} image">
-                            </a>
-                        </div>
-                    </div>
+<div class="container py-3">
+    <h1 class="mb-5 sr-only">Campaigns</h1>
```
acdha: Does this really need a margin tag if it's only for screen readers?

acdha: I'm wondering whether we should hoist the `sr-only` H1 into the base template — just put something like this, just above the `main_content` block:

```django
        {% block page-title %}
            {% if title %}
                <h1 id="page-title" class="sr-only">{{ title }}</h1>
            {% endif %}
        {% endblock page-title%}
```

concordia/templates/transcriptions/campaign_list.html
```diff
@@ -13,28 +13,24 @@
...
+    <ul class="list-unstyled">
+        {% for campaign in campaigns %}
+        <li class="p-4 mb-1 bg-footer">
```
acdha: This block should be indented to show the loop structure

concordia/templates/transcriptions/campaign_list.html
```diff
@@ -13,28 +13,24 @@
...
+            <h2 class="h1 mb-3">{{ campaign.title }}</h2>
+            <div class="row">
+                <a class="col-md-5 order-md-2" href="{% url 'transcriptions:campaign-detail' campaign.slug %}">
+                    <p class="mb-2 text-center"><img src="{{ MEDIA_URL }}{{ campaign.thumbnail_image }}" class="img-fluid" alt="{{ campaign.title }} image"></p>
+                </a>
+                <div class="col-md">
+                    <p>{{ campaign.short_description|linebreaksbr }}</p>
+                    <a href="{% url 'transcriptions:campaign-detail' campaign.slug %}" class="btn btn-primary-alternative">View Projects</a>
```
acdha: I generally prefer to have the `class` attribute earlier so it's more visible but that can get unwieldy with a bunch of modifiers

concordia/templates/transcriptions/campaign_list.html
```diff
@@ -13,28 +13,24 @@
...
-                    </div>
+<div class="container py-3">
+    <h1 class="mb-5 sr-only">Campaigns</h1>
+    <ul class="list-unstyled">
+        {% for campaign in campaigns %}
+        <li class="p-4 mb-1 bg-footer">
+            <h2 class="h1 mb-3">{{ campaign.title }}</h2>
+            <div class="row">
+                <a class="col-md-5 order-md-2" href="{% url 'transcriptions:campaign-detail' campaign.slug %}">
+                    <p class="mb-2 text-center"><img src="{{ MEDIA_URL }}{{ campaign.thumbnail_image }}" class="img-fluid" alt="{{ campaign.title }} image"></p>
```
acdha: This is reminding me that we should get @zuhairmahd to review the alt text — I'm not sure how useful the automatic “image” suffix after the repeated title is but that's out of scope for this ticket.

concordia/templates/transcriptions/campaign_list.html
```diff
@@ -13,28 +13,24 @@
...
-                            </a>
-                            {{ campaign.short_description|linebreaksbr }}
-                        </div>
-                        <div class="col-12 col-md-4">
-                            <a href="{% url 'transcriptions:campaign-detail' campaign.slug %}">
-                            <img src="{{ MEDIA_URL }}{{ campaign.thumbnail_image }}" class="img-fluid" alt="{{ campaign.title }} image">
-                            </a>
-                        </div>
-                    </div>
+<div class="container py-3">
```
acdha: Is intentional not to follow the usual `container` / `row` hierarchy? It seems like we should wrap the list in a row so it would be a peer of the pagination block below.

concordia/static/css/base.css
klee72: I'm not creating anything new as .h1 classname is from Bootstrap: https://getbootstrap.com/docs/4.1/content/typography/#headings. We can definitely create a new classname but I thought our goal was to use bootstrap classes if the style is available. But yes I think we can discuss more about this once we start using sass version.

acdha: As long as this is supposed to be identical to an h1 in every way except outline level that’s fine.

concordia/templates/transcriptions/campaign_list.html
klee72: I think most pages have correct H1 title except for this page. I wanted to show Jaime what it looks like with the H1 displayed at the top. I'll work with Jaime to see if he wants it hidden or not, and will update this.

acdha: It’s been inconsistent, especially for things like the login or static pages. We have touched a couple but for accessibility I’m wondering whether we just make a separate task to review & consolidate it site-wide.

concordia/templates/transcriptions/campaign_list.html
klee72: It is intentional, as `row` is only needed when you're creating columns. However `row`s negative left/right margin is useful for a container inside `container` to expand to the edge. And since bootstrap grid gutter is based in px (30px) and their spacing is base in rem, lining things up is not easy. I'd like to try to fix this once we implement bootstrap scss.

concordia/templates/transcriptions/campaign_list.html
klee72: I agree. Since I think it's a "decorative" image, we can remove an anchor from the image and leave `alt` blank.

concordia/templates/transcriptions/campaign_list.html
klee72: This is fixed and committed.

concordia/templates/transcriptions/campaign_list.html
klee72: Removed margin class as Jamie wants H1 to be hidden for now.

concordia/templates/transcriptions/campaign_list.html
acdha: You're referring to the spacing using the padding class? My usual approach for this is to use the container/row with no margin/padding on those elements so the margin values will match (both use `px`) and then set a margin or padding on the the contents so they'll align consistently.

Either way, this seems like a great thing to defer until we reconsider our front-end toolchain.

acdha:
@klee72 how did you run Prettier for https://github.com/LibraryOfCongress/concordia/pull/778/commits/19ff1e63c857bf9f3d82e010e03f1bf114b12eb9? It looks like it didn't use our `.editorconfig` / `.prettierrc` settings so there were a ton of whitespace-only changes. I just ran it manually and now the diff is a lot smaller — the next time you run `git pull` it should pickup https://github.com/LibraryOfCongress/concordia/pull/778/commits/7f1372118d706dd5c4b3bae5d183876dc1901d79.

klee72:
> @klee72 how did you run Prettier for [19ff1e6](https://github.com/LibraryOfCongress/concordia/commit/19ff1e63c857bf9f3d82e010e03f1bf114b12eb9)? It looks like it didn't use our `.editorconfig` / `.prettierrc` settings so there were a ton of whitespace-only changes. I just ran it manually and now the diff is a lot smaller — the next time you run `git pull` it should pickup [7f13721](https://github.com/LibraryOfCongress/concordia/commit/7f1372118d706dd5c4b3bae5d183876dc1901d79).

@acdha I'm using Atom, and installed prettier and editorconfig. Where can I find our settings for these?

acdha:
The settings files are in the repo - as long as Atom is configured to honor the .editorconfig it should pick them up automatically (the most noticeable being indentation using 4 spaces)

PR #779, opened 2019-01-24
WIP: add accessibility docs

acdha:
This is a very rough draft to give us something concrete to discuss for how we want to structure this section of the documentation. There are also a few other documentation refactorings I made while finding a place for this to live.

The general goals I started with were to collect the goals independent of the specific tools and techniques since the latter will have far more details and churn as the technology evolves, and to favor linking to outside resources such as 18F's guides rather than attempting to duplicate them ourselves.

An unprioritized list of things I want to mention which aren't currently present:
* Specific tools such as Color Oracle for testing that class of visual impairment
* Adding a general reduced-motion CSS rule and making sure that's on the test plan
* Documenting the need to support reduced-mobility users by providing alternatives to mouse/touch gestures and multi-key shortcuts as much as possible
* Something about progressive enhancement / responsive design to avoid visual churn, delaying assistive technology due to a long page load event, and having graceful fallbacks where possible so an error doesn't produce a response which is more unhelpful than necessary
* Some general philosophical guidance — I would especially like to get people thinking about this as done in the same way that your garden is done rather than a one-time task or an overwhelming amount of work which has to be done completely the first time


docs/accessibility-goals.md
```diff
@@ -0,0 +1,35 @@
+# Accessibility Goals
+
+## Background Information
+
+-   Microsoft has a site with a number of resources detailing their “Inclusive Design” strategy: https://www.microsoft.com/design/inclusive/
+
+    Of note is the [Inclusive 101" PDF](https://download.microsoft.com/download/b/0/d/b0d4bf87-09ce-4417-8f28-d60703d672ed/inclusive_toolkit_manual_final.pdf) covering the range of factors which designs should take into account and common strategies.
```
rstorey: stray quotation mark?

acdha: Indeed…

rstorey:
Should we ticket any remaining items and merge what we have so far?

acdha:
Probably best just to merge it unless @elainekamlley has any to-do points first. There are more things to add but I keep wanting to turn it into a general Library document.

rstorey:
They're out for the next few days, so I'm going to resolve the conflicts and merge it.

PR #786, opened 2019-01-29
one step closer to fully automating data dump / restore

rstorey:


db_scripts/dump.sh
```diff
@@ -1,6 +1,17 @@
...
 
-# For AMI: sudo yum -y install https://download.postgresql.org/pub/repos/yum/9.6/redhat/rhel-6-x86_64/pgdg-ami201503-96-9.6-2.noarch.rpm
-#          sudo yum -y install postgresql96
+# export ENV_NAME=prod
+# aws cloudformation create-stack --region us-east-1 --stack-name $ENV_NAME-bastionhosts --template-url https://s3.amazonaws.com/crowd-deployment/infrastructure/bastion-hosts.yaml --parameters ParameterKey=EnvironmentName,ParameterValue=$ENV_NAME ParameterKey=KeyPairName,ParameterValue=rstorey@loc.gov --disable-rollback
+# aws cloudformation delete-stack --region us-east-1 --stack-name $ENV_NAME-bastionhosts
 
-pg_dump -Fc --clean --create --no-owner --no-acl -U concordia -h $POSTGRESQL_HOST concordia -f concordia.dmp
\ No newline at end of file
+export TODAY=20190129
```
acdha: We could make the date dynamic:

```suggestion
export TODAY=$(date +%Y%m%d)
```

db_scripts/dump.sh
```diff
@@ -1,6 +1,17 @@
...
+export POSTGRESQL_PW=${POSTGRESQL_PW:-password}
+export POSTGRESQL_HOST=${POSTGRESQL_HOST:-localhost}
+export DUMP_FILE=concordia.dmp
+
+echo "$POSTGRESQL_HOST:5432:*:concordia:$POSTGRESQL_PW" >> ~/.pgpass
```
acdha: Using `>>` will make this file grow each time it runs. Is this run anywhere other than the dedicated bastion host where we would need to care about having multiple accounts? Perhaps we could simply run this only when `~/.pgpass` does not exist?

db_scripts/dump.sh
```diff
@@ -1,6 +1,17 @@
...
+chmod 600 ~/.pgpass
+
+pg_dump -Fc --clean --create --no-owner --no-acl -U concordia -h "$POSTGRESQL_HOST" concordia -f "$DUMP_FILE"
+aws s3 cp "$DUMP_FILE" s3://crowd-deployment/database-dumps/concordia.$TODAY.dmp
```
acdha: Does it make sense to have some sort of guard here which either doesn't run if the environment isn't production or which puts the environment component into the S3 path?

db_scripts/restore.sh
```diff
@@ -1,16 +1,16 @@
...
+# export ENV_NAME=dev
+# aws cloudformation create-stack --region us-east-1 --stack-name $ENV_NAME-bastionhosts --template-url https://s3.amazonaws.com/crowd-deployment/infrastructure/bastion-hosts.yaml --parameters ParameterKey=EnvironmentName,ParameterValue=$ENV_NAME ParameterKey=KeyPairName,ParameterValue=rstorey@loc.gov --disable-rollback
+# aws cloudformation delete-stack --region us-east-1 --stack-name $ENV_NAME-bastionhosts
 
+export POSTGRESQL_PW=${POSTGRESQL_PW:-password}
 export POSTGRESQL_HOST=${POSTGRESQL_HOST:-localhost}
-export DUMP_FILE=concordia.dmp
+export DUMP_FILE=/concordia.dmp
+
+echo "$POSTGRESQL_HOST:5432:*:concordia:$POSTGRESQL_PW" >> ~/.pgpass
```
acdha: Same question re: `~/.pgpass` growing on each run

db_scripts/dump.sh
rstorey: Thanks for this, saved me some time looking up the date format documentation

db_scripts/dump.sh
rstorey: So the way I've been doing this is to delete the stack after I'm done using it, so it only exists for a few minutes while I'm running the command (maybe I can even change this to a lambda function). I changed this to use `>` so it'll overwrite instead of append, just in case there's a situation where I have to rerun the script on the same host.

acdha: Yeah, it's not a huge deal but I figured either clobbering it or doing a `if [ ! -f … ]` check avoids some potential confusion later.

build_containers.sh
```diff
@@ -5,8 +5,9 @@ set -eu -o pipefail
 BUILD_ALL=${BUILD_ALL:=0}
 BUILD_NUMBER=${BUILD_NUMBER:=1}
 TAG=${TAG:-test}
+PUSH=${PUSH:=1}
```
acdha: I like the idea of decoupling builds from pushing the container images but I'm wondering whether this should be a little more explanatory a name – something like `PUBLISH_CONTAINERS`? — since Git also uses “push” as a term.

db_scripts/dump.sh
```diff
@@ -1,6 +1,31 @@
 #!/bin/bash
 
-# For AMI: sudo yum -y install https://download.postgresql.org/pub/repos/yum/9.6/redhat/rhel-6-x86_64/pgdg-ami201503-96-9.6-2.noarch.rpm
-#          sudo yum -y install postgresql96
+set -eu -o pipefail
+
+ENV_NAME=prod
```
acdha: How would the `if` check below ever be false with this assignment?

db_scripts/dump.sh
```diff
@@ -1,6 +1,31 @@
...
+    chmod 600 ~/.pgpass
+
+    pg_dump -Fc --clean --create --no-owner --no-acl -U concordia -h "${POSTGRESQL_HOST}" concordia -f "${DUMP_FILE}"
+
+    if [ -s $DUMP_FILE ]; then
+        aws s3 cp "${DUMP_FILE}" "s3://crowd-deployment/database-dumps/concordia.${TODAY}.dmp"
+        aws s3 cp "${DUMP_FILE}" s3://crowd-deployment/database-dumps/concordia.latest.dmp
+    fi
+else
+    echo "This script should only be run in the production environment."
```
acdha: I don't think it matters that much but I was wondering whether this should just be bumped to the top of the script with a simple if-not `echo "This script…”; exit 1` block so the rest of it wouldn't need to be indented.

db_scripts/restore.sh
```diff
@@ -1,16 +1,24 @@
...
 
-# For AMI: sudo yum -y install https://download.postgresql.org/pub/repos/yum/9.6/redhat/rhel-6-x86_64/pgdg-ami201503-96-9.6-2.noarch.rpm
-#          sudo yum -y install postgresql96
+set -eu -o pipefail
+
+ENV_NAME=dev
+
+# aws cloudformation create-stack --region us-east-1 --stack-name $ENV_NAME-bastionhosts --template-url https://s3.amazonaws.com/crowd-deployment/infrastructure/bastion-hosts.yaml --parameters ParameterKey=EnvironmentName,ParameterValue=$ENV_NAME ParameterKey=KeyPairName,ParameterValue=rstorey@loc.gov --disable-rollback
+# aws cloudformation delete-stack --region us-east-1 --stack-name $ENV_NAME-bastionhosts
+if [ $ENV_NAME = "prod" ]; then
```
acdha: This seems like the same question – with the earlier ENV_NAME assignment it seems like this will never be true.

build_containers.sh
rstorey: Good idea. Done

db_scripts/dump.sh
rstorey: It wouldn't. In other contexts, ENV_NAME is allowed to be dev, test, stage or prod (such as in the restore script, and in the env vars for the fargate task definition). I'm using it here to form the secret ID and get the DB password out of secretsmanager. I'm not sure, other than hard-coding the RDS endpoint, how to confirm that we're dumping the prod database.

db_scripts/dump.sh
rstorey: Good idea

db_scripts/restore.sh
rstorey: That's a good idea. I don't know why that didn't occur to me. I'll implement that the next time I use these scripts / next week.

Issue #734, opened 2018-12-27
Print friendly view for transcriptions


rstorey:
**Is your feature request related to a problem? Please describe.**
There's no way to create nice-looking PDFs of transcribed assets / items. Only plain text is exported.

**Describe the solution you'd like**
I'd like a way to print completed transcription for an asset or an item with some basic formatting. The printable version could also be made into a PDF.

**Describe alternatives you've considered**
Considered converting plain text into plain PDFs directly, but it doesn't yield a nice result and it's a lot of dev time.

acdha:
Rough thought here: add either a print-friendly stylesheet for the asset detail page (which gives the latest transcription) or a separate text/plain alternative view, both of which are useful for working with the content and allow the path to generating PDFs to be as simple as [`chrome --print-to-pdf`](https://developers.google.com/web/updates/2017/04/headless-chrome#create_a_pdf_dom) or [using puppeteer](https://www.npmjs.com/package/puppeteer#usage) with no duplication of effort beyond the standard web stack.

klee72:
Is this the page in discussion? https://crowd.loc.gov/campaigns/letters-to-lincoln/1850-1857-death-and-birth-of-children-and-re-entry-to-politics/mal0043300/mal0043300-1/

klee72:
It looks like adding a print friendly stylesheet and utilizing bootstrap print display classes shouldn't be too difficult to do. If you guys are up for it, I can start looking into this.

acdha:
That seems reasonable to me — since the basic Bootstrap approach tends to result in printing being like a mobile device it should be mostly about hiding things like form inputs & labels which don't make sense (or, in the transcription's case, adjusting the borders & max-height to act like a text block).

rstorey:
Sounds good to me

rstorey:
This has been merged

PR #788, opened 2019-01-30
#734 Added print CSS and bootstrap print display utility classes to t…

klee72:
1) Since `textarea` cannot be resized to fit its content, it'd get cut off when content is longer than the specified height. To resolve this, I've added a dedicated container for printing a transcription `<div class="print-transcription-text" aria-hidden="true" style="display: none;">{{ transcription.text }}</div>` which will only be available when the page is printed. 
2) Firefox has some issues printing flexbox elements. A workaround is to replace `display: flex` with `display: block`,  which can be accomplished by adding `.d-print-block` class.
~https://bugzilla.mozilla.org/show_bug.cgi?id=258397 (duplicate of 939897)~
~https://bugzilla.mozilla.org/show_bug.cgi?id=1414253 (duplicate of 939897)~
https://stackoverflow.com/questions/45414152/css-flex-box-not-printing-all-pages-on-firefox
https://bugzilla.mozilla.org/show_bug.cgi?id=939897
3) In most cases, `.d-print-none` (hide) and `.d-print-block` (block element) were enough to style the print screen, a small block of CSS has been added for the ones that needed additional styling.
4) I’m having difficulty scaling the image in OpenSeadragon and keep the aspect ratio. @acdha and @rstorey: Can you please take a look?

### DIFF
--- a/concordia/static/css/base.css
+++ b/concordia/static/css/base.css
@@ -252,9 +252,9 @@ p.hero-secondary.offwhite-text {
     background-color: var(--bg-darkest-gray);
 
     background-image: linear-gradient(
-        rgba(250, 250, 250, 0.25),
-        rgba(38, 50, 56, 0.25)
-    ),
+            rgba(250, 250, 250, 0.25),
+            rgba(38, 50, 56, 0.25)
+        ),
         linear-gradient(rgba(51, 51, 51, 1), rgba(51, 51, 51, 1));
 }
 
@@ -264,9 +264,9 @@ p.hero-secondary.offwhite-text {
 
 .bg-img-placeholder:hover {
     background-image: linear-gradient(
-        rgba(252, 76, 2, 0.5),
-        rgba(252, 76, 2, 0.1)
-    ),
+            rgba(252, 76, 2, 0.5),
+            rgba(252, 76, 2, 0.1)
+        ),
         linear-gradient(rgba(189, 189, 189, 1), rgba(189, 189, 189, 1));
 }
 
--- a/concordia/static/css/base.css
+++ b/concordia/static/css/base.css
@@ -252,9 +252,9 @@ p.hero-secondary.offwhite-text {
     background-color: var(--bg-darkest-gray);
 
     background-image: linear-gradient(
-            rgba(250, 250, 250, 0.25),
-            rgba(38, 50, 56, 0.25)
-        ),
+        rgba(250, 250, 250, 0.25),
+        rgba(38, 50, 56, 0.25)
+    ),
         linear-gradient(rgba(51, 51, 51, 1), rgba(51, 51, 51, 1));
 }
 
@@ -264,9 +264,9 @@ p.hero-secondary.offwhite-text {
 
 .bg-img-placeholder:hover {
     background-image: linear-gradient(
-            rgba(252, 76, 2, 0.5),
-            rgba(252, 76, 2, 0.1)
-        ),
+        rgba(252, 76, 2, 0.5),
+        rgba(252, 76, 2, 0.1)
+    ),
         linear-gradient(rgba(189, 189, 189, 1), rgba(189, 189, 189, 1));
 }
 
@@ -1211,4 +1211,42 @@ div.related-links {
 /* Registration page */
 #registration-form-container > .col-md-6 {
     max-width: 30rem;
+}
+
+/* print */
+@media print {
+    @page {
+        margin: 0.75in;
+    }
+    body {
+        border-style: none;
+        background-color: transparent;
+        font-size: 14pt;
+    }
+    header .navbar {
+        padding: 0;
+        background-color: transparent;
+    }
+    #navigation-container {
+        border-style: none;
+        padding: 0;
+    }
+    #contribute-container {
+        border-style: none;
+    }
+    #contribute-container #viewer-column {
+        height: 300px;
+    }
+    #contribute-container .gutter-horizontal {
+        display: none;
+    }
+    #contribute-container #editor-column .tx-status-display {
+        font-size: 24px;
+        margin: 2rem 0;
+    }
+    #contribute-container #editor-column .print-transcription-text {
+        display: block !important;
+        white-space: pre-wrap;
+        color: #000;
+    }
 }
--- a/concordia/static/css/base.css
+++ b/concordia/static/css/base.css
@@ -1237,6 +1237,11 @@ div.related-links {
     #contribute-container #viewer-column {
         height: 300px;
     }
+
+    #asset-image .openseadragon-container {
+        display: none;
+    }
+
     #contribute-container .gutter-horizontal {
         display: none;
     }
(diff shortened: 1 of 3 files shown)


concordia/templates/transcriptions/asset_detail.html
```diff
@@ -146,17 +146,19 @@
...
 
                     <h2>Transcription</h2>
 
                     {% spaceless %}
-                        <textarea readonly class="form-control w-100 rounded flex-grow-1" name="text" id="transcription-input" placeholder="{% if transcription_status == 'not_started' or transcription_status == 'in_progress' %}Go ahead, start typing. You got this!{% else %}Nothing to transcribe{% endif %}" aria-label="Transcription input">
+                        <textarea readonly class="form-control w-100 rounded flex-grow-1 d-print-none" name="text" id="transcription-input" placeholder="{% if transcription_status == 'not_started' or transcription_status == 'in_progress' %}Go ahead, start typing. You got this!{% else %}Nothing to transcribe{% endif %}" aria-label="Transcription input">
                             {{ transcription.text }}
                         </textarea>
 
+                        <div class="print-transcription-text" aria-hidden="true" style="display: none;">{{ transcription.text }}</div>
```
acdha: Following the pattern of the rest of the page, shouldn't we use the Bootstrap classes rather than an inline `style`?

```suggestion
                        <div class="print-transcription-text d-none d-print-block" aria-hidden="true">{{ transcription.text }}</div>
```

klee72: Actually that was intentional to prevent it from displaying when CSS is disabled.

acdha: When would we support disabling CSS and in what context does that not also disable inline styles?

acdha:
Regarding the flexbox setup, do we have identified things which don't work in a current version of Firefox? If we can't use flexbox we need to add more CSS to do some sort of column layout for the image and text, and since that's exactly what flexbox would do I'd like to confirm that there are known non-resolved bugs.

acdha:
Regarding OpenSeadragon, I think the best option would probably be not using it since there's no benefit for a print context — something like this works except for the need to handle a better flexbox disable if we have to. Also, the 300px image height is always wrong for this so we should possibly just keep it simply and do something like `display:block` on both and setting some combination of `page-break-inside: avoid;` / `object-fit: contain` on the image.

``diff
diff --git a/concordia/static/css/base.css b/concordia/static/css/base.css
index bdc940d..b60d5f1 100644
--- a/concordia/static/css/base.css
+++ b/concordia/static/css/base.css
@@ -1237,6 +1237,11 @@ div.related-links {
     #contribute-container #viewer-column {
         height: 300px;
     }
+
+    #asset-image .openseadragon-container {
+        display: none;
+    }
+
     #contribute-container .gutter-horizontal {
         display: none;
     }
diff --git a/concordia/templates/transcriptions/asset_detail.html b/concordia/templates/transcriptions/asset_detail.html
index eb27e00..81ac245 100644
--- a/concordia/templates/transcriptions/asset_detail.html
+++ b/concordia/templates/transcriptions/asset_detail.html
@@ -118,7 +118,9 @@
                 </div>
             </div>
 
-            <div id="asset-image" class="h-100 bg-dark"></div>
+            <div id="asset-image" class="h-100 bg-dark">
+                Scanned image of the current content page
+            </div>
         </div>
 
         <div id="editor-column" class="d-flex flex-column flex-nowrap justify-content-between d-print-block">
```

**Comment by acdha** [comment]
A few other things which came up in the chrome print debugger: 

1. We should set `d-print-none` on `#asset-navigation` since the page numbers don't match the content pages and there's no guarantee of them being stable
2. We should set `d-print-none` on `.anonymous-only.help-text` or add a better selector for that since the login prompt isn't useful
3. The tags should probably just be reset to black-on-white text

File: concordia/templates/transcriptions/asset_detail.html (hunk 1 above)
Comment (klee72) [reply concordia/templates/transcriptions/asset_detail.html]: I meant an user turning off stylesheets from their browser. For testing, slow connection and etc reasons. You can test it with Firefox view > page style > no style. Probably only a few would have this setting, but wanted to be sure that we'll never display two identical text blocks in any circumstances.

**Comment by klee72** [comment]
> A few other things which came up in the chrome print debugger:
> 
> 1. We should set `d-print-none` on `#asset-navigation` since the page numbers don't match the content pages and there's no guarantee of them being stable
> 2. We should set `d-print-none` on `.anonymous-only.help-text` or add a better selector for that since the login prompt isn't useful
> 3. The tags should probably just be reset to black-on-white text

These all good points. I'll update soon.

File: concordia/templates/transcriptions/asset_detail.html (hunk 1 above)
Comment (acdha) [reply concordia/templates/transcriptions/asset_detail.html]: Slow connections will block the page rendering until the stylesheets have loaded

File: concordia/static/css/base.css [hunk 2]
```diff
@@ -1234,14 +1234,15 @@ div.related-links {
     #contribute-container {
         border-style: none;
     }
-    #contribute-container #viewer-column {
-        height: 300px;
+    #contribute-container #viewer-column .print-transcription-image {
+        height: 700px;
```
acdha: Most of our pages probably want to be larger than 700px. I'm wondering whether this should be something like a calc expression to be 100% minus the header height so you can get as much image as will fit on one page.

concordia/templates/transcriptions/asset_detail.html
```diff
@@ -117,12 +117,12 @@
                     </div>
                 </div>
             </div>
-
-            <div id="asset-image" class="h-100 bg-dark"></div>
+            <div id="asset-image" class="h-100 bg-dark d-print-none"></div>
+            <div class="print-transcription-image d-none d-print-block "><img class="img-fluid" alt="Scanned image of the current content page" src="{% asset_media_url asset %}"></div>
```
acdha: We. can remove the trailing space:

```suggestion
            <div class="print-transcription-image d-none d-print-block"><img class="img-fluid" alt="Scanned image of the current content page" src="{% asset_media_url asset %}"></div>
```

klee72:
> Regarding the flexbox setup, do we have identified things which don't work in a current version of Firefox? If we can't use flexbox we need to add more CSS to do some sort of column layout for the image and text, and since that's exactly what flexbox would do I'd like to confirm that there are known non-resolved bugs.

I just updated my comment, this is the bug: https://bugzilla.mozilla.org/show_bug.cgi?id=939897 which hasn't been resolved yet. I think it's okay to keep using flexbox since the issue is only for printing. I'll come back to this when I have a chance to go through other pages.

concordia/static/css/base.css
klee72: Ok, both `100%` and `100vh` height translates inconsistently on Chrome, Firefox and IE. I think what we can do is place the image at the very bottom of the page, and let it scale to the full width.

rstorey:
If you want to keep working, feel free...we can open a new PR and merge again before I cut the 0.4.17 release later today.

PR #797, opened 2019-01-31
Print friendly transcriptions

rstorey:


acdha:
I did a quick test and `page-break-inside: avoid` on the `` doesn't keep Chrome/Safari from breaking inside the image so clearly we'll want a little more time on that in a future iteration.

PR #799, opened 2019-01-31
Optimize the query used to count unique contributors

acdha:
This form of the query doesn't require building a 250MB sort buffer.

I have casually tested it by confirming that the values didn't change for a few Lincoln views but it should get more testing before merging.


rstorey:
I gave it a test. I don't know if it needs more than that. It's an order of magnitude quicker than the old query. I sorta want to cherry-pick this into 0.4.17.

Issue #475, opened 2018-10-22
Social Share from the Asset Page


colefemeno:
User stories: 

- As a user, I want to easily share the link to the page I'm working on in a social network or a short link. Also I may want to share the link to the entire item (diary, letter, etc) from the Item page. 
- As a user I am so excited about this one letter and I want to share directly to either Facebook, Twitter, or email(?), or copy a short link so that I can show my family and friends how awesome it is. 

Problems we are trying to solve: 
- users can share an asset 
- capture basic metrics of sharing behavior

Updated acceptance criteria (based on research and conversations below): 
- [ ] Add share links on the asset level for Facebook, Twitter, email, and copy link to clipboard 
- [ ] When a user clicks Facebook, Twitter and email those links open another browser window or mail app signaling that that user is moving out of crowd.loc.gov
- [ ] When a share link has opened a new page, to build trust with our users,  there is a timed message saying that the user is now leaving crowd.loc.gov, before opening the external social platform 
- [ ] A message dialogue in the external platforms will include a short description from the title, a link to crowd.loc.gov, and campaign code
- [ ] When a user clicks copy link there is a message that pops up saying `this link has been copied`
- [ ] Add campaign codes to each link (need to follow up on what campaign codes go where)

rstorey:
As part of this we should consider what will show up as the preview image on FB, Twitter when someone posts a link to the site.

elainekamlley:
An acceptance criteria to consider - build a function that could be adopted library wide.

acdha:
#669 added the `og:image` meta header which many things (i.e. Facebook, Twitter) use for the thumbnail when sharing. We could add those to the item & asset pages but there's a big caveat: those images are potentially *MUCH* larger. I'd be tempted to examine the API response metadata which we're storing for each record to see if there's a more appropriate loc.gov URL which we could reference instead.

elainekamlley:
@jbresner if you have time, do you mind adding the update to this?

jbresner:
Michelle Springer appears to still be investigating whether AddThis is a viable option for a share tool. A reply to her question on the GSA's Social Media Community of Practice listserv suggested this as an alternative: https://github.com/usinterior/ncshare.

acdha:
In addition to the privacy risks, one of the arguments for removing AddThis from wdl.org was the substantial page weight — if memory serves, that pulled in more JavaScript + CSS than the rest of the site. We should measure the performance cost as part of the comparison.

[...]

PR #801, opened 2019-02-04
Social share from the asset page

rstorey:
Refs #475


concordia/static/js/contribute.js
```diff
@@ -415,3 +415,23 @@ $tagEditor
 
         displayMessage('error', message, 'tags-save-result');
     });
+
+var $copyUrlButton = $('#copy-url-button');
+$copyUrlButton.tooltip();
```
acdha: Currently the tooltip displays on click no matter what actually happens. I'm wondering whether that should be changed to manual activation and then the JavaScript API would be used to display it after the clipboard operation is successful:

```suggestion
$copyUrlButton.tooltip({trigger: "manual"});
```

Part of why I'm thinking about this is that while most browsers support the clipboard API past abuses have lead to a number of restrictions:

https://caniuse.com/#feat=clipboard

That makes me wonder whether the logic basically needs to be something like putting the current body of the click even handler into a try block which would display either a success message (with auto-hiding) or an error message (which doesn't go away until dismissed) after the `execCommand` call.

concordia/static/css/base.css
```diff
@@ -850,6 +850,30 @@ body.view-transcriptions--asset-detail main,
...
+.twitter-icon {
+    display: inline-block;
+    vertical-align: text-top;
+    height: 19px;
+    width: 19px;
+    background-image: url(../img/twitter-icon.png);
+    background-repeat: no-repeat;
+}
+
+.white {
```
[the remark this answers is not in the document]
rstorey: Done

concordia/static/js/contribute.js
rstorey: This is done, except for making the error message stay until dismissal... working on that part.

concordia/static/js/contribute.js
```diff
@@ -415,3 +415,37 @@ $tagEditor
...
+    var currentAssetUrl = document.getElementById('currentAssetUrl');
+    currentAssetUrl.select();
+    var tooltipMessage = '';
+    try {
+        document.execCommand('copy');
+        // Show the tooltip with a success message
+        tooltipMessage = 'This link has been copied to your clipboard';
+    } catch (e) {
+        // Display an error message in the tooltip
+        tooltipMessage = 'Could not access your clipboard';
```
acdha: We should add `Raven.captureException(e)` here so we can collect metrics on how frequently this happens

concordia/static/js/contribute.js
```diff
@@ -415,3 +415,44 @@ $tagEditor
...
+    $currentAssetUrl.removeClass('d-none');
+    var currentAssetUrl = document.getElementById('currentAssetUrl');
+    currentAssetUrl.select();
+    var tooltipMessage = '';
+    try {
+        document.execCommand('copy');
+        // Show the tooltip with a success message
+        tooltipMessage = 'This link has been copied to your clipboard';
+        $currentAssetUrl.addClass('d-none');
+        $(this)
```
acdha: Since we already have this aliased as `$copyUrlButton` we can avoid recreating the object:

```suggestion
        $copyUrlButton
```

concordia/static/js/contribute.js
```diff
@@ -415,3 +415,44 @@ $tagEditor
...
+            .tooltip('dispose')
+            .tooltip({title: tooltipMessage})
+            .tooltip('show');
+        $(this).on('shown.bs.tooltip', hideTooltipCallback);
```
acdha:
```suggestion
        $copyUrlButton
```

Or just continue the chain for this line, too.

concordia/static/js/contribute.js
```diff
@@ -415,3 +415,44 @@ $tagEditor
...
+    } catch (e) {
+        // Display an error message in the tooltip
+        tooltipMessage =
+            '<p>Could not access your clipboard.</p><button class="btn btn-light btn-sm" id="dismiss-tooltip-button">Close</button>';
+        $currentAssetUrl.addClass('d-none');
+        $(this)
```
acdha:
```suggestion
        $copyUrlButton
```

concordia/static/js/contribute.js
```diff
@@ -415,3 +415,44 @@ $tagEditor
...
+            .tooltip('dispose')
+            .tooltip({title: tooltipMessage, html: true})
+            .tooltip('show');
+        $('#dismiss-tooltip-button').on('click', function() {
+            $('#copy-url-button').tooltip('hide');
```
acdha:
```suggestion
            $copyUrlButton
```

PR #812, opened 2019-02-15
Activity-oriented UI experiment

acdha:
# Goals

* Help users stay in the zone for a particular activity
* Make it easy to pick items you want to work on even if that crosses projects
* Never leave fullscreen mode unless you navigate to an external source

## Initial Features

* [x] Display list of assets
* [x] Make asset list sortable
* [x] Make asset list filterable by campaign, project, or free-text
* [x] Create a good metadata viewer component
* [x] Show adjacent pages in the viewer
* [ ] Component-ize the code for the transcribe/review panel which does the right thing for mode changes and data updates (change from plain HTML to full component) (@acdha )
* [x] Decide where the filtering UI wants to be (a sidebar, drop-down panel, etc.) and what filters are available. Currently that's just campaign but we likely will want other things like project, text queries, tags, etc. so we'll want room for multiple controls and non-trivial width.
* [x] WebSocket-based realtime updates for other users modifying items
* [x] Resize splitter between the viewer and the text entry/review panels (@acdha )
* [x] ~~Released asset reservations with no save activity increase difficulty score~~ see wontfix comment in https://github.com/LibraryOfCongress/concordia/pull/861 (@rstorey )
* [x] Initialize ~~and update~~ difficulty scores with celerybeat (updates happen on post-save signals) (@rstorey ) https://github.com/LibraryOfCongress/concordia/pull/861
* [x] Show a message if screen width is too small (@jbresner + dev TBD)
* [ ] Determine behavior for UI feedback after button submits - modal dialog? banner? disable / enable textarea? (@acdha )
* [ ] Review OpenSeadragon controls - left/right controls, fullscreen, help (@jbresner )
* [ ] What should the transcription text box look like when you're reviewing? Particularly when there is nothing to transcribe? (@jbresner + dev TBD)
* [ ] Consistent terminology and descriptions for interface components / help guide / documentation (@elainekamlley or @lalgee ?)

## Known Bugs

* [ ] Better looking badges / UI for when something is reserved by another user (@jbresner )
* [x] Make Save/Approve type controls stay in one spot and not go below the fold (@acdha + @jbresner)
* [ ] Either use the footer for additional stuff or get rid of it (@jbresner + @acdha)
* [x] Save/Approve buttons aren't implemented
* [x] Sorting and filtering options should be preserved when new data is loaded rather than applied only once at the time of selection
* [x] Slow XHR responses can result in data for a previous mode being retrieved after the mode switch. We should confirm that the new status handling code prevents items of the wrong status from being displayed.
* [ ] Encourage observance of the rules by only showing buttons that you're allowed to click (e.g. submitted_by cannot Accept)
* [ ] What should the adjacent pages section of the OpenSeadragon viewer look like? i.e. how do we indicate which page you're transcribing and which ones are just there for context?
* [x] Behavior when there are no assets matching filter / sort options (e.g. Branch Rickey, "easy stuff")
* [ ] Side panel help text - to what extent does this need to be editable / customizable?
* [x] Campaign API request returns 500 error if `thumbnail_image` is not set for any published campaigns or projects
* [x] Nothing to transcribe checkbox behavior
## Deferred Features

* Difficulty score adjustment - slider, admin controls (campaign weights)
* Sizable / draggable / pinnable panels with a "reset to default" button
* Enhancing the viewer to allow viewing existing transcriptions for other assets within the same item
* Enhancing the viewer to allow viewing arbitrary other / all assets within the same item and any other status or transcription data about those assets
* Making the asset list display category separators and what are the dividers? e.g. if you're filtering by campaign, group by project; if showing all campaigns, group by campaign, etc.
* Making the asset list collapse based on some hierarchy concept (e.g. if the list is displaying campaign boundaries, collapse assets into an expandable group labeled by project)
* Making the whole thing work in narrow viewports
* IE11
* Changing screen real estate strategy when in fullscreen mode
* Sort by date
* View completed assets
* Back button functionality
* User interaction tracking w/ adobe analytics


klee72:
I noticed that asset-list content is only available when mouseover. I think the overlay effect can be achieved with CSS, and it'd be better to have the content available for all asset-list on the page.

klee72:
Also asset-list is not tabbable, `tabindex=0` can be applied to the `<li>`. And would probably need to add `keypress` or something to match `click` link to the asset detail page.

acdha:
The asset list tooltips are based on dynamically-loaded content which isn't present in the page itself. We can add triggering based on the focus event for keyboard navigation but it would be a big increase in memory / network transfer to pre-generate the tooltip for every asset so we could use pure CSS to show it.

acdha:
@rstorey This branch now has real-updates but there are a number of FIXME items. Several things which I flagged for further work:

* Deciding what we want to do when assets change status or haven't been dynamically loaded yet: I'm thinking that the main way we tackle this is by changing the asset list refresh method to either load *every* asset regardless of status and we hide the ones where the status is not the active mode _or_ we change the model so we keep the `assets` list containing everything and have the mode change method filter on the status value. One related challenge is that we will want a timestamp — not necessarily last-updated since even a time-served will do — to allow the code to tell when data which is fetched from the JSON API is older than data received from the WebSocket or vice versa.
* The asset reservation message needs to become two messages (obtained and released) and we need to fire the signals manually since we're not using the Django ORM so we can have Postgres guarantee atomicity: https://github.com/LibraryOfCongress/concordia/blob/f24f8bad26f313a388e541c775149288e0ebe67d/concordia/views.py#L1126-L1143
* We currently don't have a way to know whether the submitted_by or similar user-ID fields are the current user
* We need to start implementing the UI showing what happens when an asset changes status — does it just disappear (maybe even if you have it open?) or does it get badged similar to the locking? etc.

rstorey:
> * One related challenge is that we will want a timestamp — not necessarily last-updated since even a time-served will do — to allow the code to tell when data which is fetched from the JSON API is older than data received from the WebSocket or vice versa.

I added some timestamps to the channel and the asset API json response. "Time served" sounds like a jail sentence so I called it "last updated" - other ideas welcome.

rstorey:
> * We currently don't have a way to know whether the submitted_by or similar user-ID fields are the current user

I added this too

concordia/templates/action-app.html
```diff
@@ -14,6 +14,7 @@
     <script type="module">
         import { ActionApp } from "{% static 'js/action-app/index.js' %}";
         window.actionApp = new ActionApp({
+            currentUser: "{{ request.user.pk }}",
```
acdha: This should behave like a normal JavaScript variable and either be undefined or the value so checks don't need to guard against the “None” string

concordia/static/js/action-app/index.js
```diff
@@ -136,8 +136,12 @@ export class ActionApp {
                     break;
                 }
                 case 'asset_reservation_obtained':
-                    // FIXME: we need to test whether the user who reserved it is different than the user we're running as!
-                    this.markAssetAsUnavailable(assetId);
+                    if (
+                        this.config.currentUser != 'None' &&
```
acdha: I'd like this to be `if (this.config.currentUser && this.config.currentUser != message.user_pk)`

concordia/consumers.py
```diff
@@ -10,7 +12,11 @@ class AssetConsumer(AsyncJsonWebsocketConsumer):
         await self.channel_layer.group_discard("asset_updates", self.channel_name)
 
     async def asset_update(self, message):
-        await self.send_json({"message": message})
+        await self.send_json(
+            {"message": message, "message_timestamp": int(time.time())}
```
acdha: `message_timestamp` feels a little long to me. Maybe this should just be something like `sent` which also focuses on the part about this being the transmission time?

concordia/views.py
```diff
@@ -1394,6 +1394,7 @@ def serialize_object(self, obj):
                 "url": campaign.get_absolute_url(),
             },
             "latest_transcription": obj.latest_transcription,
+            "last_updated": int(time.time()),
```
acdha: `last_updated` could be interpreted as the record's modification time. I'm thinking that we could move this out of the per-object serialization and just include it a single time in the response, probably by adding it to the API view base class's `serialize_context`.

entrypoint.sh
```diff
@@ -11,14 +11,11 @@ echo Running makemigrations
 echo Running migrations
 ./manage.py migrate
 
-echo "Ensuring our base configuration is present in the database"
+echo Ensuring our base configuration is present in the database
```
acdha: 👎 on this change since it means the string will be highlighted following shell syntax rules in, for example, GitHub diffs

entrypoint.sh
```diff
@@ -11,14 +11,11 @@ echo Running makemigrations
...
 ./manage.py ensure_initial_site_configuration
 
-echo "Initializing difficulty values"
-./manage.py initialize_difficulty_values
```
acdha: I think this is reasonable but do we have an easy way to trigger this otherwise? I'm wondering whether that should be something like a nightly celery task or an admin button.

concordia/static/js/action-app/index.js
rstorey: Done

concordia/templates/action-app.html
rstorey: Done

concordia/consumers.py
rstorey: > It looks like it's working just fine but the filesystem was modified outside of the app. Do we know how that happened?

It doesn't work inside a docker container. I was able to reproduce the problem locally.

entrypoint.sh
rstorey: I will create a different way to trigger this.

entrypoint.sh
rstorey: I wasn't sure which way was preferred, I made it consistent with the most recent changes. I'll put it back

concordia/consumers.py
rstorey: Done

concordia/static/js/action-app/index.js
```diff
@@ -131,9 +131,15 @@ export class ActionApp {
                     break;
                 }
                 case 'asset_reservation_obtained':
+                    // If the user is anonymous,
```
acdha: This is long enough that it can probably be a `/*` block & rewrapped

acdha:
This has been rebased against #829 since it now uses bulk_update

concordia/tasks.py
```diff
@@ -212,11 +212,39 @@ def populate_asset_years():
...
+    for asset_chunk in chunked(asset_qs, 500):
+        changed_assets = []
+
+        for asset in asset_chunk:
+            metadata = asset.item.metadata
+
+            year = None
+            for date_outer in metadata["item"]["dates"]:
+                for date_inner in date_outer.keys():
+                    if not date_inner.isdigit() or len(date_inner) != 4:
```
rstorey: This will fail because some of the dates are in the format of "1936 to 1954"

acdha: Doesn't it just fail by logging? It ran to completion on my local dev install

rstorey: Yeah, but it won't populate the field for those assets, and I think it should, since it still works for sorting.

acdha:
@rstorey I didn't have much time available this week but the rough plan I have for the next step is refactoring all of the state management logic so we'll have a single source of truth for when an asset cannot be worked on which will be passed to both the asset list and viewer components. That should probably be two stages so we can do a single check for logged-in status (i.e. disable reviewing entirely), whether the user reviewing an asset is the same as the last person to submit it, and a separate toggle for the temporary locks like whether there's an AJAX operation in play so we can avoid even asking for things like reservations when the user won't be able to work on something anyway.

jbresner:
Here are the latest set of comps for the prototype. I still have some specs I want to add but may just add them as additional comments.

Asset View - Default View
Asset View - Help
Asset View - More Info
Asset View - Share
Asset View - Slideshow Closed
Asset View - Slideshow Next Image
Asset View - Tabs Closed
Asset View - Thumbnail Scroll
Flat View - All Items
Flat View - Help
Large Image Grid
Small Image Grid

jbresner:
And of course I wanted @rstorey and @acdha to provide feedback on the feasibility of what I'm proposing in the designs in case their are any technical hurdles with these approaches.

acdha:
We're currently using the same sharing code on both views where we display pages. Would your proposed change apply in both cases? We'd also want to flesh out the behaviour with an eye towards accessibility: I'm assuming the idea is that when clicked/selected it would stay open until either a button is picked or the user interacts with something outside of that control?

acdha:
I'm also wondering about the more information view — I like having it be visible while working but we have a considerable amount of metadata to display on many pages so there'd be a lot of scrolling. I wonder whether that might mean we should move it into the right side panel, have an optional way to further expand it / make it a popout, etc.

jbresner:
> We're currently using the same sharing code on both views where we display pages. Would your proposed change apply in both cases? 
Unless there's a problem with that approach that's what I'd recommend.

> We'd also want to flesh out the behaviour with an eye towards accessibility: I'm assuming the idea is that when clicked/selected it would stay open until either a button is picked or the user interacts with something outside of that control?
Correct. For Facebook and Twitter, selecting the buttons would open a new window with the respective Share dialog boxes. For the link button, once the button is selected, we may want to display a Copied to Clipboard" message (or some sort of wording) to know that the link is available to paste somewhere. Maybe the message could look like a tool tip below the link button? If not, I'll have to come up with a design for the display of the message.

jbresner:
> I'm also wondering about the more information view — I like having it be visible while working but we have a considerable amount of metadata to display on many pages so there'd be a lot of scrolling. I wonder whether that might mean we should move it into the right side panel, have an optional way to further expand it / make it a popout, etc.

I like the idea of adding it to the side panel but I was trying to keep the content related to the item within the main work area (the image on the left and the transcription/review area on the right). I'm worried that an "i" icon in that vertical tab bar wouldn't be understood as changing every time a new item is displayed.

Also somethingto consider is the screen real estate. I designed the mockups with a 1200-px screen width. If a user had a larger screen, they could stretch that transcription/review section wider and would probably have plenty of room to view. BUT, I understand what you're saying and maybe keeping it where it is with an optional "pop-out" button solves all issues. Not sure if we'd want to push to production as is and make this something we get feedback before making additional changes.

jbresner:
And one other side not related to screen sizes. One of the items way up top  was "Show a message if screen width is too small". Instead of a message, I'd suggest just picking a minimum screen size we accommodate (e.g. 960px) and once a user is either on a device with a smaller screen or resizes their browser window, we just have a horizontal scroll. If we are treating this product as an app, I don't see any problem with this approach and no need for an intrusive message (if they really want to review on an iPhone, they can just scroll left-right).

acdha:
I think we can eventually handle display size that way but it will require testing around the independently-scrolling regions — right now that behaviour would require care since you could scroll it normally until you hit the image viewer (or, for the other direction, the asset list) and that'd soak up further scroll events (mouse wheel scrolling has the same problem — scroll, scroll, zoom). I've typically handled that by disable scroll handling until the user interacts with the OpenSeadragon component but we'll definitely want to test it.

acdha:
>  For the link button, once the button is selected, we may want to display a Copied to Clipboard" message (or some sort of wording) to know that the link is available to paste somewhere. Maybe the message could look like a tool tip below the link button?

This is what we currently do and it'd be easiest to keep that approach.

jbresner:
Ah, didn't even notice. Yeah, I might just tweak the styling but exact same approach.

Issue #746, opened 2019-01-04
Further contact us page refinement


lalgee:
After several months of use, I would like to request a few small pages to the Contact us email form that would improve its public and CM experience using it for correspondence:

- Remove "Crowd.loc.gov" from email subject line and replace with "By the People"
- Improve instructional text in fields - currently duplicates the question, which appears right above the field and doesn't always make sense.  Example: "Have a specific page you need help with? Add the link below:"
- Remove the colon that appears after the sender's email address in the email to CMs - this has to be deleted out when copying for a reply
![image](https://user-images.githubusercontent.com/20258744/50700530-c14fc200-1018-11e9-8188-37326dc78370.png)
- May want to create question-type drop down.  Examples: How to, something isn't working, feature request, collections content, etc.

acdha:
The hostname prefix was set to disambiguate test messages:

https://github.com/LibraryOfCongress/concordia/blob/master/concordia/views.py#L928

We could either decide that we'll live with the possibility of confusion or we can change it use the site name if we start configuring those (currently the name & domain are the same value – see https://crowd.loc.gov/admin/sites/site/).

acdha:
> Remove the colon that appears after the sender's email address in the email to CMs - this has to be deleted out when copying for a reply

Why does that have to be deleted?

lalgee:
Outlook will not send an email with a colon directly after it.  It bounces it back as undeliverable. And if you wait to delete it or move into another field before doing so outlook autorecognizes it as an email address and then you have to delete the whole thing and repaste and then delete the colon.
So a small thing, but would make life easier just to get rid of the colon in the message or have the sender's email listed somewhere additionally where it has nothing around it and is easily copy-able.

acdha:
> Improve instructional text in fields - currently duplicates the question, which appears right above the field and doesn't always make sense. Example: "Have a specific page you need help with? Add the link below:"

What's happening here is that the django-bootstrap4 library uses the label as the placeholder (which displays in the field when it has no value) by default. That's applied site-wide so e.g. https://crowd.loc.gov/account/register/ will show “Username” or “Email address”  as well but those look more normal because they're not sentences and many users won't even see them if their browser autofills the values.

We could easily disable auto-filling placeholders from labels site-wide but I'm wondering whether the better fix is to be more consistent in the use of labels vs. help text.

Stripping the placeholders entirely would look like this unless we went to the trouble of selectively re-adding them:

[...]

PR #823, opened 2019-02-25
Removing colon from email field

elainekamlley:
Remove colon from email field in the contact us form. The colon makes the email invalid and harder for CM's to reply back to volunteers.


acdha:
See https://github.com/LibraryOfCongress/concordia/issues/746

elainekamlley:
@rstorey mind approving this?

PR #826, opened 2019-02-26
Set reply to on contact emails

acdha:
This continues #746 and closes #823 but I want to test it separately to confirm that this works as expected in our environment.


elainekamlley:
@acdha This makes more sense. Should I close my PR?

acdha:
@elainekamlley I started on that branch so I'd vote that we merge #823 as soon as @rstorey looks it over but test this before we merge it. I wanted to confirm that we didn't have issues with spam filtering, etc. using `Reply-To` in our environment.

rstorey:
Do you want to merge to master and test in crowd-dev or do a feature branch deployment?

elainekamlley:
@rstorey either one works for me. @acdha have a preference?

acdha:
I was thinking crowd-dev since it _should_ work since it's just using a standard protocol feature but…

PR #830, opened 2019-02-28
Load FontAwesome from CDNJS

acdha:
Now that https://github.com/cdnjs/cdnjs/issues/12307 is done we can finally complete #465 by consolidating all of our usage on a single service which uses the same CDN as the main site and also upgrades from 5.7.1 to 5.7.2.

### DIFF
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -893,6 +893,7 @@
         },
         "isort": {
             "hashes": [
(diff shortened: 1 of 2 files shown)


acdha:
Agreed — I was also wondering whether we should be migrating to only loading the symbols we use or only loading SVG but that's more of a longer-term discussion about our front-end tools than something I want to tackle now.

rstorey:
Merging so I can get the isort fix into master

acdha:
Thanks - I was going to cherry-pick it but both changes are small.

Issue #836, opened 2019-03-04
Allow periods in item IDs


rstorey:
**What behavior did you observe? Please describe the bug**
500 error occurs.

**How can we reproduce the bug?**
1. Add an item that has a . in the ID.
1. Attempt to load the item in the admin pages or view it in the site as a regular user.

**What is the expected behavior?**
Should be able to function as a regular item

**Additional context**
https://crowd-sentry.loc.gov/loc/crowd-backend/issues/242

PR #837, opened 2019-03-04
Allow item IDs to have periods in them

rstorey:
refs #836

### DIFF
--- a/concordia/converters.py
+++ b/concordia/converters.py
@@ -5,3 +5,8 @@ class UnicodeSlugConverter(SlugConverter):
     # This is similar to the slug_unicode_re pattern but is not anchored to the
     # start of the string:
     regex = r"[-\w+]+"
+
+
+class ItemSlugConverter(SlugConverter):
+    # Allows . in the item ID
+    regex = r"[-a-zA-Z0-9_\.]+"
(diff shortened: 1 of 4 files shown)


concordia/urls.py
```diff
@@ -12,6 +12,7 @@
 from . import converters, views
 
 register_converter(converters.UnicodeSlugConverter, "uslug")
+register_converter(converters.ItemSlugConverter, "islug")
```
acdha: I'd be included to give this a longer name since this is really saying that the value is something other than a slug. Maybe just go with `item_id`?

concordia/converters.py
```diff
@@ -5,3 +5,8 @@ class UnicodeSlugConverter(SlugConverter):
     # This is similar to the slug_unicode_re pattern but is not anchored to the
     # start of the string:
     regex = r"[-\w+]+"
+
+
+class ItemSlugConverter(SlugConverter):
```
acdha: Since this isn't really a slug any more, maybe it should just directly subclass [`StringConverter`](https://docs.djangoproject.com/en/2.1/_modules/django/urls/converters/) and be something like `ItemIdConverter`?

rstorey:
According to Patrick, any character that can be in a URL can be in an item ID (including slashes) - except for `?`.

concordia/converters.py
rstorey: Done

concordia/urls.py
rstorey: Done

PR #838, opened 2019-03-04
Put social share behind a feature flag

rstorey:
Uses https://github.com/cfpb/django-flags

Will define flag conditions in DB after deployment.

Refs #475

### DIFF
--- a/concordia/templates/transcriptions/asset_detail.html
+++ b/concordia/templates/transcriptions/asset_detail.html
@@ -1,7 +1,6 @@
 {% extends "base.html" %}
 
 {% load staticfiles %}
-
 {% load concordia_media_tags %}
 
 {% block title %}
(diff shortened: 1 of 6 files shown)


concordia/templates/transcriptions/asset_detail.html
```diff
@@ -67,13 +66,14 @@
...
             </div>
 
-
-            <div class="btn-group btn-group-sm" role="navigation" aria-label="Links to share this page">
-                <a class="btn btn-default" target="_blank" role="button" title="Share this page on Facebook" href="https://www.facebook.com/share.php?u={{ share_url|urlencode:"" }}"><span class="facebook-icon"></span></a>
-                <a class="btn btn-default" target="_blank" role="button" title="Share this page on Twitter" href="https://twitter.com/intent/tweet?text={{ tweet_text|urlencode:"" }}&source=webclient"><span class="twitter-icon"></span></a>
-                <a class="btn btn-default" id="copy-url-button" role="button" data-toggle="tooltip"  data-trigger="manual"  data-placement="bottom" aria-label="Copy this link to your clipboard"><i class="fas fa-link text-white"></i></a>
-                <input type="text" tabindex="-1" aria-hidden="true" class="d-none" id="currentAssetUrl" name="currentAssetUrl" value="{{ current_asset_url }}" />
-            </div>
+            {% if social_share_flag %}
```
acdha: Do you think it would be cleaner to use flags' [template tag library](https://cfpb.github.io/django-flags/api/django/)?

rstorey: I tried that, but it didn't work and I'm not sure why. I was going to dig into that a little bit to see if it has something to do with the class-based view.

rstorey: I opened an issue for this - I'm not sure why the templatetag doesn't work. 
https://github.com/cfpb/django-flags/issues/28

Issue #822, opened 2019-02-22
Banner message to users who are reviewing at the same time


elainekamlley:
When two or more users are reviewing the same asset a message banner is shown, similar to transcribing.

PR #848, opened 2019-03-08
Review banner

rstorey:
Refs #822 

- [x] Refactor find next reviewable asset to exclude those with reservations

### DIFF
--- a/concordia/static/js/asset-reservation.js
+++ b/concordia/static/js/asset-reservation.js
@@ -1,7 +1,50 @@
 /* global jQuery displayMessage displayHtmlMessage buildErrorMessage */
-/* exported attemptToReserveAsset checkForReviewers */
+/* exported attemptToReserveAsset */
 
-function addReservationUnloadListener(reservationURL) {
+function attemptToReserveAsset(reservationURL, findANewPageURL, actionType) {
+    var $transcriptionEditor = jQuery('#transcription-editor');
+
+    jQuery
+        .ajax({
+            url: reservationURL,
+            type: 'POST',
+            dataType: 'json'
+        })
+        .done(function() {
+            $transcriptionEditor
+                .data('hasReservation', true)
+                .trigger('update-ui-state');
+        })
+        .fail(function(jqXHR, textStatus, errorThrown) {
+            if (jqXHR.status == 409) {
+                if (actionType == 'transcribe') {
+                    $transcriptionEditor
+                        .data('hasReservation', false)
+                        .trigger('update-ui-state');
+                    jQuery('#asset-reservation-failure-modal').modal();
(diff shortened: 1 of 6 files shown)


concordia/static/js/asset-reservation.js
```diff
@@ -1,5 +1,28 @@
-/* global jQuery displayMessage buildErrorMessage */
-/* exported attemptToReserveAsset */
+/* global jQuery displayMessage displayHtmlMessage buildErrorMessage */
+/* exported attemptToReserveAsset checkForReviewers */
+
+function addUnloadListener(reservationURL) {
```
acdha: This name is a bit generic for something this focused

concordia/static/js/asset-reservation.js
```diff
@@ -1,5 +1,28 @@
...
+        // We'll try Beacon since that's reliable but until we can drop support for IE11 we need a fallback:
+        if ('sendBeacon' in navigator) {
+            navigator.sendBeacon(
+                reservationURL,
+                new Blob([jQuery.param(payload)], {
+                    type: 'application/x-www-form-urlencoded'
+                })
+            );
+        } else {
+            jQuery.ajax({url: reservationURL, type: 'POST', data: payload});
```
acdha: We might want to weight the importance of the user getting a response faster vs. getting the lock released by having the AJAX branch trigger the cancellation prompt (see https://developer.mozilla.org/en-US/docs/Web/API/WindowEventHandlers/onbeforeunload#Example).

concordia/static/js/asset-reservation.js
```diff
@@ -34,24 +57,45 @@ function attemptToReserveAsset(reservationURL) {
...
-                reservationURL,
-                new Blob([jQuery.param(payload)], {
-                    type: 'application/x-www-form-urlencoded'
-                })
-            );
-        } else {
-            jQuery.ajax({url: reservationURL, type: 'POST', data: payload});
-        }
-    });
+function checkForReviewers(reservationURL, findANewPageURL) {
```
acdha: This is basically the same as the preceding function except for the message text. It seems like we should consolidate both of those and consider changing the assumption for transcription reservations to also receive a find-a-new-page URL since that's useful. This could either be generic wording (“Someone else is working on this page”) or refactoring this function to be entirely generic except for the branch which handles the HTTP 409 Conflict by using a callback to display the message rather than having it hard-coded here.

rstorey: I thought the transcription reservation popped up a modal. We don't want the review reservation to do that.

acdha: It used to just be a banner message

Issue #853, opened 2019-03-14
Asset reservation modal pops up multiple times


rstorey:
**How can we reproduce the bug?**

1. As a logged-in user, go to any page where another user is already transcribing.
2. Close the modal that pops up.
3. Stay on the page for a while.
4. Repeat the same process for any page where another user is already reviewing.

**What behavior did you observe? Please describe the bug**
The asset reservation modal pops up and/or the review banner pops up after a minute or so.

**What is the expected behavior?**
1. If the page is in transcribe mode, the review banner should never pop up.
2. If the user dismisses the modal pop up, it shouldn't come up again.

rstorey:
Fixed in v0.4.23

PR #855, opened 2019-03-14
Bug fix for asset reservation modal that pops up multiple times

rstorey:
Refs #853

Removes the timed check for asset reservation update since we don't properly handle most of the possible result cases - the only way it works correctly is if the user who has the asset reservation navigates away from the page without entering any transcription text or changing the asset status. The user really needs to reload the page to try again at getting an asset reservation, until/unless we are going to update all the UI elements that related to latest transcription text and asset status - not just enable the text area. If the current asset detail page continues its life much longer, perhaps we should consider doing just that.

### DIFF
--- a/concordia/static/js/asset-reservation.js
+++ b/concordia/static/js/asset-reservation.js
@@ -1,7 +1,12 @@
 /* global jQuery displayMessage displayHtmlMessage buildErrorMessage */
 /* exported attemptToReserveAsset */
 
-function attemptToReserveAsset(reservationURL, findANewPageURL, actionType) {
+function attemptToReserveAsset(
+    reservationURL,
+    findANewPageURL,
+    actionType,
+    firstTime
+) {
     var $transcriptionEditor = jQuery('#transcription-editor');
 
     jQuery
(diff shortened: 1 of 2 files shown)


rstorey:
> I would like to carry this idea further: we could simply cancel the timer when the user closes the warning modal so it won't try again until the page is reloaded or we could change this slightly by tracking whether the lock attempt had previously failed and then calling `location.reload(true)` if it later gets the lock.

There we go. That's the right way to do it. I'll change it.

concordia/static/js/asset-reservation.js
```diff
@@ -1,7 +1,12 @@
 /* global jQuery displayMessage displayHtmlMessage buildErrorMessage */
 /* exported attemptToReserveAsset */
 
-function attemptToReserveAsset(reservationURL, findANewPageURL, actionType) {
+function attemptToReserveAsset(
+    reservationURL,
+    findANewPageURL,
+    actionType,
+    firstTime
```
acdha: I'm wondering about whether the `firstTime` variable name should be something more like `displayFailureMessage` which describes the effects.

PR #856, opened 2019-03-18
Channels for asset updates and reservations

rstorey:
This adds a websocket connection in action-app.html that gets messages on asset updates and asset reservations. Still a lot to do, but this should get us started.

### DIFF
--- a/concordia/routing.py
+++ b/concordia/routing.py
@@ -1,0 +1,7 @@
+from channels.routing import ProtocolTypeRouter
+
+application = ProtocolTypeRouter(
+    {
+        # Empty for now (http->django views is added by default)
+    }
+)
--- a/concordia/routing.py
+++ b/concordia/routing.py
@@ -1,7 +1,16 @@
-from channels.routing import ProtocolTypeRouter
+# mysite/routing.py
+from channels.auth import AuthMiddlewareStack
+from channels.routing import ProtocolTypeRouter, URLRouter
+from django.conf.urls import url
+from . import consumers
+
+websocket_urlpatterns = [
+    url(r"^ws/asset/(?P<asset_pk>[^/]+)/$", consumers.AssetConsumer)
+]
 
 application = ProtocolTypeRouter(
     {
-        # Empty for now (http->django views is added by default)
+        # (http->django views is added by default)
+        "websocket": AuthMiddlewareStack(URLRouter(websocket_urlpatterns))
     }
 )
--- a/concordia/routing.py
+++ b/concordia/routing.py
@@ -1,7 +1,7 @@
-# mysite/routing.py
 from channels.auth import AuthMiddlewareStack
 from channels.routing import ProtocolTypeRouter, URLRouter
 from django.conf.urls import url
+
 from . import consumers
 
 websocket_urlpatterns = [
--- a/concordia/routing.py
+++ b/concordia/routing.py
@@ -5,7 +5,7 @@ from django.conf.urls import url
 from . import consumers
 
 websocket_urlpatterns = [
-    url(r"^ws/asset/(?P<asset_pk>[^/]+)/$", consumers.AssetConsumer)
+    url(r"^ws/asset/(?P<asset_id>[^/]+)/$", consumers.AssetConsumer)
 ]
 
 application = ProtocolTypeRouter(
--- a/concordia/routing.py
+++ b/concordia/routing.py
@@ -4,13 +4,11 @@ from django.conf.urls import url
 
 from . import consumers
 
-websocket_urlpatterns = [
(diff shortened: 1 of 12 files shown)


acdha:
Let me rebase the activity-ui branch against master first so we don't have that chaff in the review here

concordia/signals/handlers.py
```diff
@@ -33,3 +40,25 @@ def update_asset_status(sender, *, instance, **kwargs):
...
+
+
+@receiver(post_save, sender=Asset)
+def send_asset_update(sender, *, instance, **kwargs):
+    channel_layer = get_channel_layer()
+    AsyncToSync(channel_layer.group_send)(
+        "asset_updates",
+        {
+            "type": "asset_update",
+            "asset_slug": instance.slug,
```
acdha: Slugs aren't necessarily unique across items — I'm thinking this should either be the PK or the URL so the client can match it easily. The URL would be more proper but the PK doesn't require looking up the campaign, project, and item values and that's already in our API response for that very reason.

```suggestion
            "id": instance.pk,
```

concordia/signals/handlers.py
```diff
@@ -33,3 +40,25 @@ def update_asset_status(sender, *, instance, **kwargs):
...
+            "asset_status": instance.transcription_status,
```
acdha: Are there any other values we want to be able to refresh instantly? The only one which came to mind is the difficulty since the others only change in response to admin actions but maybe there's also a thought to including the updated timestamp in both places to support a most/least-recently-updated sort?

concordia/signals/handlers.py
```diff
@@ -33,3 +40,25 @@ def update_asset_status(sender, *, instance, **kwargs):
...
+        },
+    )
+
+
+@receiver(post_save, sender=AssetTranscriptionReservation)
+def send_asset_reservation(sender, *, instance, **kwargs):
+    channel_layer = get_channel_layer()
+    AsyncToSync(channel_layer.group_send)(
+        "asset_updates",
+        {"type": "asset_reservation", "asset_slug": instance.asset.slug},
```
acdha: Same question re: PK or URL, leaning the former for performance

concordia/routing.py
```diff
@@ -0,0 +1,14 @@
...
+from channels.routing import ProtocolTypeRouter, URLRouter
+from django.conf.urls import url
+
+from . import consumers
+
+application = ProtocolTypeRouter(
+    {
+        # (http->django views is added by default)
+        "websocket": AuthMiddlewareStack(
+            URLRouter([url("^ws/asset/asset_updates/$", consumers.AssetConsumer)])
```
acdha: Do we need one for reservations as well or do these multiplex?

acdha: (I'm leaning “should” but wanted to think about that more than I have time for today)

acdha:
Have you tried any of the available ASGI servers yet? I've used [Daphne](https://github.com/django/daphne) in light testing and I was looking at [uvicorn w/gunicorn](https://www.uvicorn.org/#running-with-gunicorn) just now.

concordia/signals/handlers.py
rstorey: Looks like we don't actually have an `updated_on` field for `Asset`. Should we add one?

concordia/routing.py
rstorey: I think I should rename the endpoint. The same consumer is used for both updates and reservations currently. We could change that, but I don't see what the benefit of doing so would be. I think it makes more sense to have one stream of messages and then use the message types to distinguish the types of events - updates, reservations, reservation releases (still need to do the reservation releases bit).

rstorey:
> Have you tried any of the available ASGI servers yet? I've used [Daphne](https://github.com/django/daphne) in light testing and I was looking at [uvicorn w/gunicorn](https://www.uvicorn.org/#running-with-gunicorn) just now.

I've only used Daphne in testing. I'm thinking we should run the websockets and HTTP together and I'm sorta leaning towards using Daphne to do it. Do you have any specific opinions about deployment?

concordia/signals/handlers.py
acdha: I'm leaning no unless @elainekamlley or @jb thinks we should have that particular sort option.

concordia/routing.py
acdha: That seems reasonable to me since we currently don't have any concept for reserving something other than an asset

acdha:
I'm voting we switch to Daphne and make a point of re-doing some of our performance comparisons to make sure the load numbers are reasonable.

acdha:
Most of this has been merged into #812, which starts to expose the websocket interface into the actual app. I'm thinking we close this branch and continue working on activity-ui-prototype.

Issue #815, opened 2019-02-22
Prototype: implement asset reservation system 


elainekamlley:
asset reservation on viewer open and release on viewer close

PR #860, opened 2019-03-21
Reservation messages

rstorey:
Sends messages to websocket from Django for reservations obtained and released. Still to do: actually reserving and releasing within the prototype (ticketed in #815)

### DIFF
--- a/concordia/views.py
+++ b/concordia/views.py
@@ -1135,10 +1135,31 @@ def reserve_asset(request, *, asset_pk):
     )
 
     with connection.cursor() as cursor:
-        cursor.execute(
-            "DELETE FROM concordia_assettranscriptionreservation WHERE updated_on < %s",
+        rows_to_release = cursor.execute(
+            (
+                "SELECT asset_id as asset_pk, user_id as user_pk"
+                "FROM concordia_assettranscriptionreservation "
+                "WHERE updated_on < %s"
+            ),
             [cutoff],
         )
+        if rows_to_release:
+            reservations = []
+            for row in rows_to_release.fetchone():
+                reservations += (row["asset_pk"], row["user_pk"])
+            cursor.execute(
+                (
+                    "DELETE FROM concordia_assettranscriptionreservation "
+                    "WHERE updated_on < %s"
+                ),
+                [cutoff],
+            )
+            for reservation in reservations:
+                reservation_released.send(
+                    sender="reserve_asset",
+                    asset_pk=reservation[0],
+                    user_pk=reservation[1],
+                )
 
     # If the browser is letting us know of a specific reservation release,
     # let it go even if it's within the grace period.
--- a/concordia/views.py
+++ b/concordia/views.py
@@ -67,6 +67,7 @@ from concordia.models import (
     TranscriptionStatus,
     UserAssetTagCollection,
 )
+from concordia.signals.signals import reservation_obtained, reservation_released
 from concordia.templatetags.concordia_media_tags import asset_media_url
 from concordia.utils import get_anonymous_user, request_accepts_json
 from concordia.version import get_concordia_version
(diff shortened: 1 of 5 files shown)


concordia/views.py
```diff
@@ -1128,17 +1129,40 @@ def reserve_asset(request, *, asset_pk):
...
-    # First clear old reservations, with a grace period:
+    # First clear old reservations, with a grace period.
     cutoff = timestamp - (
         timedelta(seconds=2 * settings.TRANSCRIPTION_RESERVATION_SECONDS)
     )
 
     with connection.cursor() as cursor:
-        cursor.execute(
-            "DELETE FROM concordia_assettranscriptionreservation WHERE updated_on < %s",
+        rows_to_release = cursor.execute(
```
acdha: I think we can do this in one query using [`DELETE … RETURNING`](https://www.postgresql.org/docs/9.6/sql-delete.html)

rstorey: Oh that would be cool. I'll try that tomorrow

concordia/views.py
```diff
@@ -1128,17 +1129,34 @@ def reserve_asset(request, *, asset_pk):
...
+    reservations = []
+
     with connection.cursor() as cursor:
-        cursor.execute(
-            "DELETE FROM concordia_assettranscriptionreservation WHERE updated_on < %s",
+        rows_to_release = cursor.execute(
+            (
+                "DELETE FROM concordia_assettranscriptionreservation "
+                "WHERE updated_on < %s "
+                "RETURNING asset_id as asset_pk, user_id as user_pk"
```
acdha: This is probably pure pedantry but I like to capitalize SQL keywords:

```suggestion
                "RETURNING asset_id AS asset_pk, user_id AS user_pk"
```

concordia/views.py
```diff
@@ -1128,17 +1129,34 @@ def reserve_asset(request, *, asset_pk):
...
+            ),
             [cutoff],
         )
 
+    if rows_to_release:
+        for row in rows_to_release.fetchone():
```
acdha: Shouldn't this be `fetchall`?

```suggestion
        for row in rows_to_release.fetchall():
```

concordia/views.py
```diff
@@ -1128,17 +1129,34 @@ def reserve_asset(request, *, asset_pk):
...
+            reservations += (row["asset_pk"], row["user_pk"])
+
+    for reservation in reservations:
```
acdha:
```suggestion
    for asset_pk, user_pk in reservations:
```

concordia/views.py
```diff
@@ -1128,17 +1129,34 @@ def reserve_asset(request, *, asset_pk):
...
+        reservation_released.send(
+            sender="reserve_asset", asset_pk=reservation[0], user_pk=reservation[1]
```
acdha:
```suggestion
            sender="reserve_asset", asset_pk=asset_pk, user_pk=user_pk
```

concordia/views.py
acdha: I usually like to unpack lists like this rather than having array indices in more complex code just in case that ever changes in the future

concordia/views.py
```diff
@@ -1128,17 +1129,34 @@ def reserve_asset(request, *, asset_pk):
...
+        rows_to_release = cursor.execute(
+            (
+                "DELETE FROM concordia_assettranscriptionreservation "
+                "WHERE updated_on < %s "
+                "RETURNING asset_id AS asset_pk, user_id AS user_pk"
+            ),
             [cutoff],
         )
 
+    if rows_to_release:
```
acdha: I would have thought that this block would need to be inside the cursor context manager to avoid the cursor being closed before the results are retrieved

rstorey: It passed the unit tests... we should definitely test it more.

acdha: I tweaked it in the main prototype branch

PR #861, opened 2019-03-21
Difficulty values

rstorey:
- [x] Move `initialize_difficulty_values` from management command to celery task that can be invoked through the Django admin
- [x] Update difficulty values based on activity
- [ ] ~~released asset reservations without activity increase difficulty score~~ wontfix, initial testing shows that reservations get released and re-obtained too often for this to be a reliable indicator


rstorey:
We could run the initialize method nightly instead of doing anything fancy (or anything at all) with updating...

concordia/signals/handlers.py
```diff
@@ -89,6 +89,24 @@ def send_asset_reservation_released(sender, **kwargs):
...
     )
 
+    no_activity = True
+    # TODO: determine whether the user saved any transcriptions
+    # before releasing the reservation.
+    # If not, then no_activity should be true
+
+    if no_activity:
+        asset = Asset.objects.filter(pk=kwargs["asset_pk"]).annotate(
+            contributor_count=Count("transcription__user", distinct=True),
```
rstorey: @acdha Please tell me there's a better alternative than what I'm doing here. To me, this is terrible.

acdha:
>  initial testing shows that reservations get released and re-obtained too often for this to be a reliable indicator

Agreed — let's focus on having an explicit user signal. The main concern I have there is that it seems like we would need some way to limit it to one user, one vote, which means another database table.

acdha:
>  initial testing shows that reservations get released and re-obtained too often for this to be a reliable indicator

Agreed — let's focus on having an explicit user signal. The main concern I have there is that it seems like we would need some way to limit it to one user, one vote, which means another database table.

PR #862, opened 2019-03-25
Set CONN_MAX_AGE for celery 4.2.1 when using Docker Postgres DB

rstorey:
See https://github.com/celery/celery/issues/4878

Doesn't seem to be a problem in AWS - I couldn't reproduce it in crowd-dev. Only when using a docker container as the Postgres DB.


acdha:
I wonder whether this is something like TCP keepalives keeping the session open on the real servers. Anyway, I'm +1 on this — there could be an argument for a Celery-specific settings file but then we'd have to make sure people use it and I'm not sure there's enough benefit to make that worthwhile.

PR #864, opened 2019-03-25
Exponential backoff for retrieving pages of collection data from JSON API

rstorey:
In my local testing, this was enough to be able to get the celery tasks running for the Whitman materials: https://www.loc.gov/manuscripts/?q=feinberg+poetry+whitman+Poetry,+1842-1892

I'd prefer to make anything that makes a JSON API request into its own celery task, but I think that might be too much work for the timeline. (This campaign needs to be imported by April 1.)

### DIFF
--- a/importer/tasks.py
+++ b/importer/tasks.py
@@ -4,7 +4,6 @@ See the module-level docstring for implementation details
 
 import os
 import re
-import time
 from functools import wraps
 from logging import getLogger
 from tempfile import NamedTemporaryFile
@@ -15,7 +14,9 @@ from celery import group, task
 from django.db.transaction import atomic
 from django.utils.text import slugify
 from django.utils.timezone import now
+from requests.adapters import HTTPAdapter
 from requests.exceptions import HTTPError
+from requests.packages.urllib3.util.retry import Retry
 
 from concordia.models import Asset, Item, MediaType
 from concordia.storage import ASSET_STORAGE
@@ -40,6 +41,26 @@ ACCEPTED_P1_URL_PREFIXES = [
     "photos",
     "websites",
 ]
+
+
+def requests_retry_session(
+    retries=10,
+    backoff_factor=0.3,
+    status_forcelist=(429, 500, 502, 503, 504),
+    session=None,
+):
+    session = session or requests.Session()
+    retry = Retry(
+        total=retries,
+        read=retries,
+        connect=retries,
+        backoff_factor=backoff_factor,
+        status_forcelist=status_forcelist,
+    )
+    adapter = HTTPAdapter(max_retries=retry)
+    session.mount("http://", adapter)
+    session.mount("https://", adapter)
+    return session
 
 
 def update_task_status(f):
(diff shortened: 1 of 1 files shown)


acdha:
Also, it looks like we could pro-actively set [`rate_limit`](http://docs.celeryproject.org/en/latest/userguide/tasks.html#Task.rate_limit) for things we know are aggressively limited

rstorey:
> Can't we use the same Celery task automatic retry mechanism we're using for asset downloads?
> 
> [concordia/importer/tasks.py](https://github.com/LibraryOfCongress/concordia/blob/32441a3714d229b93eba889c3350432c491a21f5/importer/tasks.py#L417-L425)
> 
> Lines 417 to 425 in [32441a3](/LibraryOfCongress/concordia/commit/32441a3714d229b93eba889c3350432c491a21f5)
> 
>  @task( 
>      bind=True, 
>      autoretry_for=(HTTPError,), 
>      retry_backoff=True, 
>      retry_backoff_max=8 * 60 * 60, 
>      retry_jitter=True, 
>      retry_kwargs={"max_retries": 12}, 
>  ) 
>  def download_asset_task(self, import_asset_pk):

There's a significant amount of refactoring to be done in order to use that since the requests that are getting the 429 errors are building a list of items in a while loop (synchronously). I'm not really clear on how that would work and I was hoping to not make a mini project out of this just now, since I'm also in the middle of something else.

acdha:
We can make the code easier to follow by using an [`HTTPAdapter` with `max_retries`](http://docs.python-requests.org/en/latest/api/#requests.adapters.HTTPAdapter) as per https://www.peterbe.com/plog/best-practice-with-retries-with-requests. 

I'm wondering whether we should use the Django cache to store each successful response we're getting for a fairly long period of time so we could save partial progress for a day or two and then update the `import_collection_task` task to enable retries so it could eventually make it through a large collection without re-retrieving pages. That might be enough to avoid needing to refactor that task into one task per page.

importer/tasks.py
```diff
@@ -133,22 +154,9 @@ def get_collection_items(collection_url):
...
-
-            if resp.status_code != 429 and resp.status_code != 503:
-                break
-
-            # If rate limited or service unavailable, wait and try again
-            time.sleep(2 * attempts)
-            attempts = attempts + 1
-
-        resp.raise_for_status()
+        resp = requests_retry_session().get(current_page_url)
```
acdha: We could hoist this session up to the top so we reuse it for all of the URLs but I'm not sure how much we care about that given that we're actually trying to slow things down a bit.

rstorey: I thought of that, but then I realized the other requests are relying on the celery backoff / retry mechanism. I actually was able to create that scenario in my testing and I think it's better to let celery handle it for the asynchronous idempotent stuff, rather than allow the delayed request to occupy a celery worker thread.

I put a lot of thought into whether / how we should use the Django cache for the responses of the collection pages as well - I don't think that gets us much unless we build a retry function for the main import job. You can currently "manually" retry by going through the Import Items form again and putting in the same collection URL, in which case the cached results would give some benefit, but I believe the main problem we're experiencing is that we're rate limited as we page through collection results - and this solution solves that by slowing down enough to get through all the pages. I think I'll put the cache stuff in anyway, for the manual case and just in case we build that main retry function.

acdha: Oh, yes, I was definitely thinking that as part of this we update https://github.com/LibraryOfCongress/concordia/blob/8f7525de4ce0ca901d116b4f04a23a4c8f6ecb66/importer/tasks.py#L267-L268 to enable automatic retries, perhaps with just a simple multi-minute time delay.

PR #865, opened 2019-03-26
Fix the celery conn_max_age problem in ECS environment

rstorey:
Now the same celery issue is happening in ECS. I'm not sure why. See https://crowd-sentry.loc.gov/loc/crowd-backend/issues/264

### DIFF
--- a/concordia/db_routers.py
+++ b/concordia/db_routers.py
@@ -1,0 +1,21 @@
+class AppRouter:
+    def db_for_read(self, model, **hints):
+        if model._meta.app_label == "importer":
+            return "celery"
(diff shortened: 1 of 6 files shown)


acdha:
I’m wondering about whether we want a front end performance hit for a Celery regression. How hard would it be to only do this on the importer container?

rstorey:
> I’m wondering about whether we want a front end performance hit for a Celery regression. How hard would it be to only do this on the importer container?

I didn't want to create a whole separate settings mechanism for importer / celery, so I used the multiple databases approach. Seems to work.

PR #866, opened 2019-03-27
Track user interactions with share buttons in adobe analytics

rstorey:
See #475

### DIFF
--- a/concordia/static/js/contribute.js
+++ b/concordia/static/js/contribute.js
@@ -420,6 +420,8 @@ $tagEditor
         displayMessage('error', message, 'tags-save-result');
     });
 
+/* Social share stuff */
+
 var hideTooltipCallback = function() {
     // wait a couple seconds and then hide the tooltip.
     var hideTooltip = function(tooltipButton) {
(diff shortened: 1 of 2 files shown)


concordia/static/js/contribute.js
```diff
@@ -429,13 +431,33 @@ var hideTooltipCallback = function() {
     };
     setTimeout(hideTooltip($(this)), 3000);
 };
+
+function trackShareInteraction(element, interactionType) {
+    // Adobe analytics user interaction tracking
+    if (window['loc_ux_tracking'] != undefined) {
```
acdha: This would more commonly be:

```suggestion
    if ('loc_ux_tracking' in window) {
```

concordia/static/js/contribute.js
```diff
@@ -429,13 +431,33 @@ var hideTooltipCallback = function() {
...
+        let loc_ux_tracking = window['loc_ux_tracking'];
+        loc_ux_tracking.trackUserInteractionEvent(
+            element,
```
acdha: Does this expect to get jQuery objects or DOM elements? The variable is named `element` but the values being passed in are jQuery objects. If that's correct, I'd like to rename this to something more obvious like `$element` to make it clear that it's not a normal `Element`.

rstorey: It expects jQuery objects. Changed it.

PR #868, opened 2019-03-27
Set conn_max_age to nonzero if env var is present in ECS environment

rstorey:


concordia/settings_ecs.py
```diff
@@ -87,3 +87,6 @@
     "Transcribed and reviewed by volunteers participating in the "
     "By The People project at crowd.loc.gov."
 )
+
+if os.getenv("CONN_MAX_AGE_NONZERO"):
```
acdha: Since we're not setting this to the actual value, what do you think about renaming this to something like `USE_PERSISTENT_DATABASE_CONNECTIONS`?

concordia/settings_ecs.py
```diff
@@ -87,3 +87,6 @@
     "Transcribed and reviewed by volunteers participating in the "
     "By The People project at crowd.loc.gov."
 )
+
+if os.getenv("USE_PERSISTENT_DATABASE_CONNECTIONS"):
```
acdha: I'm not sure we want to explicitly test here since we don't otherwise define this value but it might be a good future-proofing move:

```suggestion
if os.getenv("USE_PERSISTENT_DATABASE_CONNECTIONS") == "1":
```

PR #875, opened 2019-03-29
Allow admins to set the path for new simple pages

rstorey:
This bug fix will allow you to be able to set the path for the new resources page.

### DIFF
--- a/concordia/admin/__init__.py
+++ b/concordia/admin/__init__.py
@@ -398,7 +398,7 @@ class TranscriptionAdmin(admin.ModelAdmin):
 @admin.register(SimplePage)
 class SimplePageAdmin(admin.ModelAdmin):
     list_display = ("path", "title", "created_on", "updated_on")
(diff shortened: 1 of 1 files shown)


acdha:
Since we're reversing part of the decision about the simple pages’ design, do we want to go further by adding validation for the path values (either to prevent conflicts or to ensure that the route exists first), or to add this as a 404 handler so new paths can be added any time they don't conflict?

rstorey:
I'm not clear on what's being reversed here. How are admins supposed to create new simple pages with the path set to read only?

acdha:
When we added the simple page feature, I had asked about the ability to add new pages and it was agreed that this would not be allowed — only the existing static pages would be available. Since we're changing that decision we probably also want to update the UI around the paths to avoid confusion.

rstorey:
Ahhhhh. Yup, changing that decision. I will ticket this for a future sprint.

PR #880, opened 2019-04-02
Flatten, de-dupe, and browse tags

rstorey:


Pipfile
```diff
@@ -65,3 +65,6 @@ flake8-assertive = "*"
 
 [requires]
 python_version = "3.6"
+
+[pipenv]
+allow_prereleases = true
```
acdha: This looks like a regression?

Pipfile
```diff
@@ -13,7 +13,7 @@ django-haystack = "*"
 "boto3" = ">=1.9.16"
 botocore = '>=1.12.16'
 requests = "*"
-Django = ">=2.1.7"
+Django = "===2.1.7"
```
acdha: Should this be `~=2.1.8`?

rstorey: I'm resetting the pipfile for this branch... there's no reason for it to be different from master

acdha: Ah, good

rstorey:
Closing until tagging becomes a priority

PR #884, opened 2019-04-04
Use Python 3.7 everywhere

rstorey:


acdha:
I just rebased & changed the commit to be more minimal — now it only has the `python:3.7` base image change and re-syncing the `apt-get` invocations to take full advantage of Docker's layer caching.

PR #888, opened 2019-04-08
Admin reopen asset feature

### LINKED ISSUE #544: Allow admins to re-open completed assets for updates
We need some kind of admin function to be able to edit, re-open, or otherwise change status of an asset that has already gone through the workflow and been marked completed. Functionality TBD

rstorey:
Closes #544

### DIFF
--- a/concordia/templates/admin/concordia/asset/change_form.html
+++ b/concordia/templates/admin/concordia/asset/change_form.html
@@ -33,7 +33,7 @@
 
     {% if transcriptions %}
         <table>
-            <caption>Transcription History</caption>
+            <caption>Transcription History - Current status: {{ original.transcription_status }}</caption>
             <thead>
                 <tr>
                     <th>ID</th>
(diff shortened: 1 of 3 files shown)


concordia/templates/admin/concordia/asset/change_form.html
```diff
@@ -33,7 +33,7 @@
 
     {% if transcriptions %}
         <table>
-            <caption>Transcription History</caption>
+            <caption>Transcription History - Current status: {{ original.transcription_status }}</caption>
```
acdha: I'm wondering whether this should be in the history table rather than displayed above so it'd be visible even if there are no transcriptions. Perhaps it might also make sense to display them in reverse order on the theory that the most recent entry will be the most interesting.

rstorey: I moved the status out of the `if transcriptions` block so it'll always be there. I'm still curious why it doesn't show up in the default form since it's an attribute of the asset.

PR #890, opened 2019-04-10
Add reservation_token to asset transcription reservation

### LINKED ISSUE #730: Anonymous user asset reservation should be browser session based
From @acdha 

> with anonymous users the locks are all shared since they’re mapped to the same user behind the
> scenes. That probably needs to be linked to a browser session cookie so we can distinguish 
> between people who happen across the same page. The find-next-page feature skips assets with
> active registrations but it’s still possible to hit them.

DIsambiguate the anonymous user asset reservations so that different browser sessions aren't  shared behind the scenes.

rstorey:
Closes #730 

In main line of code:

- [x] Add something to asset transcription reservation that ties the reservation to the browser, not just the user
- [x] Make front end aware of reservation on a per browser / per session basis

Deferred / won't fix:
* For anonymous users, tie locks to a session ID following a successful CAPTCHA (which is also just doing things like HTTP 400s for reservation requests without a session rather than starting a new one)
* Switch reservation backend from DB to redis
* Adding some metrics for something like the number of active locks, the subset which are anonymous, number of distinct real/anonymous users, etc.


concordia/migrations/0024_auto_20190410_1213.py
```diff
@@ -0,0 +1,54 @@
...
+    operations = [
+        migrations.AlterModelOptions(
+            name="sitereport", options={"ordering": ("created_on",)}
+        ),
+        migrations.AddField(
+            model_name="assettranscriptionreservation",
+            name="session_id",
+            field=models.CharField(blank=True, max_length=50, null=True),
+        ),
+        migrations.AlterField(
```
rstorey: I'm not sure why this migration came out like this - I will look into it.

rstorey: I guess there were some changes made to the models, or changes made in the underlying Django assumptions, that caused this migration to have non-reserve-id related changes in it. I loaded a fresh database and still got the same result.

acdha: I created https://github.com/LibraryOfCongress/concordia/pull/904 to handle that separately so this PR can stay clean

concordia/views.py
```diff
@@ -1599,9 +1609,12 @@ def action_app(request):
                     ).replace("http", "ws"),
                     "campaignList": reverse("transcriptions:campaign-list"),
                 },
+                "reserveId": get_or_create_reserve_id(request),
                 "urlTemplates": {
                     "assetData": "/{action}.json",
-                    "assetReservation": "/reserve-asset/{assetId}/",
+                    # FIXME: for some reason a trailing slash on the assetReservation
```
acdha: This because the URL pattern is missing the trailing slash

concordia/urls.py
```diff
@@ -100,7 +100,11 @@
     path("contact/", views.ContactUsView.as_view(), name="contact"),
     path("act/", views.action_app, name="action-app"),
     path("campaigns/", include(tx_urlpatterns, namespace="transcriptions")),
-    path("reserve-asset/<int:asset_pk>/", views.reserve_asset, name="reserve-asset"),
+    path(
+        "reserve-asset/<int:asset_pk>/<slug:reserve_id>",
```
acdha:
```suggestion
        "reserve-asset/<int:asset_pk>/<slug:reserve_id>/",
```

concordia/models.py
```diff
@@ -326,6 +326,7 @@ class AssetTranscriptionReservation(models.Model):
 
     asset = models.OneToOneField(Asset, on_delete=models.CASCADE)
     user = models.ForeignKey(User, on_delete=models.CASCADE)
+    reserve_id = models.CharField(max_length=50, blank=True, null=True)
```
acdha: `reserve_id` scans a bit oddly to me — I'd prefer something like `reservation_id` or, since this is basically an opaque value, `reservation_token`.

I'm wondering whether `blank=True` should be here, too — it seems like we only want that during the migration phase since every subsequent reservation will have a token.

Also, this is declared as a `CharField` here but mapped in the URL as a `slug`. It seems like these should be compatible if we leave it in the URL.

concordia/urls.py
acdha: I'm wondering whether this should be in the URL since it doesn't have any meaning and, unlike the asset PKs, is unique to a user. Looking at it, it's not immediately obvious why we're even passing it around since it's stored in the user session and it seems like it could stay on the server side with no exposure to HTML/JS.

concordia/static/js/action-app/index.js
```diff
@@ -620,7 +620,8 @@ export class ActionApp {
         if (canEdit) {
             this.assetReservationURL = this.urlTemplates.assetReservation.expand(
                 {
-                    assetId: encodeURIComponent(asset.id)
+                    assetId: encodeURIComponent(asset.id),
+                    reserveId: encodeURIComponent(this.config.reserveId)
```
acdha: Should we refactor this to be `reservationToken` to match the backend name?

rstorey: Oh this needs to be taken out. On it.

acdha:
We probably should do something to bump coverage a bit for this path but my main question is just whether we want to drop the user column from reservations now or later. It would clean up a branch condition for testing and since we haven't tried to limit the number of reservations per-user yet I'm not sure whether we have any reason to keep it.

acdha:
This branch should also cherry-pick https://github.com/LibraryOfCongress/concordia/pull/932/commits/f18bb31f647908696e16adae0098e1e596cca30a

concordia/signals/handlers.py
```diff
@@ -13,6 +14,14 @@
 ASSET_CHANNEL_LAYER = get_channel_layer()
 
 
+@receiver(user_logged_in)
+def clear_reservation_token(sender, user, request, **kwargs):
+    try:
```
acdha: I don't think this matters too much but I believe this could also be written `request.session.pop("reservation_token")`

rstorey: Ah you're right. If the tests pass I'm not going to fiddle with it.

acdha: Yeah, I wouldn't spend more time on it

rstorey:
I think we broke the Github PR permissions model with n=2 developers on the team. This ok to merge now? @acdha

rstorey:
> This looks good to me but I haven't tested it in the classic interface yet

If you want to wait until you've had a chance to test it in classic mode, that's fine with me. Just go ahead and merge it when you're ready then (or let me know if more changes are needed). I did test it in classic mode, but I don't trust my own testing.

acdha:
In practice that means I won't be looking at it until Thursday, which is probably longer than you want to wait

rstorey:
Okay then - into crowd-dev it goes

PR #894, opened 2019-04-11
Group projects by Topic

### LINKED ISSUE #96: As an admin, I want to be able to apply tags to a Project
- [ ] ~~As an admin, I want to be able to apply tags to a Project~~
- [ ] ~~As an admin, I want to be able to exhibit or feature certain Project-tags~~
- [x] As a curator I want to draw attention to items from several collections. I propose a campaign that would serve this need.
- [x] As a community manager, I want to pull together some projects that need a little extra love. I don’t want them to be taken out of their parent campaigns, and I want to move them out of a campaign when they’re done
- [x] As a volunteer, I want to look through some categories or topics I can transcribe, instead of just the person’s name
- [ ] As a community manager, I want to be able to apply a category when I’m importing projects
- [x] As a community manager, I want to be able to easily edit categories and group projects for temporary needs
- [x] As a community manager, I want to be able to pull an existing project into a new campaign and I need a solution for that
- [x] As a volunteer, I want to see what themes and topics are available for me to explore and see the context of them
- [x] As a volunteer, I would like to pick from a drop down menu a topic I would like to transcribe or review (campaigns and topics in same list together)
- [x] As a community manager, I want to continue to develop the topics for our volunteers beyond just “subject” into aggregating items/projects by date, activity (almost done, etc) or challenge
- [x] Add short description and published attributes
- [x] Rename to Topic
- [x] Do a sample page of listing campaigns + topics together - URL for campaigns and topics: /campaigns-topics/
- [x] ~~Add option for progress bar in Admin view - when creating topic~~ Progress bar is always included on topic detail pages
- [x] ~~Add topics to project django admin list view~~ ManyToMany fields in django admin list views aren't supported
- [x] Add ordering to Topics model similar to Campaigns
- [x] Use the same ordering field to order things in the dropdown
- [ ] ~~Grouping in the dropdown for options~~
- [x] What should the ordering for topics be? Should there be an ordering field like campaigns have? Or do it alphabetical? What about when listing alongside campaigns (in the dropdown or on a page)?
- [ ] Let's Go link for Topics that can be used in the carousel
- [ ] Use ordering field to integrate campaign and topic list

rstorey:
Closes #96


rstorey:
> What was the purpose of moving this block of code? If we need it to happen after the topic list has been loaded we need to either have it as part of that promise chain or make it work either way.

I don't see where the campaign list is being cached. The topic list should be treated in the same way.

concordia/models.py
```diff
@@ -84,16 +84,41 @@ def get_absolute_url(self):
...
+    published = models.BooleanField(default=False, blank=True)
+
+    ordering = models.IntegerField(
+        default=0, help_text="Sort order override: lower values will be listed first"
+    )
+    title = models.CharField(blank=False, max_length=255)
+    slug = models.SlugField(blank=False, allow_unicode=True, max_length=80)
+    description = models.TextField(blank=True)
+    thumbnail_image = models.ImageField(
+        upload_to="topic-thumbnails", blank=True, null=True
```
acdha: We should probably ticket using a separate storage bucket for our non-collection media at some point since we're accumulating more of them over time.

concordia/models.py
```diff
@@ -84,16 +84,41 @@ def get_absolute_url(self):
...
+
+
 class Resource(MetricsModelMixin("resource"), models.Model):
     sequence = models.PositiveIntegerField(default=1)
     title = models.CharField(blank=False, max_length=255)
     resource_url = models.URLField()
 
-    campaign = models.ForeignKey(Campaign, on_delete=models.CASCADE)
-
-    class Meta:
```
acdha: Was it intentional to remove the unique/ordering test? The history wasn't clear to me from the commit & I was wondering whether we might have wanted to add topic into the check, enforce it in a clean method, etc.

concordia/templates/base.html
```diff
@@ -56,7 +56,7 @@ <h1 class="header-text m-0 d-none d-sm-flex align-items-center">
                             <a class="nav-link {% if PATH_LEVEL_1 == 'about'%}active{% endif %}" href="{% url 'about' %}">About</a>
                         </li>
                         <li class="nav-item">
-                            <a class="nav-link {% if PATH_LEVEL_1 == 'campaigns' %}active{% endif %}" href="{% url 'transcriptions:campaign-list' %}">Campaigns</a>
+                            <a class="nav-link {% if PATH_LEVEL_1 == 'campaigns' %}active{% endif %}" href="{% url 'campaign-topic-list' %}">Campaigns</a>
```
acdha: Are we launching this immediately without a feature-gate?

concordia/admin/__init__.py
```diff
@@ -150,26 +151,25 @@ def get_urls(self):
 
 @admin.register(Resource)
 class ResourceAdmin(admin.ModelAdmin, CustomListDisplayFieldsMixin):
-    list_display = ("campaign", "sequence", "title", "resource_url")
+    list_display = ("campaign", "topic", "sequence", "title", "resource_url")
+    list_display_links = ("campaign", "topic", "sequence", "title")
+
+
+@admin.register(Topic)
+class TopicAdmin(admin.ModelAdmin):
```
acdha: This should use `prepopulated_fields` to automatically set the slug from the title when creating a new topic

concordia/views.py
```diff
@@ -473,12 +474,128 @@ def annotate_children_with_progress_stats(children):
...
+        data["topics"] = Topic.objects.published().order_by("ordering", "title")
+
+        return render(self.request, self.template_name, data)
+
+
+@method_decorator(default_cache_control, name="dispatch")
+class TopicDetailView(APIDetailView):
+    template_name = "transcriptions/topic_detail.html"
+    context_object_name = "topic"
+    queryset = Topic.objects.order_by("title")
```
acdha: Do we want to make this use `.published()` so it'll 404 if you ask it for an unpublished topic?

concordia/static/js/action-app/index.js
```diff
@@ -442,6 +442,26 @@ export class ActionApp {
...
+                    o.textContent = topic.title;
+
+                    // TODO: this does not handle the case where the last assets of a campaign change state while the app is open
+                    Object.entries(topic.asset_stats).forEach(
+                        ([key, value]) => {
+                            o.dataset[key] = value;
+                        }
+                    );
+
+                    this.campaignSelect.appendChild(o);
```
acdha: I would prefer not to lump these into the campaign selector since it seems likely to confuse both developers and users but if we're doing that I'd think that'd be something like updating the API data to include the topic, tagging these options to indicate whether they're a campaign or a topic (and since this cost is almost entirely identical to the previous block we probably want to move it into a shared function which could be called with a class attribute for the `option` element), modifying `getVisibleAssets()` to filter topics as well as campaigns, and possibly having `updateAvailableCampaignFilters()` sort the list.

acdha:
We should have some tests covering the non-API views as well, which might be a good time to refactor the transcription status filtering view & test logic out since it's being used in multiple places.

acdha:
> I don't see where the campaign list is being cached. The topic list should be treated in the same way.

The `getCachedData()` utility returns either a `Promise` for the `fetch` operation or immediately resolves it for a cache hit:

https://github.com/LibraryOfCongress/concordia/blob/f19666e02ac7a1d341df0678eb7c7e6cf8c0b667/concordia/static/js/action-app/utils/api.js#L17-L39

The `ActionApp` class has a set of `Map` objects which are used to contain the different types of data:

https://github.com/LibraryOfCongress/concordia/blob/f19666e02ac7a1d341df0678eb7c7e6cf8c0b667/concordia/static/js/action-app/index.js#L41-L44

There's a block of accessors which are basically binding `getCachedData()` to the appropriate `Map`:

https://github.com/LibraryOfCongress/concordia/blob/f19666e02ac7a1d341df0678eb7c7e6cf8c0b667/concordia/static/js/action-app/index.js#L559-L569

concordia/models.py
rstorey: It's intentional. Either campaign or topic can be specified as a foreign key so they may also be blank.

concordia/templates/base.html
rstorey: Yes

concordia/views.py
rstorey: Yes you're right, I'll fix that.

concordia/admin/__init__.py
rstorey: I thought I already added that? If not then I will.

concordia/models.py
rstorey: So then we can't use crowd-media to point to it?

rstorey:
> > I don't see where the campaign list is being cached. The topic list should be treated in the same way.
> 
> The `getCachedData()` utility returns either a `Promise` for the `fetch` operation or immediately resolves it for a cache hit:
> 
> https://github.com/LibraryOfCongress/concordia/blob/f19666e02ac7a1d341df0678eb7c7e6cf8c0b667/concordia/static/js/action-app/utils/api.js#L17-L39
> 
> The `ActionApp` class has a set of `Map` objects which are used to contain the different types of data:
> 
> https://github.com/LibraryOfCongress/concordia/blob/f19666e02ac7a1d341df0678eb7c7e6cf8c0b667/concordia/static/js/action-app/index.js#L41-L44
> 
> There's a block of accessors which are basically binding `getCachedData()` to the appropriate `Map`:
> 
> https://github.com/LibraryOfCongress/concordia/blob/f19666e02ac7a1d341df0678eb7c7e6cf8c0b667/concordia/static/js/action-app/index.js#L559-L569

We've deferred displaying the Topic info in the more info box for now. The request that populates the campaign filter dropdown isn't cached as far as I can tell. That is the only place where the topic list is being fetched at the moment. When it gets added to the more info box, we can treat it like the other cached data.

concordia/models.py
acdha: Should we make those relationships nullable so we can still have the uniqueness constraints on non-null values?

rstorey:
> What is the intention of the campaign-topic list view? It seems like we have a problem with the topics being buried below the last campaign, and making the two display together seems a little confusing in testing — I had been under the impression that we would use them to badge the list of campaigns with topics so the relationship would be clear but the current UI in a few places making them appear to be peers, which is a bit confusing since the same project can be listed under multiple things which appear to be containers.

https://github.com/LibraryOfCongress/concordia/pull/894/commits/21f5d1892ea789d3bdd3c360a6932c9f017a4ee7 addresses that

concordia/templates/transcriptions/campaign_topic_list.html
```diff
@@ -0,0 +1,44 @@
...
+{% block main_content %}
+<div class="container py-3">
+    <h1 class="sr-only">Campaigns</h1>
+    <ul class="list-unstyled">
+        {% for campaign in campaigns_topics %}
+            <li class="p-4 mb-1 bg-footer">
+                <h2 class="h1 mb-3">{{ campaign.title }}</h2>
+                <div class="row">
+                    <a class="col-md-5 order-md-2"
+                    {% if campaign.model_type == "Campaign" %}
```
acdha: Could we replace this with a `get_absolute_url()` call? That would appear to remove the need to set the `model_type` property in both cases.

rstorey: Yes, I'll do that right now.

PR #895, opened 2019-04-15
Update dependencies

acdha:
The most important one being Jinja2, which has a flagged vulnerability although it does not appear to be relevant in our usage. 

I'm testing whether the better fix is to remove the currently-unused coreapi dependency.


acdha:
This ended up removing more dependencies which were either entirely unused or specified as both the top-level dependency and a sub-level (e.g. both django-elasticsearch-dsl and elasticsearch-dsl).

PR #907, opened 2019-04-22
Upgrade sentry SDKs

rstorey:


concordia/celery.py
```diff
@@ -2,16 +2,22 @@
...
 if SENTRY_BACKEND_DSN:
-    client = Client(SENTRY_BACKEND_DSN)
-    register_logger_signal(client)
-    register_signal(client)
+    CONCORDIA_ENVIRONMENT = os.environ.get("CONCORDIA_ENVIRONMENT", None)
+    sentry_sdk.init(
+        SENTRY_BACKEND_DSN,
+        environment=CONCORDIA_ENVIRONMENT,
+        release=get_concordia_version(),
+        integrations=[CeleryIntegration()],
```
acdha: I'm wondering whether this should also include `DjangoIntegration()` — not sure whether that has any value if you're not using it in the traditional request/response cycle

concordia/templates/base.html
```diff
@@ -171,17 +171,13 @@ <h2>Follow Us</h2>
...
-            if (navigator.userAgent.indexOf("CloudFlare-AlwaysOnline") < 0) {
-                Raven.config("{{ SENTRY_FRONTEND_DSN }}", {
-                    environment: "{{ CONCORDIA_ENVIRONMENT }}",
-                    release: "{{ RAVEN_CONFIG.release }}",
-                    ignoreUrls: [
-                        /^moz-extension/
-                    ]
-                }).install();
-            }
+            Sentry.init({
```
acdha: I _think_ we're good with the filters for known crawlers but we should confirm that this doesn't spam us with the security restrictions for the CloudFlare Always Online crawler.

rstorey: I added the condition back in - I didn't see anything in the docs about filters for known crawlers. Do you have any reference about it I could look into?

concordia/celery.py
rstorey: I don't think we need it, since there's no HTTP context.

PR #909, opened 2019-04-22
Use P1 thumbnail image for social share on asset page

### LINKED ISSUE #892: Use P1 thumbnail image in asset detail page opengraph markup
The opengraph tag in the asset detail page which links to the image should use the P1 thumbnail instead of the full size image.

See https://github.com/LibraryOfCongress/concordia/commit/478de95

rstorey:
Closes #892

### DIFF
--- a/concordia/views.py
+++ b/concordia/views.py
@@ -56,6 +56,7 @@ from concordia.models import (
     TranscriptionStatus,
     UserAssetTagCollection,
 )
+from concordia.templatetags.concordia_media_tags import asset_media_url
 from concordia.utils import get_anonymous_user, request_accepts_json
 from concordia.version import get_concordia_version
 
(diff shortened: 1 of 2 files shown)


concordia/views.py
```diff
@@ -682,6 +683,16 @@ def get_context_data(self, **kwargs):
 
         ctx["social_share_flag"] = flag_enabled("SOCIAL_SHARE", request=self.request)
 
+        image_url = asset_media_url(asset)
+        if asset.download_url and "iiif" in asset.download_url:
+            thumbnail_url = asset.download_url.replace(
+                "http://tile.loc.gov", "https://tile.loc.gov"
+            )
+            thumbnail_url = thumbnail_url.replace("/pct:100/", "/!512,512/")
```
acdha: I'm wondering whether we should use https://github.com/Princeton-CDH/piffle for this — it's more processing to parse the URLs but would give us a higher degree of confidence that it won't break unexpectedly and would be more future-proof for other outside sources.

rstorey: I ticketed that here: https://github.com/LibraryOfCongress/concordia/issues/910

PR #912, opened 2019-04-24
Take out newline in entrypoint test sentry config command

rstorey:
One more bugfix for the entrypoint script.

### DIFF
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -16,7 +16,7 @@ echo "Ensuring our base configuration is present in the database"
 
 if [ -v SENTRY_BACKEND_DSN ]; then
     echo "Testing Sentry configuration"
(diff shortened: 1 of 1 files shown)


entrypoint.sh
```diff
@@ -16,7 +16,7 @@ echo "Ensuring our base configuration is present in the database"
 
 if [ -v SENTRY_BACKEND_DSN ]; then
     echo "Testing Sentry configuration"
-    echo "from sentry_sdk import capture_message;capture_message('This is a test event')\n" | ./manage.py shell
```
acdha: This seems like it might be approaching an argument for turning this into a management command which can be linted and tested more easily

rstorey: I was kind of hoping that feature would make its way into the new sentry SDK eventually.

rstorey: I ticketed that for future work

Issue #918, opened 2019-04-25
Define navigation history behavior for new activity UI


rstorey:
Do something to make the back button at least somewhat functional in the new javascript UI experience.

acdha:
We have a larger discussion about user state before we can work on this (#917 is effectively a duplicate of this issue for the same reason). We need to figure out what state people are sharing / using — the activity queues are stable so we could persist e.g. filters, etc. but the items aren't going to be persistent across sessions unless we add support for loading specific asset metadata on-demand so we can always show an asset even if it's changed statuses without having to load a bunch of other metadata first.

acdha:
Here's a first-pass proposal:

1. Serialize state in the URL hash fragment using the query-string format
2. Every time the user changes modes or filters, we update the fragment with the current values

Subsequently:
1. Enable JSON for the asset view (currently we only go as deep as items)
2. Every time the user views an asset, we update the fragment with the current item (_not_ the full history which could quickly become huge) and have the viewer load that asset's metadata immediately rather than using the streaming data as we do normally.

One factor to consider is whether we replace or add to the history. If you add to the history it can be annoying because e.g. someone who opened it and chose review -> transcribe, sort by …, filter campaign … would need to hit back twice to go back to reviewing and three times to go back to the page they entered from. We tried pushing to the history for the WDL viewer (e.g. https://www.wdl.org/en/item/211/view/1/1/) but ended up replacing pages so if you clicked 5 pages into a book and hit back it would back to the item detail page rather than page 4, 3, 2, 1, etc. This might reasonably be different for our users but it'd be good to test.

PR #923, opened 2019-04-26
State management in the activity UI view

acdha:
* [x] Serialize UI state in the hash fragment
* [x] Load asset information on-demand for the selected asset, allowing it to be viewed even when it's the wrong status
* [x] Enable JSON format for the asset detail view

See #918

### DIFF
--- a/concordia/static/js/action-app/index.js
+++ b/concordia/static/js/action-app/index.js
@@ -6,7 +6,7 @@ import {
     unmount
 } from 'https://cdnjs.cloudflare.com/ajax/libs/redom/3.18.0/redom.es.min.js';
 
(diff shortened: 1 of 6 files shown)


acdha:
This has been “works on my Firefox” tested & could use some review. I added a callback hook I'm not entirely in love with but synchronizing other async events is not trivial. We might also want to reconsider how we're structuring the list views — we have two paths for every one which could go away if we whacked in `?format=json` to the URLs but we might also want to use that as a queue to bump the page size up significantly (which has a big impact on load performance).

rstorey:
I tested this in Chrome and so far, so good. I don't have any opinion about whether to use `?format=json` - it seems like a good idea, but low priority?

PR #924, opened 2019-04-29
Update docker-compose config

acdha:
While @klee72 and I were troubleshooting his dev setup, I noticed that the importer & celerybeat weren't starting quite right and made a few docker-compose improvements:

* The Django environment config for the main app is declared as a YAML anchor so the other Django-based service definitions can reuse it
* The Django apps are now all marked as relying on both the DB & Redis, and wait for all of their services to start before launching to reduce the restart noise
* All services are marked as `restart: unless-stopped`
* `docker-compose-wait` is installed in all of our containers in a layer-cache-friendly manner


concordia/settings_dev.py
```diff
@@ -17,12 +17,6 @@
 
 ALLOWED_HOSTS = ["127.0.0.1", "0.0.0.0", "*"]
 
-CELERY_HOST = "localhost"
```
rstorey: Were these settings causing a problem? If you try to run celery locally (a la `pipenv run celery -A concordia worker -l info`) then these need to be in there. The docker environment uses `settings_docker`.

acdha: Yes, they broke running under docker-compose and the defaults in settings_template are the same.

rstorey: Ok - I just need to add more stuff to my env then.

rstorey: Actually the `settings_dev` redis port should be 63791 (assuming you're running the DB and redis in docker) and in `settings_template` it's 6379.

rstorey: I guess I'm also a bit confused how these settings broke running under docker-compose, since that uses `settings_docker` and not `settings_dev`

acdha: 63791 is the exposed public port but it wasn't available to containers running the network compose created. I changed it to link the containers and that also removed the need to use the non-standard port so the default 6379 worked.

rstorey: It works when running everything as docker, but not when only running redis and db in docker, and django / celery / celerybeat not in docker.

acdha: I was meaning to ask whether we see enough value from the non-standard port to be worth needing to customize it. For running the apps outside of Docker, I end up needing a custom settings file anyway so I have `settings_dev_cadams.py` which can override things as needed.

rstorey: That works for me

PR #928, opened 2019-05-02
Add index for assets to ElasticSearch

rstorey:


concordia/documents.py
```diff
@@ -15,18 +15,21 @@
 site_report = Index("site_reports")
 site_report.settings(number_of_shards=1, number_of_replicas=0)
 
+asset = Index("assets")
+asset.settings(number_of_shards=1, number_of_replicas=0)
```
acdha: I don't think this is the time for it but we should probably review the number of shards

concordia/documents.py
```diff
@@ -15,18 +15,21 @@
...
+
 
 @user.doc_type
 class UserDocument(DocType):
     class Meta:
         model = User
 
-        fields = ["last_login", "date_joined", "username"]
+        fields = ["last_login", "date_joined", "username", "is_active"]
```
acdha: Do we want to filter on this or restrict the set of users to `is_active=True`? Also, would we want to filter `is_staff=False`?

concordia/documents.py
```diff
@@ -143,3 +146,44 @@ def get_queryset(self):
                 "asset__item", "asset__item__project", "asset__item__project__campaign"
             )
         )
+
+
+@asset.doc_type
+class AssetDocument(DocType):
+    item = fields.ObjectField(
+        properties={
```
acdha: Should we include the item creation/modification time?

concordia/models.py
```diff
@@ -234,6 +234,9 @@ def get_absolute_url(self):
             },
         )
 
+    def latest_transcription(self):
+        return self.transcription_set.order_by("-pk").first()
```
acdha: I'm not opposed to having this as a class method but I'm thinking that for search we might want to have a prefetch to do this in bulk — something like `prefetch_related(latest_transcriptions=Prefetch('transcription_set', queryset=Transcription.objects.order_by('pk').latest())` & an accessor but I would want to test that.

concordia/documents.py
rstorey: We need to be able to tell how many users haven't activated their accounts, so we want everybody. We also want to be able to tell if any staff are in the top contributors, so we include them too.

concordia/documents.py
rstorey: We don't have `created_on` or `updated_on` for items either.

concordia/models.py
rstorey: I like the idea of having it pre-fetch. I'm following the documentation for django-elasticsearch-dsl about "Using Different Attributes for Fields" https://pypi.org/project/django-elasticsearch-dsl/ 
If you have an alternative method to propose, I'm interested.

acdha: Have you measured performance on this currently? I'm assuming it'll be slow enough to be a concern but we should probably confirm that before spending more time on it.

rstorey: No, I used it once a week ago, since then haven't looked at it.

rstorey: It's not super slow. I was able to fully rebuild all the indices in less than 15 minutes.

acdha: Okay, let's table this until we have a problem worth attention

PR #931, opened 2019-05-03
#812 Updated gridview styling

klee72:
Summary of changes:
* Updated styling of gridview items and asset list controls layout
* Updated styling of header
* Updated styling of editor view

Outstanding items:
* Cleanup Help content markup
* Use correct social media icons
* Refine styling of image viewer controls and previews
* Make "more info" heading more prominent

Screenshots:
Screen Shot 2019-05-03 at 2 31 31 PM

Screen Shot 2019-05-07 at 11 38 55 AM


concordia/static/scss/action-app.scss
```diff
@@ -9,6 +9,10 @@ body.view-action-app {
     max-height: 100vh;
     display: flex;
     flex-direction: column;
+    font-family: 'Open Sans', sans-serif;
+    border: none;
+    min-width: map-get($container-max-widths, lg);
+    background-color: #fff;
```
acdha: This is intended to match the body background color, right? If so, think this should be `body-bg`?

```suggestion
    background-color: $body-bg;
```

concordia/static/scss/action-app.scss
```diff
@@ -60,10 +64,13 @@ main {
         width: var(--asset-thumbnail-size);
         height: var(--asset-thumbnail-size);
         object-fit: contain;
-        border-width: 3px !important;
+        border-color: #808080 !important;
```
acdha: While we're working on this, do you think these should be variables because we're going to reuse them elsewhere and/or do color math on them?

concordia/static/scss/action-app.scss
```diff
@@ -82,11 +89,19 @@ main {
     }
 
     .asset-tooltip {
-        background-color: #000;
+        margin: -5px;
```
acdha: This is cancelling out the padding on `.asset`, so perhaps we should make a variable in this block which we could use there and `#{-1 * …}` here, double it for the width/height calculations below, etc.

concordia/static/scss/action-app.scss
```diff
@@ -82,11 +89,19 @@ main {
...
+        padding: 14px;
+        background-color: #fff;
```
acdha: Same question re:`$body-bg`

concordia/static/scss/action-app.scss
```diff
@@ -82,11 +89,19 @@ main {
...
+        font-size: 13px;
```
acdha: This is 1px between `$font-size-base` and `$font-size-small`. Should we be defining it in terms of one of those variables?

The same question applies to the `line-height` below, which is just .03 smaller than `$line-height-base`

concordia/static/scss/action-app.scss
klee72: I think when we import all Bootstrap scss files, we could use the gray color themes like `$gray-400`.  But yeah I'll create the variable specifically for the prototype for now so it can replaced easily later when we install full Bootstrap.

acdha: We have all of the bootstrap variables imported now. If there's anything else we need we can add it now as well.

concordia/static/scss/action-app.scss
```diff
@@ -3,12 +3,18 @@
 @import '../../../node_modules/bootstrap/scss/mixins';
 
 $concordia-app-footer-height: $line-height-lg * 1rem;
+$font-size-xs: $font-size-base * .8125; // Move to global variables location
```
acdha: If you wanted to create something like `concordia/static/scss/_variables.scss` or `concordia/static/scss/_common.scss` I'd support doing that now. It seems like a good convention to establish.

klee72: Agree, it's been added.

concordia/settings_template.py
```diff
@@ -224,7 +224,8 @@
 # Django-specific settings above
 ################################################################################
 
-MEDIA_URL = "/media/"
+#MEDIA_URL = "/media/"
```
acdha: let’s omit changes to this file in this branch

docker-compose.yml
```diff
@@ -35,7 +35,7 @@ services:
             POSTGRESQL_PW: ${POSTGRESQL_PW}
             CONCORDIA_ENVIRONMENT: development
             S3_BUCKET_NAME: crowd-dev-content
-            DJANGO_SETTINGS_MODULE: concordia.settings_docker
+            DJANGO_SETTINGS_MODULE: ${DJANGO_SETTINGS_MODULE:-concordia.settings_docker}
```
acdha: This is getting changed in a separate branch

concordia/static/scss/_variables.scss
```diff
@@ -2,5 +2,5 @@
 $font-size-xs: $font-size-base * .8125;
 $line-height-xs: 1.4;
 
-// activity app specific variables
+// activity-ui-prototype specific variables
```
acdha: Too broad search-and-replace?

concordia/static/scss/base.scss
```diff
@@ -363,7 +363,7 @@ ul.nav-secondary {
 .btn:disabled {
     opacity: 0.3 !important;
 }
-
+/*
```
acdha: We prefer not to commit commented-out code blocks since if we need to pull something back we can get it from the Git history

concordia/static/scss/_variables.scss
```diff
@@ -0,0 +1,33 @@
...
+$concordia-app-font-size-xl: $font-size-base * 1.5;
+$concordia-app-line-height-xs: 1.4;
+
+// toolbar
+$concordia-app-toolbar-border: $gray-600;
+$concordia-app-toolbar-background: $gray-300;
+$concordia-app-active-color: $white;
+
+// thumbnail
+$concordia-app-thumbnail-width: 140px;
```
acdha: We should probably pick a name for this which conveys what's it's used for since e.g. we have asset thumbnails but also campaign thumbnails which could be displayed as part of the info. One of the other things we need to decide is how this interacts with the size adjustments, which currently uses CSS variables which are changed by the JavaScript event handler.

concordia/static/scss/action-app.scss
```diff
@@ -144,32 +228,60 @@ main {
...
         display: none;
     }
 }
 
-#asset-info-modal {
-    &.modal.show {
-        width: 100vw;
-        .modal-dialog {
-            max-width: 90vw;
+#toolbar {
```
acdha: This should have a name which describes what it applies to in more detail, especially since it's declared as a  global ID

concordia/static/scss/action-app.scss
```diff
@@ -245,15 +376,17 @@ main {
...
+        font-size: $concordia-app-font-size-xs;
     }
 }
 
-.control-toolbar {
-    display: flex;
-    flex-wrap: wrap;
-    justify-content: space-around;
-    align-items: center;
+.concordia-app-control-toolbar {
```
acdha: This should have a name which better conveys what it applies to — e.g. the asset list, the image viewer, etc. — since this app is going to expand functionality over time

concordia/templates/action-app.html
```diff
@@ -288,7 +289,7 @@ <h5>Keyboard Shortcuts</h5>
             </p>
         </div>
     </main>
-    <footer id="action-app-footer" class="footer fixed-bottom bg-light">
+    <footer id="action-app-footer" class="footer fixed-bottom bg-light" style="display: none;">
```
acdha: The `style` attribute feels like something which should be in the associated CSS in general, but in this specific case I'm wondering whether we should just delete the footer since support is being removed for it in other places.

concordia/settings_template.py
```diff
@@ -224,7 +224,8 @@
...
+MEDIA_URL = "https://crowd-media.loc.gov/"
```
acdha: These changes should be removed before this merges

concordia/templates/action-app.html
rstorey: Let's get rid of it.

rstorey: Comment out lines 538 and 348 of static/js/action-app/index.js and then the asset list will load without the footer.

acdha: Yeah, that was it – the browser errors were triggered by the code which updated the counter in the footer. This commit removes those along with the footer: https://github.com/LibraryOfCongress/concordia/commit/b5ab82e73bc5cfcf4b862a45e82d4a827dc13ead

acdha: After you've done a Git fetch, you can `git cherry-pick b5ab82e73bc5cfcf4b862a45e82d4a827dc13ead` to pull that onto your branch.

concordia/static/scss/action-app.scss
```diff
@@ -2,6 +2,27 @@
 @import '../../../node_modules/bootstrap/scss/variables';
 @import '../../../node_modules/bootstrap/scss/mixins';
 @import 'variables';
+@import '../../../node_modules/bootstrap/scss/reboot';
```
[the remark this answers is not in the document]
klee72: If our custom variables is placed before the Bootstrap, then it'd generate "Error: Undefined variable" error. If placed after, then it bypasses the mixins/functions of buttons.scss, reboot.scss and etc. Any suggestions?

concordia/templates/action-app.html
klee72: I don't see any errors. @rstorey lines 348 and 538 in `static/js/action-app/index.js` are blank for me. Can you give me little more info on this?

rstorey: > I don't see any errors. @rstorey lines 348 and 538 in `static/js/action-app/index.js` are blank for me. Can you give me little more info on this?

If you cherry-picked Chris' commit, then you don't need to do this.

rstorey:
I want to get this in crowd-dev and have a chance to test it myself before my 3pm, so I'm merging now. @acdha if you have more feedback we can do it in a new branch from master.

concordia/static/scss/_variables.scss
klee72: Name change is updated. We can talk about the size variables when we're discussing about Bootstrap theming policy.

concordia/templates/action-app.html
acdha: Are there any errors in the console?

PR #932, opened 2019-05-03
Finish activity UI asset reservations

acdha:
This is based on #890 and assumes that will be merged first

* [x] Save reservation tokens from WebSocket updates
* [x] canEdit checks whether an asset is reserved using our token
* [x] Asset list is updated as soon as we receive a status update using the full canEdit check
* [x] Asset list & viewer update when a reservation has been obtained
* [ ] Asset list & viewer update when a reservation cannot be obtained due to a conflict

### DIFF
--- a/concordia/api_views.py
+++ b/concordia/api_views.py
@@ -96,6 +96,14 @@ class APIDetailView(APIViewMixin, DetailView):
 class APIListView(APIViewMixin, ListView):
     """ListView which can also return JSON with consistent pagination"""
 
+    def get_paginate_by(self, queryset):
+        per_page = self.request.GET.get("per_page")
+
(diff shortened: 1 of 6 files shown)


acdha:
This still needs some work around what happens when you click on an item which is reserved. It doesn't incorrectly indicate that the asset is available but it shows the “asset reservation in progress” message even after getting a 409 because we need to save that affirmative response somewhere which will be checked for the list/viewer updates and I want to refactor that code more.

acdha:
This branch is now current against master after #890 merged. The status messages still need updating as mentioned in https://github.com/libraryofcongress/concordia/pull/932#issuecomment-489238191 but there might be an argument for merging it to avoid further re-reports of the same known issue.

PR #944, opened 2019-05-09
#812 Fix for review transcription text box and list asset thumbnail

klee72:
* Fix for review transcription text box breaking the layout when the content is long
* Adjusted list asset thumbnail strip to have correct padding values applied.

Screen Shot 2019-05-09 at 10 40 40 AM

### DIFF
--- a/concordia/static/scss/action-app.scss
+++ b/concordia/static/scss/action-app.scss
@@ -204,9 +204,6 @@ main {
             order: 2;
             flex-grow: 0;
             flex-shrink: 0;
-            display: block;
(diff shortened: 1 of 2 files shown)


concordia/static/scss/action-app.scss
```diff
@@ -318,9 +315,7 @@ main {
 #review-transcription-text {
     border: 1px solid $gray-800;
     padding: 10px;
-    flex-grow: 1;
-    flex-shrink: 1;
-    max-height: 100%;
+    flex: 1 1 0;
```
acdha: I generally prefer not using the shorthand forms but this is otherwise fine by me

Issue #930, opened 2019-05-02
Prototype punch list


elainekamlley:
Original PR for context: #812 

# Goals: 
* Help users stay in the zone for a particular activity
* Make it easy to pick items you want to work on even if that crosses projects
* Never leave fullscreen mode unless you navigate to an external source

## Items from original list: 
* [x] Component-ize the code for the transcribe/review panel which does the right thing for mode changes and data updates (change from plain HTML to full component) (@acdha )
* [x] Determine behavior for UI feedback after button submits - modal dialog? banner? disable / enable textarea? (@acdha )
* [x] Review OpenSeadragon controls - left/right controls, fullscreen, help (@jbresner )
     * [x] OpenSeadragon Help should open Help tab on the right
     * [x] The "home" button should be grouped with the "zoom in" and "zoom out" buttons
     * [x] The left -right arrows that allow you to go to the next tile source should be (temporarily) removed (until we restore / customize the reference strip navigator)
* [x] What should the transcription text box look like when you're reviewing? Particularly when there is nothing to transcribe? (@jbresner + dev TBD)
* [x] Consistent terminology and descriptions for interface components / help guide / documentation (@elainekamlley or @lalgee ?) -- ticketed as #871
* [x] Support for non-logged in users needs to be re-implemented -- ticketed as #999
* [x] ~~What should the adjacent pages section of the OpenSeadragon viewer look like? i.e. how do we indicate which page you're transcribing and which ones are just there for context?~~ Remove the adjacent pages of OpenSeadragon viewer for now - and the Display Region Navigator -- ticketed as #1000 
* [x] Side panel help text - to what extent does this need to be editable / customizable?

## Known bugs: 
* [x] Better looking badges / UI for when something is reserved by another user (@jbresner )
* [x] Either use the footer for additional stuff or get rid of it (@jbresner + @acdha)
* [x] Fullscreen icon is missing from mocks (@jbresner )
* [x] Get rid of Display Region Navigator
* [x] Make the Help/Filmstrip tabs mutually exclusive
* [x] Don't display difficulty score in asset card hover
* [x] Make an asset hover display that works when asset list is in film strip column -- ticketed as #998 
* [x] The ? icon next to Nothing to Transcribe doesn't do anything

## Testing feedback (can be de-duped if it is captured above)
- [x] When I zoom in and then click on a page to move it after about 30 seconds or so it resets the page zoom level (and clears the textarea box if I entered text) - ticketed here: #949 
[...]

PR #954, opened 2019-05-13
Remove display of the actual difficulty score

acdha:
The actual values are still used for sorting but now the numeric score will no longer be visible.

See #930

### DIFF
--- a/concordia/static/js/action-app/components.js
+++ b/concordia/static/js/action-app/components.js
@@ -24,10 +24,6 @@ export class AssetTooltip {
         this.el = html('.asset-tooltip', [
             html('.item-title'),
             html('.asset-title'),
-            html('.difficulty-score-container', [
-                text('Difficulty Score: '),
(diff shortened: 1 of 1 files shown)


concordia/static/js/action-app/components.js
```diff
@@ -265,7 +260,6 @@ class AssetListItem {
         this.el.classList.add('asset', 'border');
         this.el.dataset.image = thumbnailUrl;
         this.el.dataset.id = assetData.id;
-        this.el.dataset.difficulty = assetData.difficulty;
```
rstorey: Do we need this to maintain sorting functionality?

acdha: We used to but 16261847ef38b4bd36bf5fc2d553c0d1e9769706 removed that dependency because we're now using the re:dom keyed-list model which always receives our asset record instead of the DOM element. I removed it on the theory that it was work we weren't currently benefiting from.

PR #956, opened 2019-05-13
Move help center link above keyboard shortcuts

rstorey:


rstorey:
> I think `<section>` usually needs a heading, so maybe you don't need a `<section>` And now I'm looking at this, I think the entire help_content should be in `<aside>` instead of `<div>`. And maybe add `<h2 class="sr-only">Help</h2>`? something like this? sorry this should've been my task. I'll take care of it once you merge.
> 
> ```
> <aside id="help-panel">
>     <h2 class="sr-only">Help</h2>
>     <section>
>         <h3 class="h4">
> ```

Sounds good to me, thanks!

PR #957, opened 2019-05-13
Clarify asset reserved and asset reservation in progress in activity UI

rstorey:
Display different messages for when asset reservation is in progress and when it cannot be obtained.

### DIFF
--- a/concordia/static/js/action-app/index.js
+++ b/concordia/static/js/action-app/index.js
@@ -828,9 +828,14 @@ export class ActionApp {
 
         let {canEdit, reason} = this.canEditAsset(asset);
 
-        if (!this.assetReserved) {
+        if (this.assetReservationInProgress) {
             canEdit = false;
-            reason = 'Asset reservation in progress';
+            reason = 'Page reservation in progress';
+        } else {
+            if (!this.assetReserved) {
+                canEdit = false;
+                reason = 'Somebody else is working on this page.';
+            }
         }
 
         if (this.actionSubmissionInProgress) {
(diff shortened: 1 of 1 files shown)


concordia/static/js/action-app/index.js
```diff
@@ -828,9 +828,14 @@ export class ActionApp {
 
         let {canEdit, reason} = this.canEditAsset(asset);
 
-        if (!this.assetReserved) {
+        if (this.assetReservationInProgress) {
```
acdha: Couldn't this be inferred from `this.assetReservationURL` being defined?

rstorey: I don't think so, because the reservationURL stays alive as long as the viewer is open.

acdha: Maybe I'm rushing out the door, but wouldn't the sequence be something like this?

1. Open viewer
2. asset reservation URL is set if there isn't some hard policy reason blocking editing (e.g. self-review)
3. XHR happens for the reservation
4. “in progress” message displays
5. `assetReserved` is set true on success

It seemed like `assetReservationInProgress` is defined roughly the same as `this.assetReservationURL && !this.assetReserved` would be from the perspective of the UI badging.

rstorey: The asset reservation URL isn't deleted until the viewer is closed.

acdha: Isn't that what we want? If it's defined, we're at some stage in the process of attempting to get a reservation (either waiting on XHR or retrying an error) but the intent is to attempt to be able to work on it until the viewer is closed?

This is also getting improved in #964 to consolidate the different places which handle open asset checks. So far I haven't found a situation where that combination isn't enough.

rstorey: I look forward to seeing your solution on indicating to end users when the asset reservation is in progress and when it's either successful or failed.

PR #960, opened 2019-05-14
Restore psycopg2 dependency

acdha:
The psycopg2-binary package was encouraged by the deprecation warnings but is reportedly unsafe to use in practice so we'll keep the dependency for psycopg2 and ignore the deprecation warnings if something happens to install a wheel until they go away in the next release as promised.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -4,7 +4,6 @@ verify_ssl = true
 name = "pypi"
 
 [packages]
-psycopg2-binary = "*"
 gunicorn = "*"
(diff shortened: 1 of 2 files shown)


acdha:
The main caveat I have for this is that we'll want to keep an eye on the dev/test deployments after we merge it. I think it'll be fine either way — it builds for me — but it is technically changing things like the libssl psycopg2 is linked against.

rstorey:
Let's merge it now and I'll test it in crowd-dev this afternoon.

Issue #949, opened 2019-05-10
Saved transcription text clobbers textarea input in Activity UI


rstorey:
**What behavior did you observe? Please describe the bug**
The text I entered in the textarea as I was transcribing was reverted to the last saved transcription.

**How can we reproduce the bug?**

1. Load the Activity UI.
2. Click Transcribe.
3. Open an asset.
4. Enter some text in the textarea.
5. Wait about a minute.

**What is the expected behavior?**
The text I entered in the textarea should remain.

acdha:
We need to always make that update call because the text is not the only thing which can change. I think the answer is probably changing the text editor component to track the value which was set before any changes were made and then making the policy decisions about what happens in the various non-default flows (e.g. do we nag if you try to leave without saving, what happens if you lose the lock, etc.?)

rstorey:
Fixed in crowd-dev

PR #964, opened 2019-05-14
Transcription View: avoid updates for unmodified sources

acdha:
This avoids the problem in #949 where the update triggered by the reservation system would cause
the transcription view to replace any unsaved text with the original starting text. Now the update
method will not replace the text when the upstream value has not changed.

* [x] Confirm that text changes are not replaced by the reservation renewal updates
* [x] Confirm that updates made by another user are reflected when the reservation is eventually obtained
* [x] Confirm that unavailable assets will be retried 
* [ ] Confirm that the UI does not unlock before it has been updated to the latest transcription made by someone else

### DIFF
--- a/concordia/api_views.py
+++ b/concordia/api_views.py
@@ -13,7 +13,7 @@ behaviour for the non-JSON endpoint:
 The base APIViewMixin implements a base implementation of serialize_object which
 uses the generic django.forms.models.model_to_dict and can be overridden as needed.
 """
-import time
+from time import time
 
 from django.core.serializers.json import DjangoJSONEncoder
 from django.forms.models import model_to_dict
@@ -107,7 +107,7 @@ class APIListView(APIViewMixin, ListView):
     def serialize_context(self, context):
         data = {
             "objects": [self.serialize_object(i) for i in context["object_list"]],
-            "sent": int(time.time()),
+            "sent": time(),
         }
 
         page_obj = context["page_obj"]
--- a/concordia/signals/handlers.py
+++ b/concordia/signals/handlers.py
@@ -1,3 +1,5 @@
+from time import time
+
 from asgiref.sync import AsyncToSync
 from channels.layers import get_channel_layer
 from django.conf import settings
@@ -105,5 +107,6 @@ def send_asset_reservation_message(
             "type": message_type,
             "asset_pk": asset_pk,
             "reservation_token": reservation_token,
+            "sent": time(),
         },
     )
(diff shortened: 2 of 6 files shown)


rstorey:
Looks good so far.

acdha:
I'm heading out for the day; this definitely needs more testing for the sequencing around reservation races

rstorey:
I did a bit of testing in Chrome: It seems like when an asset is first opened, the reserve asset call doesn't work - there's a message in the console about reserve asset being called without an open asset. The subsequent reservation calls do work okay.

acdha:
That one was working correctly, too, but the log message was now being called in situations where it previously couldn't be. 6203cade700c36ed1f6b498b1b9ed35735206b75 and 698127394fc7959d3958dca3a82bd3eeff731370 improve the logging.

acdha:
@klee72 would it be useful to cherry-pick https://github.com/LibraryOfCongress/concordia/pull/964/commits/ca434876a1ae68e4b3c8fb7f8aa9f697ffae7231 and https://github.com/LibraryOfCongress/concordia/pull/964/commits/eb114700f8fe983442c02992a9774d788924c1c3 into a separate branch for independent review/refinement?

acdha:
I think it's probably time to review and merge this with a known caveat: the “most recently updated” sort isn't stable and so there's an annoying jump when you accept a translation because it bumps the item to the bottom of the list, triggering an on-demand load, which triggers another update. That can be improved in another branch, however.

rstorey:
Looks like the issue in #949 is still occurring in crowd-dev.

acdha:
That doesn't reproduce in Chrome or Firefox. Is it possible that the test system hadn't picked up the new code yet? It might be easiest to check by opening the dev tools and confirming that line 634 of components.js looks like this: 

https://github.com/LibraryOfCongress/concordia/blob/80124c768bc43ade9831b6af2060b45f8b00fceb/concordia/static/js/action-app/components.js#L634

rstorey:
I just reproduced it in chrome on crowd-dev and confirmed that the line you quoted is in components.js on line 658. It occurs both with not started and in progress transcriptions. Crowd-dev is at version 2f813f18 which is the latest version of the code.

rstorey:
I also reproduced in Firefox. I can make a video if you really need it.

acdha:
Reproducing this turns out to depend on what assets you're looking at. It does not affect any asset with at least some progress but the test in https://github.com/LibraryOfCongress/concordia/blob/80124c768bc43ade9831b6af2060b45f8b00fceb/concordia/static/js/action-app/components.js#L626 needs to handle the case where no transcription exists at all.

PR #971, opened 2019-05-17
“More info” improvements

acdha:
# Changes

* Increases the width of the “More Info” dialog to reduce text wrapping. This is a modest increase because making it wider would require us to start setting maximum sizes & responsiveness thresholds but there might be a merit to doing that to improve the raw metadata viewing experience.
* Changes the related links display to a list display to make better use of horizontal space by allowing long URLs to wrap under the titles when necessary. URLs will still be truncated with an ellipsis when they cannot fit on one line.
* Set max-width on the description text for readability

## Before 

More Info - before

## After

More Info - after

### DIFF
--- a/concordia/static/js/action-app/components.js
+++ b/concordia/static/js/action-app/components.js
@@ -147,7 +147,7 @@ class RelatedLink {
         this.el = html(
             'li.link',
             (this.title = html('h5.title')),
-            (this.link = html('a'))
+            (this.link = html('a', {target: '_blank'}))
         );
     }
     update(relatedLink) {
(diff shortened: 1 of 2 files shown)


rstorey:
While you're in there, do you mind adding the "?sp=[image number]" to the end of the "View this item on loc.gov" link? That way they can go straight to the same page they're on in concordia.

acdha:
@rstorey I just rebased against master & added the exact page query-string option:

https://github.com/LibraryOfCongress/concordia/pull/971/commits/b42eca7ded55bed93c92f11c1bfb022ffe47bc4e

PR #972, opened 2019-05-17
Bourne Memorial Day carousel update

lalgee:
Added for challenge launching Monday, May 20.  Please merge and add a flag to launch on Monday.


concordia/templates/home.html
```diff
@@ -26,17 +26,17 @@
...
                     </div>
                     <div class="carousel-item" data-overlay-position="top-right" data-title="Review Now" data-hero-text="Approve or correct other volunteers' transcriptions to help them cross the finish line." data-link-url="{% url 'transcriptions:redirect-to-next-reviewable-asset' 'mary-church-terrell-advocate-for-african-americans-and-women' %}">
                         <img class="d-block w-100" src="{% static 'img/homepage-carousel/review.jpg' %}" alt="Close up of hand-written text of Mary Church Terrell's first speech to the NAACP">
                     </div>
-                    <div class="carousel-item"  data-title="Where to start?" data-hero-text="Find instructions and help to get started in our Welcome Guide" data-link-url="{% url 'welcome-guide' %}">
+                    <div class="carousel-item active" data-title="Walt Whitman at 200" data-hero-text="Transcribe and review poetry, letters, and writings of Walt Whitman. Discover how he wrote and lived for yourself." data-link-url="{% url 'transcriptions:redirect-to-next-transcribable-asset' 'walt-whitman' %}">
```
rstorey: The "active" attribute needs to be removed here.

concordia/templates/home.html
```diff
@@ -26,17 +26,17 @@
...
+                        <img class="d-block w-100" src="{% static 'img/homepage-carousel/mem-day-challenge.jpg' %}" alt="Ornate left-handed penmanship specimen submitted by Alfred D. Whitehouse in which photograph portrait of Whitehouse is surrounded by caligraphy dedicating his work to his Union regiment. Whitehouse has a mustache and sideburns and wears a suit and bowtie.">
                     </div>
                     <div class="carousel-item" data-overlay-position="top-right" data-title="Review Now" data-hero-text="Approve or correct other volunteers' transcriptions to help them cross the finish line." data-link-url="{% url 'transcriptions:redirect-to-next-reviewable-asset' 'mary-church-terrell-advocate-for-african-americans-and-women' %}">
                         <img class="d-block w-100" src="{% static 'img/homepage-carousel/review.jpg' %}" alt="Close up of hand-written text of Mary Church Terrell's first speech to the NAACP">
                     </div>
-                    <div class="carousel-item"  data-title="Where to start?" data-hero-text="Find instructions and help to get started in our Welcome Guide" data-link-url="{% url 'welcome-guide' %}">
```
rstorey: If you're going to want this carousel slide restored at any point in the future, I suggest saving the copy to Confluence page somewhere - getting it out of Git history is a pain.

elainekamlley: Thats a good idea. @lalgee we can create a page in confluence under site content to keep carousel slides so we can rotate them in and out without having to redoing or cherry picking from past commits.

lalgee: Agreed.  I wondered about this and had already grabbed the Jump in code just in case.  Made space for these on the page I just created w/carousel update instructions. - 
https://staff.loc.gov/wikis/display/CS/Carousel+updates

rstorey:
In the future, release patches like this need to be made against the release branch rather than master.

acdha:
Is the intention that this replaces the Whitman image or is just an addition? 

Does “add a flag” mean we should display the Whitman slide until then or be one short until Monday?

Does “add a flag” mean “Add a feature flag which I will manually toggle on Monday” or “add a flag which will become true at some time on Monday” and, in the latter case, what time should that be?

concordia/templates/home.html
acdha: Getting stuff out of Confluence is also going to be a problem since it'll require manual review to keep it in sync with other updates to that page. I'm wondering whether the right answer is not to remove slides but feature-flag them.

lalgee:
this replaces whitman while we run a challenge from May 20-27.  then i will take it out and add whitman back in. Display the Whitman slide until then.

The flag request came from @elainekamlley who trained me on carousel updating this morning. But I believe either manual or automatic toggling would be fine. We just want it live Monday morning.  Ideally by 8 am.  As early as Sunday night is fine.

elainekamlley:
> In the future, release patches like this need to be made against the release branch rather than master.

Ah, did not know that. @lalgee we can make that update in the documentation you created

acdha:
That sounds like we should leave the Whitman markup alone and just make this the other branch of the feature flag test

elainekamlley:
Closing - work was completed in #974

PR #973, opened 2019-05-17
Don’t overwrite unsaved text on an empty asset

acdha:
The check in #949 needed to handle the case where an asset has not been transcribed at all, which caused the boolean test of `this.currentAsset.latest_transcription` to fail. This commit handles that case and cleans up the control flow.


acdha:
One thing we might want @klee72 to look at is reviewing the CSS for the read-only transcription input state. 7d337bbc589e29857ed3e854c95b2b052b1a9e2d restores Bootstrap's default styling so the `<textarea>` visually changes when it's disabled but that could probably be further improved.

PR #975, opened 2019-05-17
Use pikaweb to pull in ES6 modules

acdha:
These can be served natively in an HTTP/2-friendly manner while still getting the usual NPM process for audits & updates.

## TODO

* [ ] Decide whether we'll handle old packages through any means other than continuing to use CDNJS
* [ ] Decide whether we want to bundle our code or workaround Django staticfiles not rewriting ESM imports.

### DIFF
--- a/concordia/static/.gitignore
+++ b/concordia/static/.gitignore
@@ -1,0 +1,1 @@
+web_modules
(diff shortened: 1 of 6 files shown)


rstorey:
I think we basically decided not to go this route

acdha:
I think we should plan to come back to that as the number of dependencies which are not available as ESM imports goes down.

PR #981, opened 2019-05-21
Use single tile source for openseadragon viewer

rstorey:
If you load multiple tile sources with sequenceMode set to False, it displays the last one and ignores the initialPage setting.

### DIFF
--- a/concordia/static/js/action-app/components.js
+++ b/concordia/static/js/action-app/components.js
@@ -784,45 +784,26 @@ class ImageViewer {
     }
 
     update(asset) {
-        let tileSources = [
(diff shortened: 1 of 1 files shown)


concordia/static/js/action-app/components.js
```diff
@@ -784,45 +784,26 @@ class ImageViewer {
     }
 
     update(asset) {
-        let tileSources = [
+        let tileSource = [
```
acdha: I think this could technically not be an array when there's only one source (e.g. `tileSource = {"type": "image", …};`) but I'm not sure it's worth touching.

Issue #986, opened 2019-05-21
Transcription text in activity UI doesn't match text in classic UI


rstorey:
**What behavior did you observe? Please describe the bug**
* The latest version of the transcription text is not always loaded - it does not match what is shown in the classic UI. 

**How can we reproduce the bug?**

1. Go to the activity UI in Chrome browser.
1. Click Transcribe.
1. Sort by most difficult in order to get some assets in the list that have multiple transcription text versions.
1. Copy and paste the URL and open it in Safari and Firefox. 
1. Note the transcription text and refresh the browser.
1. Refresh the URL in Chrome as well.

**What is the expected behavior?**

The text for the current asset should be displayed filling the available content area

**Got screenshots? This helps us identify the issue**
![Screen Shot 2019-05-21 at 5 41 52 PM](https://user-images.githubusercontent.com/5270415/58132632-ce4c0700-7bef-11e9-97a2-9a884bcb88ba.png)

**Desktop (please complete the following information):**

-   OS: Mac
-   Browse: Chrome, Safari, Firefox

**Additional context**
* This is the URL in the screenshot: https://crowd-dev.loc.gov/act/#mode=transcribe&sort=hardest&asset=40967

*  the data is always correctly loaded (confirm with `actionApp.assets.get(actionApp.openAssetId).latest_transcription`) but the display varies:
* #991: A CSS regression causes the text area to be only a couple of lines tall (previously it stretched to fill the full height), and the browsers are not 100% consistent in the default scroll positions. We need to fix this anyway.
* When an asset is already reserved by someone else, the initial attempt to get the lock fails and that disrupts changing the textarea value from whatever the previous asset had been. When I tested it, this resolved when it got the lock.

rstorey:
As you can see in the screenshot I posted, I had the reservation lock but the transcription text was an older version and not the latest text.

acdha:
I changed the description to match what happened in the debugger. We need to restore the earlier CSS (not sure if @klee72 has time) to get the height to be consistent and we need to make sure that failing to reserve the asset still updates the textarea with the current item rather than leaving the previous contents unmodified. This should probably also clear it on close so it's more obvious and there's no delay.

acdha:
Following the instructions given, I can't reproduce any case where an older version of the same asset is displayed. #991 makes the display consistent and the text is identical for just viewing. If you make changes in one window, they will not _display_ until you either obtain the asset lock after moving away in the other session or click on the asset again.

[...]

PR #992, opened 2019-05-22
Improved contention display

acdha:
While attempting to reproduce #986, I noticed a couple of areas where the UI can be nicer:

* [x] The viewer can display updated transcription text immediately even if you can't make changes (currently it updates as soon as you get the reservation).
* [ ] Handle the 409 reservation errors more gracefully than the generic error banner — probably just by suppressing them unless we want to do something like display a “Will try again in n seconds…” timer message with a “try right now” button


rstorey:
Created #995 for the remaining task.

PR #994, opened 2019-05-22
Correct latest_transcription retrieval for asset bulk views

### LINKED ISSUE #986: Transcription text in activity UI doesn't match text in classic UI

acdha:
The previous version of this did not calculate the latest transcription PK  correctly and so the retrieved values didn’t match what you’d receive through other views. 

Closes #986

### DIFF
--- a/concordia/tests/test_api_views.py
+++ b/concordia/tests/test_api_views.py
@@ -1,0 +1,88 @@
+from django.test import TestCase, override_settings
+from django.urls import reverse
+
+from concordia.models import Asset, Item, Transcription, TranscriptionStatus
+from concordia.utils import get_anonymous_user
+
+from .utils import JSONAssertMixin, create_asset, create_item, create_project
+
+
+@override_settings(RATELIMIT_ENABLE=False)
+class ConcordiaViewTests(JSONAssertMixin, TestCase):
+    @classmethod
+    def setUpTestData(cls):
+        anon_user = get_anonymous_user()
+
+        project = create_project()
+
+        cls.items = [
+            create_item(
+                item_id=f"item_{i}", title=f"Item {i}", project=project, do_save=False
+            )
+            for i in range(0, 3)
+        ]
+        Item.objects.bulk_create(cls.items)
+
+        cls.assets = []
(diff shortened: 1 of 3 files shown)


acdha:
@rstorey I marked this as ready for review because it fixes the specific problem we identified. I want to add more test coverage for some of the other API views (e.g. the campaign, project, and item endpoints) but I need to leave now. If you want to add that feel free otherwise I'm planning to fit it in tomorrow.

acdha:
One thing this did touch is that we might want to consider removing the `.json` URL paths in favor of requiring use of `format=json` in the query-string or add a test which simply confirms that e.g. calling `/review.json` calls the same code as `/review/?format=json`:

https://github.com/LibraryOfCongress/concordia/blob/152269eb2fbc12f945707f0da6a52e4cc7191482/concordia/urls.py#L121-L130

PR #996, opened 2019-05-23
Add more test coverage for API endpoints

acdha:
Stems from #994 

Add more test coverage for API endpoints
- [x] Campaign
- [x] Project
- [x] Item


acdha:
The changes for the response formats in the API views were for fields which weren't used directly by the existing HTML templates. There's probably an optimization pass to have the templates start using the same variables rather than calling `asset_media_url` again but it's cheap and I think we have other parts of that pipeline we want to change such as simply generating the URLs from the sequence numbers and/or switching to IIIF.

rstorey:
Bug in GitHub UI won't let me leave an approving review.

Issue #920, opened 2019-04-25
Add banner to asset detail page to try the new thing


rstorey:
**User story/persona**
As a user, I want to switch from the old UI to the new UI while I'm on a specific asset.

**Acceptance Criteria**

- [x] If the advertise feature flag is enabled, there should be a banner on the asset detail page that provides a link to try the new UI.
- [x] The link to try the new UI should (somehow) load the asset the user was on as the currently-opened asset.
- [x] There should be a way to dismiss the banner once and for all (until browser cookies are cleared).

Caveat: no banner when looking at completed pages.

acdha:
We should decide whether this is always visible or the user can hide it

rstorey:
I think the user should be able to hide it (once and for always).

rstorey:
![Screen Shot 2019-05-24 at 2 39 16 PM](https://user-images.githubusercontent.com/5270415/58349651-df845600-7e31-11e9-9d12-2e16f6ecec51.png)

Does this look for Tuesday? @elainekamlley @klee72

PR #1006, opened 2019-05-23
add banner to classic asset detail page 

rstorey:
Refs #920

"Try the new interface" banner added to asset detail page for transcribe-able and review-able pages only.


rstorey:
> Do we want to add some sort of “Stop asking” option if you don't want to see this again?

Yes, I'm going to add that tomorrow.

Issue #998, opened 2019-05-22
Make a grid item hover that works for smaller sizes


rstorey:
> > * Make an asset hover display that works when asset list is in film strip column
> 
> Based on the way the sizes are contained in the list, I suspect the answer is either going to be ruthlessly truncating the amount of information or changing it to be `position:absolute` so it can be larger than the containing element, but that will require a bit of code (maybe just reusing the Bootstrap popover we're already using in other places?) to update it when the list position changes.

I think Bootstrap popover is the better option (left direction) as 'position: absolute' would look little wonky no matter how we place it. Another solution I can think of is that we could just rely on `title` attribute and hide the `.asset-tooltip` on edit view.

_Originally posted by @klee72 in https://github.com/LibraryOfCongress/concordia/issues/930#issuecomment-494802772_

Ki also noted that the same problem occurs when using smaller thumbnail sizes in the Browse All grid view.

rstorey:
This is fixed as of #1045

PR #1009, opened 2019-05-23
Basic Popper.js re-implementation of asset tooltips

acdha:
This attempts to display the tooltip to the left of the current asset, but Popper.js is smart enough not to fix it there if that would keep it offscreen. 

There’s just enough CSS to add a border — this could probably use some additional TLC for more distinct colors and we probably want to adjust widths and font sizes, too, along with possibly using some of [the available options](https://github.com/FezVrasta/popper.js/blob/master/docs/_includes/popper-documentation.md) to adjust offsets for some of the overlap situations.

See #998

## Before

image

## After

image
image

### DIFF
--- a/concordia/static/js/action-app/components.js
+++ b/concordia/static/js/action-app/components.js
@@ -378,7 +378,7 @@ export class AssetList extends List {
                 tooltip.update(asset);
                 mount(target, tooltip);
                 popper = new Popper(target, tooltip.el, {
-                    placement: 'left'
+                    placement: 'bottom'
                 });
             }
         };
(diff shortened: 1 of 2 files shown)


klee72:
@acdha, is this going to be included in the Tuesday launch? If so, I don't really have the time for refinement work but I can play with the style and options next week some time.

acdha:
@klee72 probably not unless we want to put some time into styling it. I don't think it's a show-stopper.

acdha:
I made two changes to attempt to position the tooltip under the asset image (it still goes to the sides when the bottom would be outside of the viewport) and to make it a little more distinct which asset is being worked with:

Screenshot 2019-05-24 16 37 14

One question which this raised was whether we still want to preserve the `asset-active` class when the viewer is closed. It's not immediately obvious why that one asset is orange in the big list but it does make it easier to see the last page you were looking at.

PR #1012, opened 2019-05-24
Add button to dismiss try-the-new-thing banner

### LINKED ISSUE #920: Add banner to asset detail page to try the new thing

rstorey:
Closes #920 

Looks like this: 
Screen Shot 2019-05-24 at 2 39 16 PM

### DIFF
--- a/concordia/templates/transcriptions/asset_detail.html
+++ b/concordia/templates/transcriptions/asset_detail.html
@@ -519,7 +519,7 @@
 <script src="{% static 'js/contribute.js' %}"></script>
 <script src="{% static 'js/asset-reservation.js' %}"></script>
 <script>
-    if(localStorage.getItem("no-interface-banner")==null) {
+    if(!localStorage.getItem("no-interface-banner")) {
         $("#new-interface-banner").attr("hidden", false);
     }
     $("#no-interface-banner").click(function() {
--- a/concordia/templates/transcriptions/asset_detail.html
+++ b/concordia/templates/transcriptions/asset_detail.html
@@ -32,10 +32,11 @@
 {% flag_enabled 'ADVERTISE_ACTIVITY_UI' as ADVERTISE_ACTIVITY_UI %}
 
 {% if ADVERTISE_ACTIVITY_UI and activity_mode %}
-    <div class="alert alert-dark alert-dismissible w-100" role="alert">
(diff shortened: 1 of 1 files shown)


concordia/templates/transcriptions/asset_detail.html
```diff
@@ -517,6 +518,16 @@ <h5 id="captcha-modal-title" class="modal-title">Please confirm you are not a ro
 <script src="https://cdnjs.cloudflare.com/ajax/libs/openseadragon/2.4.0/openseadragon.min.js" integrity="sha256-grVJpqgDSGBx1q2llmvY+J5zU9hEHbO7UvftFY2bK1w=" crossorigin="anonymous"></script>
 <script src="{% static 'js/contribute.js' %}"></script>
 <script src="{% static 'js/asset-reservation.js' %}"></script>
+<script>
+    if(localStorage.getItem("no-interface-banner")==null) {
```
acdha: I'm wondering whether this should be less strictly a null check:

```suggestion
    if(!localStorage.getItem("no-interface-banner")) {
```

acdha: I haven't tested in awhile but wanted to confirm that this doesn't trigger a prompt by default in any of our supported browsers. It's possible that it'll fail for other reasons but [the code to detect that is a bit complex](https://developer.mozilla.org/en-US/docs/Web/API/Web_Storage_API/Using_the_Web_Storage_API#Testing_for_availability) and I think the failure mode in this case is probably better than trying something using cookies.

acdha: I confirmed that this doesn't affect any of our supported browsers but confirming that on mobile reminded me that we might want to leave it hidden when “mobile” is in the user-agent and/or the screen size is below the 1024x768 threshold we're currently focused on:

https://github.com/LibraryOfCongress/concordia/blob/fc5cc5da4787280431592d5f46b7e81b5495baac/concordia/templates/action-app.html#L27

rstorey: Done

PR #1015, opened 2019-05-24
Reduce coverage stat churn

acdha:
We need to pick a good front-end test coverage option but until we do so
this is just adding noise


acdha:
Okay, this isn't why some builds which don't touch Python randomly report different coverage numbers. Maybe it's something about which Python version it runs on?

PR #1034, opened 2019-05-31
Gracefully handle invalid modes

acdha:
This gracefully handles typos like `https://crowd.loc.gov/act/#mode=review?loclr=twcwd` which caused the first asset load to fail because it's given a mode which could not normally occur.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -49,16 +49,16 @@ repos:
           - id: eslint
             additional_dependencies:
                 - eslint@5.16.0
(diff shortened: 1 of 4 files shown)


rstorey:
Should we not check if the mode value contains transcribe or review, rather than checking for exact matches? That way we could still load the requested mode even with the incorrect URL.

acdha:
My feeling was that it was better just to fallback to review mode than to attempt to salvage it but since we're talking n=2 here we could do something like `mode.contains("transcribe")` in an attempt to work both ways.

rstorey:
Works for me

Issue #1032, opened 2019-05-31
Add a retry for API failures


acdha:
We should update the promise chain in `fetchJSON` to more gracefully handle failures: I'm wondering whether that is adding a few time-delayed retries to everything using [`fetchJSON`](https://github.com/LibraryOfCongress/concordia/blob/f19666e02ac7a1d341df0678eb7c7e6cf8c0b667/concordia/static/js/action-app/utils/api.js#L11 ) or specifically in calls such as [`fetchAssetPage()`](https://github.com/LibraryOfCongress/concordia/blob/f19666e02ac7a1d341df0678eb7c7e6cf8c0b667/concordia/static/js/action-app/index.js#L520) but we can make the experience smoother.

This has apparently happened on a dozen times or so: https://errorlogging.loc.gov/share/issue/836067cff35041759988db0700417796/

rstorey:
This is good-enough done for now. I will test it along with the other changes made in Sprint 26.

PR #1035, opened 2019-05-31
Retry fetch errors

acdha:
We can improve the experience for transient failures (see #1032) by retrying failed fetches automatically rather than waiting for user action. This PR currently does a basic implementation which makes `fetchJSON()` retry a certain number of times before aborting but I marked this as a draft because a) it's non-trivial to test and we should probably build in some tricks to make that easier (I've been wondering about having a debug mode for awhile) and b) I was debating whether the best answer is “retry n times”, “keep retrying endlessly with a cap at a suitably large back-off value”, “retry n times before triggering a UI cue and restarting the process”, etc.

### DIFF
--- a/concordia/static/js/action-app/index.js
+++ b/concordia/static/js/action-app/index.js
@@ -518,6 +518,13 @@ export class ActionApp {
         let startingMode = this.currentMode;
 
         return fetchJSON(url)
+            .catch(error => {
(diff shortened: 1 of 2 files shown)


rstorey:
Would you please rebase this against the latest master when you have a chance? Thanks.

PR #1045, opened 2019-06-03
#998 Restyled grid hover text container

klee72:
I worked with @jbresner on this one. It was decided that hover text container should just be extended out if it's larger than the parent container. 

1. Removed Popper script
2. Added hover style
3. Fixed the stacking order of active state grid

Screenshots:
Screen Shot 2019-06-03 at 3 17 17 PM
Screen Shot 2019-06-03 at 3 17 38 PM
Screen Shot 2019-06-03 at 3 18 03 PM

### DIFF
--- a/concordia/static/scss/action-app.scss
+++ b/concordia/static/scss/action-app.scss
@@ -117,12 +117,11 @@ main {
 
     .asset {
         display: flex;
+        align-items: flex-start;
         width: var(--asset-thumbnail-size);
         height: var(--asset-thumbnail-size);
         object-fit: contain;
-
         border: $border-width solid $gray-600;
-
         background-color: #f6f6f6;
         background-size: contain;
         background-repeat: no-repeat;
@@ -141,6 +140,8 @@ main {
         &.asset-active {
             outline: 2px solid $orange;
             outline-offset: -2px;
+            position: relative;
+            z-index: 1;
         }
 
         &[data-unavailable] {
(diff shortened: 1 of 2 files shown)


acdha:
We should probably retire this branch name to avoid re-opening & closing it repeatedly.

concordia/static/scss/action-app.scss
```diff
@@ -186,23 +187,15 @@ main {
...
-
-        overflow: hidden;
-
+        border: 2px solid $blue;
         pointer-events: none;
-
         background-color: $white;
-
-        font-size: $concordia-app-font-size-xs;
+        font-size: $concordia-app-font-size-xxs !important;
```
acdha: Instead of using `!important`, why not delete the declaration on https://github.com/LibraryOfCongress/concordia/blob/cbe3bc1bbcd91b3fa3261fce306ccf02acee9257/concordia/static/scss/action-app.scss#L284-L286 which is setting `font-size: smaller` on the same element?

```suggestion
        font-size: $concordia-app-font-size-xxs;
```

concordia/static/scss/action-app.scss
```diff
@@ -186,23 +187,15 @@ main {
     }
 
     .asset-tooltip {
-        width: calc(100% + ($asset-thumbnail-gap * 2));
-        height: calc(100% + ($asset-thumbnail-gap * 2));
-
-        margin: $asset-thumbnail-gap * -1;
         padding: 14px;
```
acdha: Should this value also change? On the small size it means there's a pretty large border (~20% of the total width) around the text which looks a bit wasteful.

klee72: I'll leave this to @jbresner

rstorey:
Merging a little prematurely because I need to get 0.5.1 out the door. Please continue considering the requested change and we can reopen for anything that remains to do on this.

concordia/static/scss/action-app.scss
jbresner: That's my fault. I never mocked that up and only had the warning label displayed ("Another person is working on this asset") on the very small size displayed in the asset list. However, I think the padding for the warning label on the small size is closer to what @acdha was suggesting for the asset title. @klee72 offered to go ahead and make the change.

PR #1053, opened 2019-06-05
Use optgroup to distinguish between campaigns and topics 

rstorey:
Use optgroup to distinguish between campaigns and topics in activity UI dropdown.

### DIFF
--- a/concordia/static/js/action-app/index.js
+++ b/concordia/static/js/action-app/index.js
@@ -821,8 +821,8 @@ export class ActionApp {
     }
 
     assetHasTopic(asset, topicId) {
-        for (let topic in asset.topics) {
-            if (topic.id === topicId) {
+        for (let idx in asset.topics) {
+            if (asset.topics[idx].id === topicId) {
                 return true;
             }
         }
--- a/concordia/static/js/action-app/index.js
+++ b/concordia/static/js/action-app/index.js
@@ -107,6 +107,11 @@ export class ActionApp {
 
     addToState(key, value) {
         this.persistentState.set(key, value);
+        if (key == 'campaign') {
+            this.persistentState.delete('topic');
+        } else if (key == 'topic') {
+            this.persistentState.delete('campaign');
+        }
         this.serializeStateToURL();
     }
 
@@ -424,6 +429,9 @@ export class ActionApp {
         this.campaignSelect = $('#selected-campaign');
         fetchJSON(this.config.urls.campaignList)
             .then(data => {
+                let campaignOptGroup = document.createElement('optgroup');
+                campaignOptGroup.label = 'Campaigns';
+                this.campaignSelect.appendChild(campaignOptGroup);
                 data.objects.forEach(campaign => {
                     let o = document.createElement('option');
                     o.value = campaign.id;
@@ -436,7 +444,7 @@ export class ActionApp {
                         }
                     );
 
-                    this.campaignSelect.appendChild(o);
+                    campaignOptGroup.appendChild(o);
                 });
             })
             .then(() => {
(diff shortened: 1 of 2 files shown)


concordia/static/js/action-app/index.js
```diff
@@ -792,6 +805,28 @@ export class ActionApp {
         return sortBy(assetList, keyFromAsset);
     }
 
+    getCampaignOrTopic() {
```
acdha: I'm wondering about the name for this – it read at first like an accessor but it's really something which gets us the type of the currently selected `<option>`. Maybe something like `getSelectedOptionType()`?

concordia/static/js/action-app/index.js
```diff
@@ -839,6 +882,13 @@ export class ActionApp {
                     continue;
                 }
 
+                if (
+                    currentTopicId &&
+                    !this.assetHasTopic(asset, currentTopicId)
```
acdha: I'm wondering whether this should simply be `!asset.topics.contains(currentTopicId)`

concordia/static/js/action-app/index.js
```diff
@@ -807,10 +842,18 @@ export class ActionApp {
...
-        let currentCampaignId = this.campaignSelect.value;
-        if (currentCampaignId) {
+        let currentCampaignSelectValue = this.campaignSelect.value;
+        let currentCampaignId;
+        let currentTopicId;
+        if (currentCampaignSelectValue) {
+            let campaignOrTopic = this.getCampaignOrTopic();
+
             // The values specified in API responses are integers, not DOM strings:
-            currentCampaignId = parseInt(currentCampaignId, 10);
```
acdha: We could remove this block by converting `currentCampaignSelectValue` first so there's only one `parseInt` call.

concordia/views.py
```diff
@@ -1717,6 +1717,14 @@ def serialize_object(self, obj):
             },
         }
 
+        if project.topics:
+            metadata["topics"] = {}
```
acdha: Is this supposed to be a list of dictionaries? The code below will lose data if there's more than one topic.

concordia/static/js/action-app/index.js
rstorey: Only if I flatten `asset.topics` to a plain array of just IDs - it's currently an array of dictionaries.

concordia/static/js/action-app/index.js
rstorey: I can de-dupe the `parseInt` call but I only want `currentCampaignId` to be defined if it's a campaign that was selected and not a topic.

concordia/static/js/action-app/index.js
```diff
@@ -792,6 +805,28 @@ export class ActionApp {
         return sortBy(assetList, keyFromAsset);
     }
 
+    getCampaignOrTopic() {
+        let campaignOrTopic = 'campaign';
+
+        if (
+            this.campaignSelect.options[this.campaignSelect.selectedIndex]
+                .parentElement.label == 'Topics'
```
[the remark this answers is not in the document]
rstorey: > Reviewing this is making me think that we should rename the campaign selector to a less-specific name now that it's doing double-duty but am not coming up with non-awkward generic language. Maybe something like “containerFilter”?

I hear you on this, but I think we should wait until we see what happens next and whether we're going to refactor Campaigns/Topics so that the whole design is cleaner.

concordia/views.py
rstorey: Fixed

concordia/static/js/action-app/index.js
rstorey: Done

concordia/static/js/action-app/index.js
acdha: The de-dupe is what I was thinking about. Leaving the variables is a little busy but the ways to refactor that would either take repeated extra string comparisons against `campaignOrTopic`, function calls, etc. and I'd prefer not to tackle something bigger yet.

concordia/static/js/action-app/index.js
acdha: Ah, right – I hadn't gotten that far in the code yet. It might be nice to avoid a check-at-a-distance but I'm not sure that's worth changing `topics` into a dictionary.

concordia/static/js/action-app/index.js
```diff
@@ -792,6 +806,29 @@ export class ActionApp {
         return sortBy(assetList, keyFromAsset);
     }
 
+    getSelectedOptionType() {
+        let campaignOrTopic = 'campaign';
+
+        if (
+            this.campaignSelect.options[
+                this.campaignSelect.selectedIndex
+            ].parentElement.classList.contains('topic-optgroup')
```
acdha: I'd probably favor a class name which doesn't describe the tag (e.g. `topics`) but I'm not sure it's that big a deal.

rstorey:
Okay, I'm going to merge this and test in crowd-dev, then prep the release and test in crowd-stage. If all goes well it's going out by COB today.

PR #1058, opened 2019-06-11
Admin content editing enhancements

### LINKED ISSUE #1061: Restrict script and style tags in any places where admin can input HTML
> The format changes seem reasonable but since this allows substantially more scope for mistakes which might not be obvious, I'm wondering whether we should enable CodeMirror for the admin change UI and enable the HTML linter (see https://codemirror.net/doc/manual.html#addon_lint) at the very least. There might be an argument for restricting the contents to, say, not allow `<script>` or `<style>`, too, to limit the damage if someone's account is compromised.

_Originally posted by @acdha in https://github.com/LibraryOfCongress/concordia/pull/1058#pullrequestreview-250576647

rstorey:
* [x] Convert all description, short_description, and body fields to HTML
* [x] Add admin rich editing and HTML sanitization for HTML fields (closes #1061)
* [x] Enable `lineWrapping` the rich editor
* ~~Allow Django template markup in simple content blocks~~


rstorey:
> > The other thing in this PR is enabling lineWrapping in CodeMirror which is used for SimplePage editing.
> 
> We might want this to be in a separate commit where we can confirm that it doesn't change any of our existing content but merge it while we work on HTML editing.

It's a separate commit, but it's not a separate PR. I can make it a separate PR if you think that's useful.

acdha:
Ah, yes, I meant PR – basically it’s a question of whether we want to launch this without rich HTML editing support. I’m inclined to say we should at the very least enable CodeMirror + syntax highlighting since that’ll catch basic errors (i.e. it highlights mismatched / missing end tags) before merging.

acdha:
I just pushed a commit which enables the HTML linter. One thing we will want to do is possibly to consider always running Prettier (on the change event?) to catch minor things but I noticed that we are not currently catching the validation exception which Prettier will throw when you ask it to parse invalid HTML (e.g. `foo</div>`). I think simply catching that and displaying it is enough.

acdha:
I took a basic pass at making CodeMirror line markers for the syntax errors:

Screenshot_2019-06-19 Change project Concordia

rstorey:
Looks good to me. Thanks for doing that.

acdha:
I just rebased this against master. We may still want to consider whether this gets enabled for short descriptions but it's in a fairly good place right now — we also have an updated Prettier now. The only thing I couldn't see is an obvious way to have Prettier do HTML formatting for HTML embedded in the Markdown pages but that's not a deal-breaker.

acdha:
I'm ambivalent about https://github.com/LibraryOfCongress/concordia/pull/1058/commits/c38c10fa7d7ab6a3b295eba3194cd0f9ee41d34f — this allows using template tags but increases the possibilities for accidental breakage. It looks like the only thing we're using that for is URL reversing, which makes me wonder whether it'd be better not to have this feature and instead make sure everyone knows to use relative URLs (which could even be enforced as a validation rule to avoid hard-coding links to the staging servers).

rstorey:
> I'm ambivalent about [c38c10f](https://github.com/LibraryOfCongress/concordia/commit/c38c10fa7d7ab6a3b295eba3194cd0f9ee41d34f) — this allows using template tags but increases the possibilities for accidental breakage. It looks like the only thing we're using that for is URL reversing, which makes me wonder whether it'd be better not to have this feature and instead make sure everyone knows to use relative URLs (which could even be enforced as a validation rule to avoid hard-coding links to the staging servers).

I agree - let's train everyone to use relative URLs so we don't get in trouble with accidental breakage.

acdha:
Since we decided not to add support for embedded templating, I added a data migration and converted the only usage (`{% url "help-center" %}`) with relative URLs.

concordia/models.py
```diff
@@ -362,6 +362,7 @@ class SimpleContentBlock(models.Model):
     updated_on = models.DateTimeField(editable=False, auto_now=True)
 
     label = models.CharField(
+        unique=True,
```
acdha: @rstorey this raised one question: should this just be a `SlugField`?

rstorey: Yes that's a good idea

acdha: This is done and I squashed the migrations since they've never lived outside of this branch

Issue #901, opened 2019-04-18
Enhancements to admin text editing functions


rstorey:
**User story/persona**
- [x] As a community manager, I want the text in the simple / static page editor to soft wrap so I don't have to scroll horizontally.
- [x] As a community manager, I want to be able to provide links to content within the campaign description (and project description) and add breaks
- [x] ~~As a community manager, I want to display a system message in a banner on (some? all?) pages~~ ticketed in #1059 
- [x] As a community manager, I want to be able to edit sections of text that appear on the site such as the QuickTips in the activity interface sidebar in HTML format

**Is your feature request related to a problem? Please describe.**

**Additional context**

**Acceptance Criteria**

**Acceptance Test:**

rstorey:
Sprint 24

acdha:
> As a community manager, I want to be able to easily edit text and formatting on static pages (specifically wrapping text)

As written, this is already satisfied. Was there some additional changes requested to the way the HTML wrapping occurs or the Markdown formatting rules are applied?

acdha:
> As a community manager, I want to be able to provide links to content within the campaign description (and project description) and add breaks

Line-break support was added in #474 (ecb4612b84300c318c9ff55f17cf48dce679f717), which also discussed other formatting. This should probably be a separate ticket since it will involve [the other questions in that issue](https://github.com/LibraryOfCongress/concordia/issues/474#issuecomment-442467717) about whether this means converting the field to HTML or using Markdown, and what formatting operations are supported.

rstorey:
> **User story/persona**
> * [ ]  As a community manager, I want to display a system message in a banner on (some? all?) pages

Attention @elainekamlley : we'll need @jbresner to draft something for how the system message would look. And probably @klee72 to work on the front end code for that.

rstorey:
> **User story/persona**
> * [ ]  As a community manager, I want to be able to edit sections of text that appear on the site such as the QuickTips in the activity interface sidebar

@elainekamlley : what format should be used for this editing these types of content blocks?  HTML? Plain text? Markdown?

PR #1060, opened 2019-06-13
Simple content blocks

rstorey:
This addresses the last item in #901 

Content blocks must be predefined in the code with a label. That label is then used to populate the content block in the database through the Django admin. A feature flag is used to globally enable or disable the usage of these content blocks.

It's expected that this feature will be used for the Quick Tips section in the classic UI and the Help side panel in the activity UI.

This PR is built on the admin-enhancements branch.


concordia/templates/transcriptions/asset_detail.html
```diff
@@ -306,14 +308,19 @@ <h2>Tags</h2>
...
-                <ul>
-                    <li>Transcribe original spelling, punctuation, word order, and any page numbers or catalog marks.</li>
-                    <li>Preserve line breaks except when a word breaks over a line or page. Then transcribe it on the line or page where it starts.</li>
-                    <li>Use brackets [ ] around deleted, illegible or partially legible text, and square brackets and asterisks around text in margins [ * * ].</li>
-                    <li>Transcribe any words or letters you can't identify as [?].</li>
-                    <li>Include insertions where you would read them in the text.</li>
-                </ul>
-                Find more detailed instructions in the <a href="{% url 'help-center' %}">Help Center</a>
+                {% if SIMPLE_CONTENT_BLOCKS %}
+                    {% simple_content_block "classic_quicktips" as content_block%}
```
acdha:
```suggestion
                    {% simple_content_block "classic_quicktips" as content_block %}
```

acdha:
```suggestion
                    {% simple_content_block "classic_quicktips" as content_block %}
```

acdha: Replacing the existing behaviour will also require rendering embedded Django template tags such as the `{% url %}` link to the help center.

I wrote a [`render_inline`](https://django-bittersweet.readthedocs.io/en/latest/quickstart.html#tag-render-inline) tag for this awhile back in `django-bittersweet` which we've already installed so we should be able to either change this to something like:

```
{% render_inline %}{% simple_content_block "classic_quicktips" as content_block %}{% end_render_inline %}
```

… but that's ungainly and it might be preferable to change the `simple_content_block` tag to either extend that or perform the same operation.

concordia/templatetags/simple_content_block.py
```diff
@@ -0,0 +1,11 @@
+from django import template
+
+from ..models import SimpleContentBlock
+
+register = template.Library()
+
+
+@register.simple_tag()
+def simple_content_block(block_label):
+    content_block = SimpleContentBlock.objects.get(label=block_label)
```
acdha: Currently this will cause a 500 error if a content block with the requested label doesn't exist. Would it be better to trap that and return an empty string?

rstorey: Done

rstorey:
> I'd probably write this using `SimpleContentBlock.DoesNotExist` since that avoids the extra import and has a benefit in more complicated code of not catching exceptions from an unexpected object but for something like this it's really not worth spending more time on.

That answers a question I had in my head after reading the Django docs. Thanks.

PR #1070, opened 2019-06-26
ElasticSearch indices

rstorey:
* Adds transcription count for User
* Adds submission count for Asset

### DIFF
--- a/concordia/documents.py
+++ b/concordia/documents.py
@@ -1,15 +1,11 @@
+from django.contrib.auth.models import User
+from django.db.models import Count
 from django_elasticsearch_dsl import DocType, Index, fields
 
-from .models import (
-    Asset,
-    SiteReport,
-    Transcription,
-    UserAssetTagCollection,
-    UserProfile,
-)
+from .models import Asset, SiteReport, Transcription, UserAssetTagCollection
 
-user_profile = Index("user_profiles")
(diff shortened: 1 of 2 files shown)


concordia/documents.py
```diff
@@ -21,6 +22,12 @@
 
 @user.doc_type
 class UserDocument(DocType):
+    transcription_count = fields.IntegerField()
+
+    def prepare_transcription_count(self, instance):
+        qs = User.objects.filter(id=instance.id).annotate(Count("transcription"))
+        return qs[0].transcription__count
```
acdha: If we're only ever getting a single record, should this be 

```python
user = User.objects.annotate(Count("transcription")).get()
return user.transcription__count
```

concordia/documents.py
```diff
@@ -176,6 +183,13 @@ class AssetDocument(DocType):
         }
     )
 
+    submission_count = fields.IntegerField()
+
+    def prepare_submission_count(self, instance):
+        return Transcription.objects.filter(
+            asset=instance, submitted__isnull=True
+        ).count()
```
acdha: Do we need to filter this on uniqueness in case there's a submit->request changes cycle?

concordia/documents.py
rstorey: I will fix it. Thx

Issue #900, opened 2019-04-18
Kibana / reporting punch list


rstorey:
- [x] Make sure we're on the latest version
- [x] Load the sample data
- [x] Create a new index for assets
- [x] Activity day by day for number of distinct assets accepted, rejected, saved, submitted (asset activity count by status)
- [x] How many times is an asset submitted (re-submitted) in a campaign?
- [x] ~~Add links to Kibana and Adobe analytics from Django admin~~ #1069 
- [x] Retroactively load site report data going back for whole life of site - data is here: https://staff.loc.gov/wikis/pages/viewpage.action?pageId=41126084
- [x] User activity - number of transcriptions
- [ ] Fully automate create the charts that @lalgee is doing semi-manually
- [x] ~~Check confluence for other reporting questions that should be answered by ES/Kibana~~
- [x] Move status filter for assets to the top of the RH column of filters (in Django admin)
- [x] Any open issues with reporting label
- [x] ~~Add last-updated-transcription timestamp to assets~~ See #1029
- [x] Set up daily or real-time index updates
- [x] ~~Determine if there's any way to put text box on visualizations~~ @lalgee I didn't find any reference to this capability for Visualize in the docs. The Canvas functionality might provide what you're looking for. https://www.elastic.co/guide/en/kibana/current/canvas.html

PR #1072, opened 2019-06-27
Management command to import site report data

rstorey:
This provides a management command to import site report data that was collected manually before it was done automatically every day. This way when the site report is exported, there's no need to combine it with a second data source to get all the known data for the life of the site.

This is kind of a mess because in order to set the `created_on` attribute of `SiteReport`, `auto_now_add` has to be set to `False` when importing the data. Then it should be turned back on to `True` (and migration created / applied). Suggestions on that problem welcome.

Site report data to import (cleaned up version) is here: https://staff.loc.gov/wikis/pages/viewpageattachments.action?pageId=41126084&sortBy=date&highlight=site_reports.csv&

Related to #900


rstorey:
I already imported this data in prod and applied the migrations without doing a full deployment, so merging.

PR #1086, opened 2019-07-31
Anonymize user accounts

### LINKED ISSUE #712: Allow community managers to delete accounts but preserve its transcription/tags
Community managers need to be able to meet user requests to delete their accounts without losing any transcription or tagging data accumulated by that account.  Currently all of that data is lost if a user is deleted.

At the time of deletion of a user's account in the admin interface, allow community managers to choose to either delete the users' contribution or transfer to anonymous.  The preferred/more common choice is expected to be to keep the user data, so perhaps that is the default that can be unchecked or overridden as necessary.

rstorey:
Closes #712

### DIFF
--- a/concordia/admin/actions.py
+++ b/concordia/admin/actions.py
@@ -1,7 +1,29 @@
+import hashlib
+
 from django.contrib import messages
 from django.utils.timezone import now
 
 from ..models import Asset, Transcription, TranscriptionStatus
+
+
+def anonymize_action(modeladmin, request, queryset):
+    count = queryset.count()
+    for user_account in queryset:
+        username_md5 = hashlib.md5()
+        email_md5 = hashlib.md5()
+        password_md5 = hashlib.md5()
+        username_md5.update(user_account.username.encode())
+        email_md5.update(user_account.email.encode())
+        password_md5.update(user_account.password.encode())
+        user_account.username = username_md5.hexdigest()
+        user_account.email = email_md5.hexdigest()
+        user_account.password = password_md5.hexdigest()
+        user_account.save()
+
+    messages.info(request, f"Anonymized {count} user accounts")
+
+
+anonymize_action.short_description = "Anonymize user accounts"
 
 
 def publish_item_action(modeladmin, request, queryset):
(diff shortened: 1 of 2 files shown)


concordia/admin/actions.py
```diff
@@ -1,9 +1,25 @@
...
 
 from ..models import Asset, Transcription, TranscriptionStatus
 
 
+def anonymize_action(modeladmin, request, queryset):
+    count = queryset.count()
+    for user_account in queryset:
+        user_account.username = uuid.uuid4()
+        user_account.email = ""
+        user_account.password = ""
```
acdha: Can we change this to call [`set_unusable_password()`](https://docs.djangoproject.com/en/2.2/ref/contrib/auth/#django.contrib.auth.models.User.set_unusable_password)?
```suggestion
        user_account.set_unusable_password()
```

concordia/admin/actions.py
```diff
@@ -1,9 +1,25 @@
...
+        user_account.save()
```
acdha: We should also clear `is_active` since we're resetting the email address as well:

```suggestion
        user_account.is_active = False
        user_account.save()
```

concordia/admin/actions.py
```diff
@@ -1,9 +1,25 @@
...
+
+    messages.info(request, f"Anonymized {count} user accounts")
+
+
+anonymize_action.short_description = "Anonymize user accounts"
```
acdha: This might want to be a bit stronger about completely disabling the account as well.

concordia/admin/actions.py
```diff
@@ -1,9 +1,25 @@
...
 from django.contrib import messages
 from django.utils.timezone import now
 
 from ..models import Asset, Transcription, TranscriptionStatus
 
 
+def anonymize_action(modeladmin, request, queryset):
+    count = queryset.count()
+    for user_account in queryset:
+        user_account.username = uuid.uuid4()
```
acdha: Should we prefix this for easy filtering?

```suggestion
        user_account.username = "Anonymized %s" % uuid.uuid4()
```

PR #1087, opened 2019-07-31
Add topic to site reports

### LINKED ISSUE #1079: Add topic filtering to site reports in Django Admin
**User story/persona**
As a community manager I want to filter by by topic in assets view and in site reports

**Acceptance Criteria**
- Select multiple projects in asset view
- Add to Kibana reporting

rstorey:
Closes #1079


concordia/models.py
```diff
@@ -396,6 +396,7 @@ class SiteReport(models.Model):
     campaign = models.ForeignKey(
         Campaign, on_delete=models.DO_NOTHING, blank=True, null=True
     )
+    topic = models.ForeignKey(Topic, on_delete=models.DO_NOTHING, blank=True, null=True)
```
acdha: Do we want `DO_NOTHING` or `SET_NULL`? The former would appear to prevent a topic from ever being deleted without manually updating the database first.

concordia/tasks.py
```diff
@@ -89,9 +90,84 @@ def site_report():
...
+    assets_total = Asset.objects.filter(item__project__topics=topic).count()
+    assets_published = (
+        Asset.objects.published().filter(item__project__topics=topic).count()
+    )
+    assets_unpublished = (
+        Asset.objects.unpublished().filter(item__project__topics=topic).count()
+    )
+
+    items_published = Item.objects.published().filter(project__topics=topic).count()
+    items_unpublished = Item.objects.published().filter(project__topics=topic).count()
```
acdha: The `Item.objects.published()` start here looks like it should be `unpublished`?

```suggestion
    items_unpublished = Item.objects.unpublished().filter(project__topics=topic).count()
```

concordia/tasks.py
```diff
@@ -89,9 +90,84 @@ def site_report():
...
+        asset__item__project__topics=topic
+    )
+
+    stats = asset_tag_collections.order_by().aggregate(tag_count=Count("tags"))
+    tag_count = stats["tag_count"]
+
+    distinct_tag_list = set()
+
+    for tag_collection in asset_tag_collections:
+        distinct_tag_list.add(tag_collection.tags.values_list("pk", flat=True))
```
acdha: If I'm reading this correctly, it should be calling `set.update` to add each `pk` value to the set rather than each unique group of pk values:

```suggestion
        distinct_tag_list.update(tag_collection.tags.values_list("pk", flat=True))
```

If I'm not rushing this review too much, we'd also want to update https://github.com/LibraryOfCongress/concordia/blob/de024d80f015a65cbdc3e66eb465968430a350a1/concordia/tasks.py#L148

concordia/tasks.py
```diff
@@ -89,9 +90,84 @@ def site_report():
...
+        Asset.objects.filter(item__project__topics=topic)
+        .values_list("transcription_status")
+        .annotate(Count("transcription_status"))
+    )
+
+    for status, count in asset_count_qs:
+        logger.debug("Topic %s assets %s: %d", topic.slug, status, count)
+        report[f"assets_{status}"] = count
+
+    assets_total = Asset.objects.filter(item__project__topics=topic).count()
```
acdha: There are a fair number of queries going on here — if performance is acceptable we can table it but if not we might want to have these count calls count the distinct values for the published field & sum the results so we only walk the assets table once.

rstorey: Ticketed in #1107

concordia/tasks.py
rstorey: I think they both do the same thing, but `.update` is clearer.

rstorey:
Any thoughts how to do unit tests for the site report task without copying and pasting any code from `tasks.py` ?

PR #1088, opened 2019-08-01
Multiple project select filter for items and assets in admin

rstorey:
Refs #1079


concordia/admin/__init__.py
```diff
@@ -46,6 +49,25 @@
 )
 
 
+class ProjectListFilter(MultipleChoiceListFilter):
+    title = "Project"
+
+    def lookups(self, request, model_admin):
+        qs = Project.objects.all()
```
acdha: This block would be a little more efficient as something like `choices = Project.objects.values_list("pk", "title")`

PR #1096, opened 2019-08-12
Graceful handling of network errors

acdha:
I'm working on a fix for https://errorlogging.loc.gov/share/issue/3569473f7b46460f91ccbaaa428fedfb/ which seems to only affect users of Chrome smartphones. I have not yet been able to reproduce it in a standard Chrome build.

### DIFF
--- a/concordia/static/js/action-app/components.js
+++ b/concordia/static/js/action-app/components.js
@@ -11,7 +11,7 @@ import {
     setChildren,
     setAttr,
     List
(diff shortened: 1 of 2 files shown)


acdha:
@rstorey I'm a little iffy about this because it appears to mostly affect Chrome on Chrome OS or smarthphones and I haven't been able to find a way to reproduce it. This commit changes the code to at least handle that situation more gracefully than it does now.

PR #1098, opened 2019-08-14
Jenkins cloudformation

rstorey:
Had to provision a new Jenkins server and took the opportunity to check in more of the configuration in case it needs to happen again.


Dockerfile
```diff
@@ -16,7 +16,7 @@ RUN apt-get update -qy && apt-get dist-upgrade -qy && apt-get install -o Dpkg::O
     libtiff-dev libjpeg-dev libopenjp2-7-dev libwebp-dev zlib1g-dev \
     graphviz \
     locales \
-    nodejs && apt-get -qy -t stretch-backports install npm && apt-get -qy autoremove && apt-get -qy autoclean
+    nodejs node-gyp && apt-get -qy -t stretch-backports install npm && apt-get -qy autoremove && apt-get -qy autoclean
```
acdha: The `node-gyp` package might not be installable from NPM just due to upstream breakage, which is common enough that I remember it by name. Installing it from APT should be a big stability improvement.

cloudformation/infrastructure/jenkins-server.yaml
```diff
@@ -44,10 +44,9 @@ Resources:
                       software-properties-common \
                       docker-ce docker-ce-cli containerd.io \
                       openjdk-8-jdk jenkins \
-                      nginx
+                      nginx awscli
                     usermod -aG docker jenkins
                     snap install postgresql96
-                    pip3 install awscli --upgrade
```
acdha: This is an interesting judgement call: awscli releases frequently enough that I tend to pip install it to get access to new features but if we're not using it having that be cached should be more robust.

PR #1123, opened 2019-09-27
Cloudformation cleanup

### LINKED ISSUE #1101: Fix CloudFormation problems reported by security linter

`docker run -it --rm -v (realpath cloudformation):/templates --entrypoint cfn_nag_scan stelligent/cfn_nag --input-path /templates/ `

rstorey:
Fixes #1101 

Still to do:

- [x] Test IAM role restrictions in dev, test, stage:
    - [x] import
    - [x] export
    - [x] send email
    - [x] get secrets
- [x] Apply same IAM role restrictions to resources for the prod IAM role
- [x] Test all cloudformation changes by setting up a new cluster

To test:
```
docker run -it --rm -v `pwd`/cloudformation:/templates --entrypoint cfn_nag_scan stelligent/cfn_nag --input-path /templates/
```
 should return 0 warnings and 0 failures.

### DIFF
--- a/cloudformation/infrastructure/fargate-cluster.yaml
+++ b/cloudformation/infrastructure/fargate-cluster.yaml
@@ -1,392 +1,390 @@
 Description: >
-    This template deploys a fargate cluster to the provided VPC and subnets
+  This template deploys a fargate cluster to the provided VPC and subnets
 
 Parameters:
-    EnvironmentName:
-        Description: An environment name that will be prefixed to resource names
-        Type: String
+  EnvironmentName:
+    Description: An environment name that will be prefixed to resource names
+    Type: String
 
-    PublicSubnets:
-        Description: The subnets for the load balancer
-        Type: List<AWS::EC2::Subnet::Id>
+  PublicSubnets:
+    Description: The subnets for the load balancer
+    Type: List<AWS::EC2::Subnet::Id>
 
-    PrivateSubnets:
-        Description: Choose which subnets this ECS cluster should be deployed to
-        Type: List<AWS::EC2::Subnet::Id>
+  PrivateSubnets:
+    Description: Choose which subnets this ECS cluster should be deployed to
+    Type: List<AWS::EC2::Subnet::Id>
 
-    SecurityGroup:
-        Description: Select the Security Group to use for the ECS cluster hosts
-        Type: AWS::EC2::SecurityGroup::Id
+  SecurityGroup:
+    Description: Select the Security Group to use for the ECS cluster hosts
+    Type: AWS::EC2::SecurityGroup::Id
 
-    LoadBalancerSecurityGroup:
-        Description: The SecurityGroup for load balancer
-        Type: AWS::EC2::SecurityGroup::Id
+  LoadBalancerSecurityGroup:
+    Description: The SecurityGroup for load balancer
+    Type: AWS::EC2::SecurityGroup::Id
 
-    VpcId:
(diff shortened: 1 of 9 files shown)


cloudformation/infrastructure/fargate-cluster.yaml
```diff
@@ -1,392 +1,407 @@
...
+  VpcId:
+    Description: The Id of the VPC for this cluster
+    Type: AWS::EC2::VPC::Id
+
+  ConcordiaVersion:
+    Type: String
+    Description: version of concordia docker images to pull and deploy
+    Default: latest
+
+  DjangoKeyId:
```
acdha: I'm assuming this is intended to avoid issues where CloudFormation locks resources for long periods of time? That made me wonder whether the should not be a default and whether we could reuse one ID for both the Django and DB secrets.

cloudformation/infrastructure/fargate-cluster.yaml
```diff
@@ -1,392 +1,407 @@
...
-                          - 'kms:Decrypt'
-                          - 'kms:GenerateDataKey'
-                          - 's3:PutObject'
-                          - 's3:GetObject'
-                          - 's3:AbortMultipartUpload'
-                          - 's3:ListMultipartUploadParts'
-                          - 's3:ListBucket'
-                          - 's3:ListBucketMultipartUploads'
-                      Resource:
-                          - '*'
```
acdha: Since this is getting tightened — which is great! — why not break it into multiple statements? The S3 rules seem to be otherwise unrelated and that would allow the linter suppression to only apply to that particular block — especially since it would already need to have both `- !Sub "arn:aws:s3:::crowd-${EnvironmentName}-content/*"` and an additional `- !Sub "arn:aws:s3:::crowd-${EnvironmentName}-content/"` to have `s3:ListBucket` work.

rstorey: Good idea. I'll do that. I actually already did it that in one of the environments because it makes testing easier. The AWS IAM user interface leaves much to be desired.

cloudformation/infrastructure/fargate-cluster.yaml
rstorey: The problem is that the ARNs are generated with random five character suffix when the secrets are created using https://github.com/LibraryOfCongress/concordia/blob/master/cloudformation/create_secrets.sh - the name of the secret is specified, but the ARN is made up by AWS. I need to modify the master template so we can plug in those values at stack creation time.

rstorey:
Currently getting
```
Status reason | CannotPullContainerError: Error response from daemon: Get https://619333082511.dkr.ecr.us-east-1.amazonaws.com/v2/: net/http: request canceled while waiting for connection (Client.Timeout exceeded while awaiting headers)
```
when trying to start up ECS tasks. Seems the NAT gateway for outbound access isn't working, or something.

acdha:
That's definitely what I'd check – I'm assuming it's either the NAT setup or possibly an outbound security group rule. It used to require general internet access — i.e. HTTPS to 0.0.0.0/0 — but https://github.com/aws/containers-roadmap/issues/98 was updated to note that they have added PrivateLink to avoid that: https://aws.amazon.com/about-aws/whats-new/2019/01/aws-fargate--amazon-ecs--and-amazon-ecr-now-have-support-for-aws/

rstorey:
Test stack finally works.

PR #1126, opened 2019-10-16
Remove unnecessary hard memory limits from task containers

rstorey:
Containers in a task used shared memory, the amount of which is specified at the task level. The hard memory limit per container is unnecessary.


rstorey:
In prod the app container is in a separate task already. I need to document that in cloudformation.

PR #1130, opened 2019-10-25
Restore single project ID filtering for AssetAdmin

acdha:
This relied on the permissions added by the default admin list filter interface. That changed from a single value to multiple, breaking the links we had on the item admin page.


acdha:
While we're working on this, I was wondering whether we might want to allow do something like filter the project list filters by the active campaign filter.

PR #1134, opened 2019-10-29
Registration form needs to display error messages

### LINKED ISSUE #1127: Registration page should display error message if user tries to register a duplicate account
The register page refreshed. The password and confirm password fields emptied.

1. Go to the Register page
2. Fill out the form.
3. Go back to a blank Register page.
4. Fill out the form using the same username or email address as in the first try

If a user attempts to register either a username or email address that is already in use by another account, an error message should be displayed and the user should be prompted to use a different username / email address. Perhaps there should also be a link to the reset password page.

-   OS: MacOS
-   Browser Chrome

rstorey:
Fixes #1127 in a very ugly way.

### DIFF
--- a/concordia/static/scss/base.scss
+++ b/concordia/static/scss/base.scss
@@ -1217,6 +1217,14 @@ div.related-links {
     max-width: 30rem;
 }
 
+/* This is here to fix
+    https://github.com/LibraryOfCongress/concordia/issues/1127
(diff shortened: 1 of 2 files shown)


rstorey:
I didn't intentionally change npm-shrinkwrap.json - those changes are an artifact of performing `npm install`.

PR #1144, opened 2019-11-01
WIP: Use staticfiles for front-end dependencies instead of CDNJS

acdha:
* [x] Install django-npm
* [x] Replace cdnjs references with {% static %} calls
* [ ] Tune include list to only have the dist files we care about

I'm not entirely in love with this approach but it avoids the support burden of dealing with Webpack. [`@pika/web`](https://github.com/pikapkg/web) is intriguing but not everything we are using supports ESM.

### DIFF
--- a/concordia/settings_template.py
+++ b/concordia/settings_template.py
@@ -36,10 +36,43 @@ LOGOUT_REDIRECT_URL = "/"
 ROOT_URLCONF = "concordia.urls"
 STATIC_ROOT = "static-files"
 STATIC_URL = "/static/"
+
+STATICFILES_FINDERS = [
+    # We let the filesystem override the app directories so Gulp can pre-process
+    # files if needed:
+    "django.contrib.staticfiles.finders.FileSystemFinder",
+    "django.contrib.staticfiles.finders.AppDirectoriesFinder",
+    # See https://github.com/kevin1024/django-npm
+    "npm.finders.NpmFinder",
+]
+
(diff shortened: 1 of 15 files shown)


concordia/settings_template.py
```diff
@@ -36,10 +36,43 @@
...
+    "openseadragon": ["build/*"],
+    "codemirror": ["lib/*", "addon/*", "mode/*"],
+    "prettier": ["*.js"],
+    "remarkable": ["dist/*"],
+    "jquery": ["dist/*"],
+    "js-cookie": ["src/*"],
+    "popper.js": ["dist/*"],
+    "bootstrap": ["dist/*"],
+    "screenfull": ["dist/*"],
+    "@fortawesome/fontawesome-free/": [
```
rstorey: I think you meant fontawesome here, not fortawesome

package.json
```diff
@@ -16,10 +16,19 @@
         "stylelint-config-recommended": "^3.0.0"
     },
     "dependencies": {
+        "@fortawesome/fontawesome-free": "^5.8.2",
```
rstorey: Fortawesome or fontawesome?

acdha: FortAwesome is the organization: https://fontawesome.com/how-to-use/on-the-web/setup/using-package-managers#installing-free

rstorey: Oh I was not aware of that.

Issue #1146, opened 2019-11-04
Modify BagIt export to properly export item-level transcriptions


rstorey:
**User story/persona**
As a community  manager, I want to export the Clara Barton completed items with item-level transcription text.

**Is your feature request related to a problem? Please describe.**
It was discovered that the item-level transcriptions text for previously exported / P1-available Clara Barton items had added files at the reel level in order to attach item-level text. It would be better if the export straight from concordia had the item-level text instead of reel-level text so that no additional changes are needed to the bag after export.

rstorey:
CC: @VVH

rstorey:
The existing code uses the tile.loc.gov download URL and assumes that the next level directory is the item level. When in fact, at least for the Clara Barton papers, it's the reel level and item level is one level even higher. So we need an alternate way to export item-level transcription text, one that's based on the P1-level definition of what is an item.

acdha:
Does the stored P1 object JSON have any element with the item-level path?

rstorey:
> Does the stored P1 object JSON have any element with the item-level path?

No :( I think the only way to determine the item-level path is to take the whole list of asset download URLs for all the assets in an item and determine the root / what they all have in common.

PR #1148, opened 2019-11-06
Include list of item resource URLs in bagIt export

rstorey:
Refs #1146

### DIFF
--- a/exporter/views.py
+++ b/exporter/views.py
@@ -70,6 +70,31 @@ def get_original_asset_id(download_url):
         return download_url
 
 
+def write_item_resource_file(assets, export_base_dir):
+    item_resource_file = os.path.join(export_base_dir, "item-resource-urls.txt")
+
+    with open(item_resource_file, "a") as f:
+
+        # Find the URL for the item that starts with http://www.loc.gov/resource/
+        items = Item.objects.filter(asset__in=assets)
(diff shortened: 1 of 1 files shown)


exporter/views.py
```diff
@@ -80,19 +80,21 @@ def write_item_resource_file(assets, export_base_dir):
...
-                    "Could not determine item resource URL for item %s", item.item_id
-                )
-                raise AssertionError
+            if item.metadata:
+                for item_url in item.metadata["item"]["aka"]:
+                    if "http://www.loc.gov/resource/" in item_url:
+                        item_resource_url = item_url
+                        break
+
+                if item_resource_url != "":
```
acdha: Couldn't this be `if item_resource_url:`?

exporter/views.py
```diff
@@ -80,19 +80,21 @@ def write_item_resource_file(assets, export_base_dir):
...
+                    f.write(item_resource_url)
+                    f.write("\n")
+                else:
+                    logger.error(
+                        "Could not determine item resource URL for item %s",
+                        item.item_id,
+                    )
+                    raise AssertionError
```
acdha: I'd probably vote for `RuntimeError` or `ValueError` since `AssertionError` implies a more specific context

PR #1149, opened 2019-11-07
WIP: Slim docker builds

acdha:
The container image was ~1.6GB and this avoids needing to deal with CVEs on things which we don't directly depend on, such as ImageMagick.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,1 +1,2 @@
 node_modules
+static-files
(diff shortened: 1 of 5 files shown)


rstorey:
I checked in ECR and the images there (built by jenkins) are about 600MB. Using the slim base image brings it down to about 400MB. The celerybeat and importer Dockerfiles need updating too - I'm working on this in #1152

Issue #1084, opened 2019-07-30
Research and diagnose "Unable to save" issue in prototype


elainekamlley:
**What behavior did you observe? Please describe the bug**
We've seen/heard feedback about this. 

1 - during the Congressional Scavenger hunt event. Roughly about 10 people at any given time were using the prototype during a two-hour span. Many were getting "Unable to save your work." 

2 - A user also emailed about it. Here is the message below: 

> Carefully, I thought about how to proceed, remembering your advice.
>
> At first, I did not click "edit", I went directly to the interface link.  When I could not interact with the page I returned to page 55.  I made no changes to the page, just clicked "edit", then followed the link to the new interface.  When I clicked "edit" on this page, this message appeared:
> 
> "Unable to save your work This transcription has already been reviewed."
> 
>  Now, not only am I unable to continue on this interface page, when I closed this interface page and returned to page 55, I am UNABLE TO OPEN THIS PAGE NOW, the "edit" button does nothing.
> 
> There is definitely a hitch in my maneuvering between these two.   I shall abandon your new interface for now.  I have no problems with your site otherwise and thoroughly enjoy the work I am doing.  I have been honestly moved by the stories I am reading.  And I enjoy being able to still use all my book publishing skills.

A few users will be typing full pages and then come to this message. This reads to me that they are all trying to work on the same asset and the indicators that help them know that are not clear.

acdha:
Do we think this is mostly about providing more prominent explanations for why editing is disabled or is there a confirmed case where it was not disabled but they were unable to save their work? Your first example sounds like the former, which could reflect a problem with the network connection dropping & not receiving the message that someone else has reserved a particular asset.

elainekamlley:
Maybe it was a network issue for the first one. Not quite sure. We can do further testing of this in a coffee shop nearby to try and do this not on LOC's network. Having more than 10 people jumping on the same network will be a regular occurrence here and elsewhere. We are being asked more and more to do transcribing cafes by visitor services, so any suggestions as to how to understand the problem at LOC and potential ways to mitigate this are welcomed. It could be that the prototype is not the right interface for these events and we just use the "find a new page" button to avoid reservation issues. 

For the other example. Here is the page link below: 

https://crowd.loc.gov/campaigns/civil-war-soldiers-disabled-but-not-disheartened/left-hand-penmanship-contest/mss13375024/mss13375024-55/

[...]

Issue #1004, opened 2019-05-22
Messages and edit behavior for review activity in new UI


rstorey:
Stems from #951 
For review mode:
- [ ] After clicking Edit: After clicking Edit, the user should get a confirm message that reads, "Thank you for helping to review this transcription! Submit when you've finished your edits." It should be possible to Save and Submit as if you're in Transcribe mode from here. If the user Submits the transcription to be (re-)reviewed, the confirm message should be "Thanks! Want to keep reviewing?" - and then provide a button that when clicked takes them to the next page to review.
- [ ] After clicking Accept: the message at the bottom should be "Thanks! Want to keep reviewing?" and then provide a button that when clicked takes them to the next page to review (button text TBD)

acdha:
For this, does “message” mean “text which displays where the buttons you just clicked used to be”, “a modal dialog like we currently have”, or something else? I'm leaning towards the former since I feel like part of the spirit of this is not blocking the ability to do things like click on something else in the UI as a modal would.

PR #1152, opened 2019-11-13
Consolidate pull requests

rstorey:
- closes #1133 
- closes #1096
- closes #1065 
- closes #1035 
- closes #1144
- closes #1149
- closes #1017 

This is going to be deployed in a new cluster to test all the changes. There may be to-do items in the above reference PRs that need to be separately ticketed.

This branch will also be the foundation for the work I'm doing in #1004 and #1084 for Sprint 26


.pre-commit-config.yaml
```diff
@@ -43,15 +43,15 @@ repos:
...
-    - repo: https://github.com/pre-commit/mirrors-eslint
-      rev: v6.6.0
-      hooks:
-          - id: eslint
-            additional_dependencies:
-                - eslint@6.4.0
-                - eslint-plugin-prettier@3.1.1
-                - prettier@1.18.2
-                - eslint-plugin-unicorn@11.0.2
+#    - repo: https://github.com/pre-commit/mirrors-eslint
```
acdha: I'd generally just run the merge / commit with `--no-verify` if there's just a one-time need to ignore errors until we revisit some code.

rstorey: I'm having a problem with the certificate on eslint. All the other hooks work okay. I don't want to introduce a bunch of ugly crap while I figure out what's wrong with my pre-commit environment / mirrors-eslint.

celerybeat/Dockerfile
```diff
@@ -8,14 +8,16 @@ ENV DEBIAN_FRONTEND="noninteractive"
...
     libmemcached-dev \
+    # Pillow/Imaging: https://pillow.readthedocs.io/en/latest/installation.html#external-libraries
+    libz-dev libfreetype6-dev \
     libtiff-dev libjpeg-dev libopenjp2-7-dev libwebp-dev zlib1g-dev \
-    graphviz \
-    locales && apt-get -qy autoremove && apt-get -qy autoclean
+    # Postgres client library to build psycopg2
+    libpq-dev \
+    locales \
+    gcc && apt-get -qy autoremove && apt-get -qy autoclean
```
acdha: I don't know that I want to do this immediately but it feels like we should at some point switch to the builder pattern so we can ship only the runtime dependencies

PR #1162, opened 2019-11-22
Restore eslint check to pre-commit for master branch

rstorey:
Finally fixed a problem with my pre-commit environment. I had too many versions of python installed and accidentally had a leftover DYLD_LIBRARY_PATH setting in my .zshrc that shouldn't have been there. I also needed to install the certificates for python.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,75 +1,75 @@
-exclude: '.*/vendor/.*'
+exclude: ".*/vendor/.*"
 repos:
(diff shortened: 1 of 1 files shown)


rstorey:
I think that's because I had `yaml.format.enable` turned on in VS Code, which uses some kind of built in formatter. We don't have prettier running on yaml files in our pre-commit config. Should we change that?

acdha:
That’d explain it - I’d vote yes

PR #1165, opened 2019-11-27
Admin warning banner

### LINKED ISSUE #1156: As an admin, I need to accept the terms upon reading the warning banner so that I can acknowledge this system
Having the language on the admin page does not meet the ISSO requirements. Need to have a page after admin log in that shows the warning language and needs the admin user to click "Okay" to acknowledge that this is a library/US gov system. 

Acceptance criteria: 
- upon a successful admin log in, there is a page that renders

This page needs to include 
- the warning banner language:
> This computer system is the property and/or operated on behalf of the Library of Congress and may be accessed only by authorized users. Users may access and use the Library's computer system only for official business and in accordance with Library regulations. Any usage of the Library's computer system is subject to monitoring by the Library and inappropriate usage may subject the user to loss or limitation of system access rights, adverse administrative action, and criminal prosecution. By accessing and using the Library's computer system, users are consenting to monitoring of their activities and communications on the system.
- An accept button with "I Accept" as the button label
- After click of "I Accept" the admin dash appears

rstorey:
- [x] Move warning banner from admin site dashbaord to admin login template
- [x] Only allow admin access through the admin login form, and not through the regular login form

Closes #1156

### DIFF
--- a/concordia/views.py
+++ b/concordia/views.py
@@ -277,7 +277,7 @@ class ConcordiaLoginView(RatelimitMixin, LoginView):
 
         if user.is_staff:
             logger.info("Staff tried to log in using the regular form, redirecting")
-            return redirect("/admin/login")
+            return redirect(reverse("admin:login"))
         else:
             return super().form_valid(form)
 
--- a/concordia/views.py
+++ b/concordia/views.py
@@ -272,7 +272,14 @@ class ConcordiaLoginView(RatelimitMixin, LoginView):
 
     def form_valid(self, form):
         self.request.session["captcha_validation_time"] = time()
-        return super().form_valid(form)
+
+        user = form.get_user()
+
+        if user.is_staff:
+            logger.info("Staff tried to log in using the regular form, redirecting")
+            return redirect("/admin/login")
+        else:
+            return super().form_valid(form)
 
 
 def ratelimit_view(request, exception=None):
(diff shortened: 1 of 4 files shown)


rstorey:
@acdha Do you have any suggestions or tips how to disallow admin login through the "regular" login form? We want them only to be able to log in using the "admin" login form. It doesn't seem like an option that's built in at all to django.contrib.auth.

acdha:
I'm not sure if there's an existing decorator for it but I was thinking that `ConcordiaLoginView` could implement that — I think we could just add that test to the `form_valid` path to test the user after a successful login:

https://ccbv.co.uk/projects/Django/2.2/django.contrib.auth.views/LoginView/#form_valid

Maybe just tear-down their session and redirect to the admin login page?

concordia/views.py
```diff
@@ -272,7 +272,14 @@ def post(self, request, *args, **kwargs):
 
     def form_valid(self, form):
         self.request.session["captcha_validation_time"] = time()
-        return super().form_valid(form)
+
+        user = form.get_user()
+
+        if user.is_staff:
+            logger.info("Staff tried to log in using the regular form, redirecting")
+            return redirect("/admin/login")
```
acdha: I think that should have a trailing slash but would probably use reversing just in case we ever relocate this:

```suggestion
            return redirect(reverse("admin:login"))
```

PR #1174, opened 2019-12-23
WIP: Redownload images task

rstorey:


concordia/admin/views.py
```diff
@@ -10,11 +10,100 @@
...
+                if not download_url.startswith("http"):
+                    messages.warning(
+                        request, f"Skipping unrecognized URL value: {download_url}"
+                    )
+                    continue
+
+                try:
+                    # Use the download_url to look up the related asset.
+                    # Then queue the task to redownload the image file.
+                    assets = Asset.objects.filter(download_url=download_url)
```
acdha: Can this ever match more than one `Asset`? I'm wondering whether this should just be an `Asset.objects.get()` with an exception handler for the `DoesNotExist` path.

rstorey: Yes and it does in our current database. Not sure why.

rstorey:
> Should we have a strong warning that this will replace assets and potentially invalidate any efforts to transcribe them and/or some sort of warning if a transcription has already been started?

That would only occur if the images changed between the original download and the re-download. Which, in this case, they did. :) So that's why there's a report that's displayed afterward with all the asset statuses, the latest transcriptions for the old assets, and if a second tile.loc.gov URL was in the spreadsheet, a link to that asset as well. It's a manual task for the admin to reopen or reject existing incorrect transcriptions and/or copy transcription text into the correct asset.

concordia/admin/views.py
rstorey: If the asset does not exist (in our current situation that's about 40% of the time) the "No matching asset" message is displayed in the messages output.

acdha:
I was just thinking that it might be useful to have a full warning that this will require the admin to look at any existing transcriptions just in case that's not as obvious as we'd like. Not sure how much more time we'd want to spend on it but I was wondering whether it warranted a “This will affect n transcriptions, continue?” prompt.

rstorey:
Yeah actually also the "reopen" action on assets could also use a confirm prompt. I'm going to ticket both of those.

PR #1183, opened 2020-01-29
Allow failures for travis CI syntax change

rstorey:
Travis CI changed its syntax on how to specify which jobs are allowed to fail.


rstorey:
Well shoot that didn't work.

Issue #1150, opened 2019-11-12
Enforce hard expiration to release asset reservations


rstorey:
**User story/persona**
As a user, if I leave a page open on my browser for 24 hours I see a modal pop up that lets me know that my reservation on that page has expired. 

**Is your feature request related to a problem? Please describe.**
Other volunteers are prevented from working on pages where the reservation is triggered but no work is actively happening 

**Acceptance Criteria**
- user can dismiss the message
- user will remain on the page 
- if no else is on the page, the user can reload and start a new reservation cycle

**Acceptance Test:**
- sit on a page for 24 hours on anonymous
- sit on a page for 24 hours as a logged-in user

acdha:
Can we precisely define the problem: should we set a maximum reservation time, be able to clear all of them, etc.? One of the thoughts I had is that we’d need to tombstone things so the automatic relock doesn’t reclaim it within a minute or two.

rstorey:
There should be a maximum reservation time. In this case the asset has been reserved for over six days by anonymous.

acdha:
I reviewed the code: we'll need to tombstone the expired reservations to prevent the open client from automatically obtaining a new one on the next call.

PR #1186, opened 2020-01-29
Tombstone old asset reservations

rstorey:
Refs #1150 
- [x] Add a task that retrieves reservations more than 24 hours old
- [x] Tombstone the old reservations
- [x] Show the user a message in the browser
- [x] Allow the user to refresh the reservation if manual action is taken (idle browser window would not continue the reservation) - this is allowed after a configurable period of hours


concordia/models.py
```diff
@@ -367,6 +367,8 @@ class AssetTranscriptionReservation(models.Model):
 
     created_on = models.DateTimeField(editable=False, auto_now_add=True)
     updated_on = models.DateTimeField(auto_now=True)
+    last_reserve_time = models.DateTimeField(auto_now_add=True, blank=True, null=True)
```
acdha: It's not immediately obvious to me why `last_reserve_time` isn't `updated_on`?

rstorey:
Because updated_on gets a fresh timestamp when something is tombstoned.


________________________________

concordia/models.py
rstorey: I had played around with dirtying the reservation token, which updated the timestamp. Since I added a separate boolean field to track whether the reservation is valid, the extra timestamp isn't needed. I removed it.

rstorey:
I tested it locally, for whatever that's worth. I expect some combination of Community Managers to do functional testing in dev / test after it's merged to master. This needs celery tasks to run, so it can't be fully tested in a feature branch deployment.

PR #1202, opened 2020-03-04
Restore update_task_status decorator to download_asset task

### LINKED ISSUE #1197: Restore update_task_status decorator to download asset task
I removed it when we needed to use the redownload images feature. It should be restored for troubleshooting regular imports. If we need to use the redownload feature again, we should create a data model for the redownload task so the update_task_status decorator can be applied.

rstorey:
Closes #1197


rstorey:
I disabled it in the most crude way possible and put a message on the admin page.

PR #1209, opened 2020-03-16
Add unlisted attribute to campaigns and topics

### LINKED ISSUE #1206: Be able to publish a campaign that's not listed on the campaigns page
In order to enable projects to be worked on by specific sets of people, we want to be able to import and publish campaigns that will not be listed on the home page or the campaigns page.

rstorey:
Closes #1206


concordia/views.py
```diff
@@ -1563,7 +1568,7 @@ def filter_and_order_reviewable_assets(
 @never_cache
 @atomic
 def redirect_to_next_reviewable_asset(request):
-    campaign = Campaign.objects.published().order_by("ordering")[0]
+    campaign = Campaign.objects.published().listed().order_by("ordering")[0]
```
acdha: This reminds me that we might want to handle the case where `[0]` would fail if the queryset was empty but I'm not sure whether it's worth changing it to use`.first()` and checking the result since we haven't had this problem yet.

PR #1210, opened 2020-03-18
Fix for unlisted campaigns using find a new page to transcribe or review

rstorey:


rstorey:
Still open to review on this, but merging so it can be tested in dev.

PR #1214, opened 2020-03-24
Tune channels config

rstorey:
Configure channels with greater capacity than the default and shorter expiry than the default, so channels don't fill up


rstorey:
Should I upgrade redis also?

rstorey:
I'm going to merge this to master so I can load test it in crowd-dev and see what happens. Please feel free to leave your comments and I will incorporate them into this stream of work.

acdha:
This seems reasonable at first but I wonder whether we're going to hit a problem where more channels means more work accepted before a container is torn down.

PR #1234, opened 2020-07-07
regex fix

rstorey:


rstorey:
We need to upgrade Django, separately from this.

Issue #1227, opened 2020-05-28
As a community manager I want to move projects to different campaign


elainekamlley:
Currently when a CM moves projects to a new campaign in the admin dashboard an additional step to move the projects in s3 needs to be made. There needs to be a signal to S3 to move projects/items/assets when those changes are made in the admin dashboard.

rabiloc:
@elainekamlley Hey Elaine, can you please give me a quick overview of how you currently do this? Maybe a meeting next week. Thanks.

PR #1236, opened 2020-07-14
Project move 1227

rabiloc:
Transfer project to campaign #1227

### DIFF
--- a/concordia/admin/__init__.py
+++ b/concordia/admin/__init__.py
@@ -281,8 +281,10 @@ class ProjectAdmin(admin.ModelAdmin, CustomListDisplayFieldsMixin):
                 if res != []:
                     bucket.delete_objects(Delete={"Objects": res})
 
-        except Exception as error_message:
-            logger.error(error_message)
+        except Exception:
+            logger.exception(
+                "Unhandled exception saving from %s to %s", new_campaign, new_project
+            )
             raise
 
     @method_decorator(permission_required("concordia.add_campaign"))
--- a/concordia/admin/__init__.py
+++ b/concordia/admin/__init__.py
@@ -1,5 +1,4 @@
 import logging
-import os
 from urllib.parse import urljoin
 
 import boto3
@@ -212,7 +211,7 @@ class TopicAdmin(admin.ModelAdmin):
 
 @admin.register(Project)
 class ProjectAdmin(admin.ModelAdmin, CustomListDisplayFieldsMixin):
-    AWS_DEFAULT_REGION = os.getenv("AWS_DEFAULT_REGION", "us-east-1")
+
     form = BleachedDescriptionAdminForm
 
     # todo: add foreignKey link for campaign
@@ -248,12 +247,10 @@ class ProjectAdmin(admin.ModelAdmin, CustomListDisplayFieldsMixin):
 
         super().save_model(request, obj, form, change)
         # call boto3 api to move project files from older campaign to newer campaign
-        # call boto3 api to update project name within same campaigns and transfer files
         try:
 
             new_campaign = obj.campaign.slug
             new_project = obj.slug
-            # aws_profile = "AWSCrowdlocgov"
             bucket_name = getattr(settings, "S3_BUCKET_NAME", None)
             session = boto3.session.Session()
             s3_resource = session.resource("s3")
@@ -264,7 +261,7 @@ class ProjectAdmin(admin.ModelAdmin, CustomListDisplayFieldsMixin):
                 for obj in top_level_bucket.objects.filter(
                     Prefix=filter_from + "/"
                 ).all():
-                    # print(obj.key.replace(old_campaign,new_campaign).replace(old_project,new_project))
+
                     s3_resource.Object(
                         bucket_name,
                         obj.key.replace(old_campaign, new_campaign).replace(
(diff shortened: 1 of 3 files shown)


concordia/settings_dev.py
```diff
@@ -45,3 +45,4 @@
     ("concordia.utils", "get_anonymous_user"),
     ("concordia.models", "TranscriptionStatus"),
 ]
+S3_BUCKET_NAME = "crowd-dev-content"
```
rstorey, line 48: Unless you put the other setting(s) such as MEDIA_URL, this should not be here.

concordia/admin/__init__.py
```diff
@@ -233,6 +241,53 @@ def get_urls(self):
...
+
+                res = []
+                bucket = s3_resource.Bucket(bucket_name)
+                for obj_version in bucket.object_versions.filter(
+                    Prefix=filter_from + "/"
+                ).all():
+                    res.append(
+                        {"Key": obj_version.object_key, "VersionId": obj_version.id}
+                    )
+                    # print(res)
```
rstorey, line 283: Either convert this to a logging debug call or remove it

concordia/admin/__init__.py
```diff
@@ -233,6 +241,53 @@ def get_urls(self):
...
+            session = boto3.session.Session()
+            s3_resource = session.resource("s3")
+            top_level_bucket = s3_resource.Bucket(bucket_name)
+            filter_from = old_campaign + "/" + old_project
+            if old_campaign != new_campaign or old_project != new_project:
+
+                for obj in top_level_bucket.objects.filter(
+                    Prefix=filter_from + "/"
+                ).all():
+                    # print(obj.key.replace(old_campaign,new_campaign).replace(old_project,new_project))
```
rstorey, line 267: Either convert this to a logging debug call or remove it

concordia/admin/__init__.py
```diff
@@ -233,6 +241,53 @@ def get_urls(self):
...
+        old_project = Project.objects.get(pk=obj.pk).slug
+
+        super().save_model(request, obj, form, change)
+        # call boto3 api to move project files from older campaign to newer campaign
+        # call boto3 api to update project name within same campaigns and transfer files
+        try:
+
+            new_campaign = obj.campaign.slug
+            new_project = obj.slug
+            # aws_profile = "AWSCrowdlocgov"
```
rstorey, line 256: This comment should be removed

concordia/admin/__init__.py
```diff
@@ -205,6 +212,7 @@ class TopicAdmin(admin.ModelAdmin):
 
 @admin.register(Project)
 class ProjectAdmin(admin.ModelAdmin, CustomListDisplayFieldsMixin):
+    AWS_DEFAULT_REGION = os.getenv("AWS_DEFAULT_REGION", "us-east-1")
```
rstorey, line 215: Is this used anywhere? If it is needed, it should probably be fed into the Django settings from the environment, and then from the Django settings into whatever code needs it.

concordia/admin/__init__.py
```diff
@@ -233,6 +241,53 @@ def get_urls(self):
 
         return custom_urls + urls
 
+    def save_model(self, request, obj, form, change):
+
+        old_campaign = Project.objects.get(pk=obj.pk).campaign.slug
+        old_project = Project.objects.get(pk=obj.pk).slug
+
+        super().save_model(request, obj, form, change)
+        # call boto3 api to move project files from older campaign to newer campaign
```
rstorey, line 250: These comments make it sound as if you're doing these two actions separately when really they're part of the same move process. A bit confusing.

concordia/admin/__init__.py
```diff
@@ -233,6 +241,53 @@ def get_urls(self):
...
+                    res.append(
+                        {"Key": obj_version.object_key, "VersionId": obj_version.id}
+                    )
+                    # print(res)
+
+                if res != []:
+                    bucket.delete_objects(Delete={"Objects": res})
+
+        except Exception as error_message:
+            print(error_message)
```
rstorey, line 289: Exceptions should be properly handled and not printed - at a minimum log it and then raise it up to the calling method

concordia/settings_dev.py
rabiloc: Thanks removed it.

concordia/admin/__init__.py
rabiloc: It was left over, removed it.

concordia/admin/__init__.py
rabiloc: Removed the lower comment.

concordia/admin/__init__.py
rabiloc: put it as an error in the logger and raised the exception.

concordia/admin/__init__.py
acdha: I would be a strong +1 for removing this – this is the kind of thing which should be set globally in the environment to avoid very confusing errors.

concordia/admin/__init__.py
```diff
@@ -280,13 +277,13 @@ def save_model(self, request, obj, form, change):
...
                         {"Key": obj_version.object_key, "VersionId": obj_version.id}
                     )
-                    # print(res)
 
                 if res != []:
                     bucket.delete_objects(Delete={"Objects": res})
 
         except Exception as error_message:
-            print(error_message)
+            logger.error(error_message)
```
acdha, line 285: Two thoughts: any logger call should have a message explaining the context in case the exception doesn't have enough information to be meaningful in the logs:

```suggestion
            logger.exception(f"Unhandled exception saving {obj}")
```

However, since we're not doing anything else before raising it again, should we even bother catching it instead of letting the caller's normal error handling take care of it? Since it's going into Sentry anyway I'm not sure whether this is giving us enough context to be worth the extra code.

rabiloc: updating the code to catch the exception with a better contextual message and then raising it to sentry.

concordia/admin/__init__.py
rabiloc: Yes that was not needed, I removed it from the code.

rabiloc:
Merge Project-Move

PR #1252, opened 2020-08-26
add diacritical marks

rabiloc:
allow diacritical marks when adding tags.


rstorey:
I assume the TAG_VALIDATOR in models.py doesn't need updating?

rstorey:
Seems like adding a unit test for this would be an easy win as well.

rabiloc:
Users tested it, looks good, we can use Unicode, if needed, but good for now.

PR #1257, opened 2020-09-16
Improve performance of Item delete form for large numbers of items

rabiloc:
Display just the items and not the related assets in the summary page.

### DIFF
--- a/concordia/admin/__init__.py
+++ b/concordia/admin/__init__.py
@@ -1,8 +1,9 @@
 import logging
 from urllib.parse import urljoin
+
 from django.conf import settings
-from django.contrib.auth import get_permission_codename
 from django.contrib import admin
+from django.contrib.auth import get_permission_codename
 from django.contrib.auth.admin import UserAdmin
 from django.contrib.auth.decorators import permission_required
 from django.contrib.auth.models import User
@@ -311,20 +312,27 @@ class ItemAdmin(admin.ModelAdmin):
 
     actions = (publish_item_action, unpublish_item_action)
 
-    def get_deleted_objects(self, objs, request):  
- 
+    def get_deleted_objects(self, objs, request):
+
         deleted_objects = [str(obj) for obj in objs]
         perms_needed = set()
         for model in (Item, Asset, Transcription):
-            perm = '%s.%s' % (model._meta.app_label, get_permission_codename('delete', model._meta))
+            perm = "%s.%s" % (
+                model._meta.app_label,
+                get_permission_codename("delete", model._meta),
+            )
             if not request.user.has_perm(perm):
                 perms_needed.add(model._meta.verbose_name)
         protected = []
         model_count = {
             Item._meta.verbose_name_plural: len(objs),
-            Asset._meta.verbose_name_plural: Asset.objects.filter(item__in=objs).count(),
-            Transcription._meta.verbose_name_plural: Transcription.objects.filter(asset__item__in=objs).count(),
-                }
+            Asset._meta.verbose_name_plural: Asset.objects.filter(
+                item__in=objs
+            ).count(),
+            Transcription._meta.verbose_name_plural: Transcription.objects.filter(
+                asset__item__in=objs
+            ).count(),
+        }
         return (deleted_objects, model_count, perms_needed, protected)
 
     def get_queryset(self, request):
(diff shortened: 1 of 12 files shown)


concordia/admin/__init__.py
```diff
@@ -310,6 +310,13 @@ class ItemAdmin(admin.ModelAdmin):
 
     actions = (publish_item_action, unpublish_item_action)
 
+    def get_deleted_objects(self, objs, request):
+        deleted_objects = [str(obj) for obj in objs]
+        model_count = {'Items': len(deleted_objects)}
+        perms_needed = []
```
acdha, line 316: This should probably at the least include the item modify permission requirement

rabiloc: sounds good, let me see what that permission looks like.

rabiloc: ok added set() to perms_needed propery

concordia/admin/__init__.py
```diff
@@ -310,13 +310,13 @@ class ItemAdmin(admin.ModelAdmin):
 
     actions = (publish_item_action, unpublish_item_action)
 
-    def get_deleted_objects(self, objs, request):
+    def get_deleted_objects(self, objs, request):  
         deleted_objects = [str(obj) for obj in objs]
-        model_count = {'Items': len(deleted_objects)}
-        perms_needed = []
+        perms_needed = set()
```
acdha, line 315: This doesn't check that the user has permissions to delete related objects. I think in this case we could either hard-code the list or do something like this:

```python
from django.contrib.auth import get_permission_codename

…

perms_needed = set()
for model in (Item, Asset, Transcription):
    p = '%s.%s' % (model._meta.app_label, get_permission_codename('delete', model._meta))
    if not user.has_perm(p):
        perms_needed.add(model._meta.verbose_name)
```

concordia/admin/__init__.py
```diff
@@ -310,13 +310,13 @@ class ItemAdmin(admin.ModelAdmin):
...
         protected = []
+        model_count = {'Items': len(deleted_objects)}
```
acdha, line 317: This could be efficiently calculated as something like this — note the use of `verbose_name_plural` as per [the ModelAdmin docs](https://docs.djangoproject.com/en/3.1/ref/contrib/admin/#django.contrib.admin.ModelAdmin.get_deleted_objects) — to calculate the totals using one query per model rather than n-per-instance:

```python
model_count = {
    Item._meta.verbose_name: len(objs),
    Asset._meta.verbose_name_plural: Asset.objects.filter(item__in=objs).count(),
    Transcription._meta.verbose_name_plural: Transcription.objects.filter(asset__item__in=objs).count(),
}
```

concordia/admin/__init__.py
```diff
@@ -310,13 +310,13 @@ class ItemAdmin(admin.ModelAdmin):
 
     actions = (publish_item_action, unpublish_item_action)
 
-    def get_deleted_objects(self, objs, request):
+    def get_deleted_objects(self, objs, request):  
         deleted_objects = [str(obj) for obj in objs]
```
acdha, line 314: There are two things to be concerned with here: Item does not use a related lookup in its default string representation but if it did this would trigger O(n) queries unless the `QuerySet` used `select_related` and/or `prefetch_related`. For bulk deletion, there could be some value in a heuristic which did something like using `str(obj.pk)` if the queryset length was over a certain value.

Related, this will generate one list item per object. If you are deleting hundreds or thousands of objects this is not always desirable and it might be preferable to have that heuristic do something like this:

```python
if len(objs) < 30:
    deleted_objects = [str(obj) for obj in objs]
else:
    deleted_objects = [str(obj) for obj in objs[:3], f"… and {len(objs) - 3} more {obj._meta.verbose_name_plural}"]
```

rabiloc: Ok all the changes from above are completed and pushed except the one about the deleted_objects, not sure how to code this part - > deleted_objects = [str(obj) for obj in objs[:3], f"… and {len(objs) - 3} more {obj._meta.verbose_name_plural}"]

concordia/admin/__init__.py
rabiloc: done

concordia/admin/__init__.py
rabiloc: done

PR #1260, opened 2020-09-24
Padding 1259

### LINKED ISSUE #1259: Add padding to Help Center cards
The bottom of the help center cards are flush with the footer

rabiloc:
Based @klee72 suggestion

Closes #1259

### DIFF
--- a/concordia/static/scss/base.scss
+++ b/concordia/static/scss/base.scss
@@ -811,10 +811,9 @@ body.view-transcriptions--asset-detail main {
 
 .help-center-card {
     color: #fff;
-    // width: 300px;
     height: 225px;
     background-color: var(--primary-color);
-    margin-bottom:1rem;
+    margin-bottom: 1rem;
 }
 
 .help-center-card a {
@@ -996,7 +995,6 @@ div.related-links {
     background-color: var(--bg-lightest-gray);
     border-top: 1px solid var(--bg-medium-gray);
     padding: 22px 0;
-
 }
 
 .footer-crowd-row {
--- a/concordia/static/scss/base.scss
+++ b/concordia/static/scss/base.scss
@@ -811,9 +811,10 @@ body.view-transcriptions--asset-detail main {
 
 .help-center-card {
     color: #fff;
-    width: 300px;
+    // width: 300px;
     height: 225px;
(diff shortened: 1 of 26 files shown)


concordia/static/scss/base.scss
```diff
@@ -811,9 +811,10 @@ body.view-transcriptions--asset-detail main {
 
 .help-center-card {
     color: #fff;
-    width: 300px;
+    // width: 300px;
```
acdha, line 814: If this is not needed it should be removed. If there was any need to get it back in the future, Git has a full history.

concordia/static/scss/base.scss
```diff
@@ -811,9 +811,10 @@ body.view-transcriptions--asset-detail main {
...
     height: 225px;
     background-color: var(--primary-color);
+    margin-bottom:1rem;
```
acdha, line 817: Follow the instructions for enabling pre-commit: https://github.com/LibraryOfCongress/concordia/blob/master/docs/how-we-work.md#configure-your-local-checkout-with-code-quality-hooks

This will make sure that every commit runs Prettier even if made using an editor which doesn't have Prettier enabled.

concordia/static/scss/base.scss
```diff
@@ -995,6 +996,7 @@ div.related-links {
     background-color: var(--bg-lightest-gray);
     border-top: 1px solid var(--bg-medium-gray);
     padding: 22px 0;
+
```
acdha, line 999: This will also be removed by Prettier

acdha (changes requested):

acdha:
It looks like this was based on #1253 rather than `master`. It'd be a little cleaner to review if it was just the changes for this issue — whether that's merging #1253 first or using `git rebase -i` to have this be only the change to the stylesheets.

concordia/static/scss/base.scss
rabiloc: ok got pre-commit installed now.

concordia/static/scss/base.scss
rabiloc: ok removed the line, and pushed it back up.

acdha (approved):

PR #1265, opened 2020-09-29
css changes for moving asset status

rabiloc:
css changes.


rstorey (approved):
It's kind of hard to know whether these changes will have the intended effect since the pull request doesn't describe what these changes are intended to do.

PR #1267, opened 2020-10-07
url links

rabiloc:
For multi-volume links, the sp indicator should by pass adding the sp value.

### DIFF
--- a/concordia/templates/transcriptions/asset_detail.html
+++ b/concordia/templates/transcriptions/asset_detail.html
@@ -77,9 +77,15 @@
             {% share_buttons current_asset_url asset.item.title %}
 
             {% if asset.resource_url %}
+                {% if 'sp' in asset.resource_url %}
+                <div class="btn-group btn-group-sm ml-2" role="navigation" aria-label="Link to the original source for this item">
+                    <a class="btn btn-default" target="_blank" title="View the original source for this item in a new tab" href="{{ asset.resource_url }}">View on www.loc.gov <i class="fa fa-external-link-alt"></i></a>
+                </div>
+                {% else %}
                 <div class="btn-group btn-group-sm ml-2" role="navigation" aria-label="Link to the original source for this item">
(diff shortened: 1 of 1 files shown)


rstorey (approved):

concordia/templates/transcriptions/asset_detail.html
```diff
@@ -77,9 +77,15 @@
             {% share_buttons current_asset_url asset.item.title %}
 
             {% if asset.resource_url %}
+                {% if 'sp' in asset.resource_url %}
+                <div class="btn-group btn-group-sm ml-2" role="navigation" aria-label="Link to the original source for this item">
```
acdha, line 81: We've historically indented nested inside a template block for clarity

concordia/templates/transcriptions/asset_detail.html
```diff
@@ -77,9 +77,15 @@
...
+                    <a class="btn btn-default" target="_blank" title="View the original source for this item in a new tab" href="{{ asset.resource_url }}">View on www.loc.gov <i class="fa fa-external-link-alt"></i></a>
```
acdha, line 82: Since this code is otherwise identical except for the link target, I'd nest it closer to the source – leave the identical <div> outside of the second `if` and only put the conditional around the `<a>` tag which is being changed to avoid the possibility of someone forgetting to keep them in sync otherwise when making changes in the future.

Since this is so small a change, I'm wondering whether it be cleaner to do something like this instead to make it more obvious that only one attribute is being modified:

```django
<a class="btn btn-default" target="_blank" title="View the original source for this item in a new tab" href="{{ asset.resource_url }}{% if 'sp=' not in asset.resource_url %}?sp={{ asset.sequence }}{% endif %}">View on www.loc.gov <i class="fa fa-external-link-alt"></i></a>
```

There might also be an argument for adding an accessor property to `Asset` with a more explanatory name (`asset.source_page_url`? maybe @rstorey has a better name) which would allow the template to avoid the `if` test entirely and could perform a more reliable test by using `urllib.parse` to apply standard querystring parsing and avoid the possibility of unexpected results when given a URL which has any other values which end in `sp`, matching `sp=`.

rabiloc: Good suggestions Chris , I will incorporate those and push them in next.

PR #1269, opened 2020-10-08
fix opacity settings

rabiloc:
Fix opacity settings, and move status text to the top and number to the bottom of the assets.


rstorey (approved):
Approving even though the "// add this line" comments should have been removed... clean up next time

Issue #1266, opened 2020-09-29
Add labels for campaign short and long description fields in Admin Campaign editor


lalgee:
**User story/persona**
As an Community Manager, I need to know which Campaign description field I am editing when making changes via the Admin Campaign page.

**Is your feature request related to a problem? Please describe.**
Campaigns have 2 description fields which are editable via text boxes in the Concordia Admin but these fields are not labeled so it is unclear to which field you are editing. The long description field is currently ordered first and the short description field is below.

**Additional context**
As showin in the screenshot below, text box 1 is the campaign long description and box 2 is the short description.
![image](https://user-images.githubusercontent.com/20258744/94602108-70f3ab80-0262-11eb-9693-73bbdf32f86a.png)

**Acceptance Test:**
Both text boxes are labeled to indicate which description field they contain. As an Admin, I can easily see the field names and understand what text I am editing.

jkueloc:
@lalgee this is ready for testing

lalgee:
Looks good!

PR #1281, opened 2020-11-04
Exposing Campaign editor description field labels

jkueloc:
This request is for ticket #1266

### DIFF
--- a/concordia/static/admin/editor-preview.js
+++ b/concordia/static/admin/editor-preview.js
@@ -20,7 +20,7 @@
             .parents('.form-row')
             .first();
         $formRow.addClass('codemirror-with-preview');
-        $formRow.find('label').remove();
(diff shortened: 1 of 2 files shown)


rabiloc (approved):
Looks good

rabiloc:
merged

concordia/static/admin/editor-preview.js
```diff
@@ -20,7 +20,7 @@
             .parents('.form-row')
             .first();
         $formRow.addClass('codemirror-with-preview');
-        $formRow.find('label').remove();
+        //$formRow.find('label').remove();
```
acdha, line 23: If we're changing this we should delete the line so there's no question about intent

jkueloc: I agree. I'll make the change.

PR #1290, opened 2020-12-10
asset counts

rabiloc:
made fixes to allow async calls to the loc.gov api, and a wait time in between calls.

### DIFF
--- a/concordia/admin/views.py
+++ b/concordia/admin/views.py
@@ -1,5 +1,5 @@
-import asyncio
 import re
+import time
 
 from bittersweet.models import validated_get_or_create
 from django.contrib import messages
@@ -12,7 +12,7 @@ from django.views.decorators.cache import never_cache
 from tabular_export.core import export_to_csv_response, flatten_queryset
 
 from importer.tasks import (
-    import_item_count_from_url,
+    fetch_all_urls,
     import_items_into_project_from_url,
     redownload_image_task,
 )
@@ -120,19 +120,35 @@ def redownload_images_view(request):
 @permission_required("concordia.change_project")
 @permission_required("concordia.add_item")
 @permission_required("concordia.change_item")
+def celery_task_review(request):
+    request.current_app = "admin"
+    context = {"title": "Active Celery Tasks"}
+    return render(request, "admin/celery_task.html", context)
+
+
+@never_cache
+@staff_member_required
+@permission_required("concordia.add_campaign")
+@permission_required("concordia.change_campaign")
+@permission_required("concordia.add_project")
+@permission_required("concordia.change_project")
+@permission_required("concordia.add_item")
+@permission_required("concordia.change_item")
 def admin_bulk_import_review(request):
     request.current_app = "admin"
     url_regex = r"[-\w+]+"
     pattern = re.compile(url_regex)
     context = {"title": "Bulk Import Review"}
-    event_loop = asyncio.new_event_loop()
-    asyncio.set_event_loop(event_loop)
-
+    # event_loop = asyncio.new_event_loop()
+    # asyncio.set_event_loop(event_loop)
+    urls = []
+    all_urls = []
+    url_counter = 0
+    sum_count = 0
     if request.method == "POST":
         form = AdminProjectBulkImportForm(request.POST, request.FILES)
 
         if form.is_valid():
-            total_count = 0
             rows = slurp_excel(request.FILES["spreadsheet_file"])
             required_fields = [
                 "Campaign",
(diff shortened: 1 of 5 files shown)


concordia/admin/views.py
```diff
@@ -125,14 +139,16 @@ def admin_bulk_import_review(request):
     url_regex = r"[-\w+]+"
     pattern = re.compile(url_regex)
     context = {"title": "Bulk Import Review"}
-    event_loop = asyncio.new_event_loop()
-    asyncio.set_event_loop(event_loop)
-
+    # event_loop = asyncio.new_event_loop()
```
acdha, line 142: If this is unused, it should be removed instead of commented out

concordia/admin/views.py
```diff
@@ -192,27 +208,30 @@ def admin_bulk_import_review(request):
...
 
                         try:
+                            urls.append(url)
+                            url_counter = url_counter + 1
 
-                            count = event_loop.run_until_complete(
-                                import_item_count_from_url(url)
-                            )
-                            total_count = total_count + count
+                            if url_counter == 30:
```
acdha, line 214: This could be simplified to `url_counter % 30 == 0` or `len(all_urls) % 30 == 1`

importer/tasks.py
```diff
@@ -239,14 +240,33 @@ def get_item_info_from_result(result):
...
-    resp = requests.get(import_url, params={"fo": "json"})
-    resp.raise_for_status()
-    item_data = resp.json()
-    return len(item_data["resources"][0]["files"])
+    try:
+
+        resp = requests.get(import_url, params={"fo": "json"})
+        resp.raise_for_status()
+        item_data = resp.json()
+        output = len(item_data["resources"][0]["files"])
```
acdha, line 265: Do we know that there will only ever be one resource?

importer/tasks.py
```diff
@@ -239,14 +240,33 @@ def get_item_info_from_result(result):
...
+        return f"{import_url} {output}", output
+
+    except Exception as exc:
+        return f"{exc}", 0
```
acdha, line 269: This exception should be logged to Sentry to save time when you're asked to troubleshoot:

```python
logging.warning("Unable to retrieve file count from %s: %s", import_url, exc)
```

importer/tasks.py
```diff
@@ -239,14 +240,33 @@ def get_item_info_from_result(result):
...
-    resp.raise_for_status()
-    item_data = resp.json()
-    return len(item_data["resources"][0]["files"])
+    try:
+
+        resp = requests.get(import_url, params={"fo": "json"})
+        resp.raise_for_status()
+        item_data = resp.json()
+        output = len(item_data["resources"][0]["files"])
+        return f"{import_url} {output}", output
```
acdha, line 266: The return signature for this function seems a bit inconsistent with it either being a string containing the URL and the count (which would need to be parsed), followed by the count or the string value of an exception and the count.

It might make sense to add an explicit success value and clarify that the intention is to return a human-meaningful message. 

You can add a type annotation to have tools like mypy or pylance (in VSCode) validate these calls:

```python
from typing import Tuple

def import_item_count_from_url(import_url: str) -> Tuple[bool, str, int]:
```

concordia/admin/views.py
```diff
@@ -192,27 +208,30 @@ def admin_bulk_import_review(request):
...
-                                f"{project_title} - [{url}] : {count}",
-                            )
                         except Exception as exc:
                             messages.error(
                                 request,
                                 f"Unhandled error attempting to count {url}: {exc}",
                             )
+
+                all_urls.append(urls)
+                for i, val in enumerate(all_urls):
```
acdha, line 226: `i` doesn't appear to being used so this `enumerate` call could be removed

concordia/admin/views.py
```diff
@@ -192,27 +208,30 @@ def admin_bulk_import_review(request):
...
+                    return_result = fetch_all_urls(val)
```
acdha, line 227: Since the return type is not obvious I would unpack this into logical variables:

```python
message, count = fetch_all_urls(val)
```

concordia/admin/views.py
```diff
@@ -192,27 +208,30 @@ def admin_bulk_import_review(request):
...
+                    messages.info(request, f"Asset Count:{return_result[0]}")
+                    sum_count = sum_count + return_result[1]
+                    time.sleep(10)
```
acdha, line 230: This will block the current process until the task returns. If the intention is to prevent overloading the remote server it might be better to add some delay to the actual request function to avoid issuing 15 requests at the same time and then idling for 10 seconds after the last one responds. A simple `time.sleep(random.random())` would avoid a dog-pile, and I especially like using that when retrying errors — note that this is extremely easy with something like [backoff](https://pypi.org/project/backoff/) on the actual retrieval function.

concordia/admin/views.py
```diff
@@ -112,6 +112,20 @@ def redownload_images_view(request):
...
 
+@never_cache
+@staff_member_required
+@permission_required("concordia.add_campaign")
+@permission_required("concordia.change_campaign")
+@permission_required("concordia.add_project")
+@permission_required("concordia.change_project")
+@permission_required("concordia.add_item")
+@permission_required("concordia.change_item")
+def celery_task_review(request):
```
acdha, line 123: This function is named in a way which suggests that it's using Celery but it doesn't appear to do anything other than display the messages pending for the active user. That will also mean that this will be empty most of the time since messages only display once so you would only see it if you went to this page without hitting any other page.

PR #1291, opened 2020-12-10
1261 importer overview

rabiloc:
formatting fixes.

### DIFF
--- a/concordia/admin/views.py
+++ b/concordia/admin/views.py
@@ -225,13 +225,15 @@ def admin_bulk_import_review(request):
                 all_urls.append(urls)
                 for i, val in enumerate(all_urls):
                     return_result = fetch_all_urls(val)
-                    messages.info(request, f"Asset Count:{return_result[0]}")
+                    for res in return_result[0]:
+                        messages.info(request, f"{res}")
+
                     sum_count = sum_count + return_result[1]
                     time.sleep(10)
 
                 messages.info(request, f"Total Asset Count:{sum_count}")
             finally:
-                messages.info(request, "All processes Complete")
+                messages.info(request, "********************************************")
 
     else:
         form = AdminProjectBulkImportForm()
--- a/concordia/admin/views.py
+++ b/concordia/admin/views.py
@@ -233,7 +233,7 @@ def admin_bulk_import_review(request):
 
                 messages.info(request, f"Total Asset Count:{sum_count}")
             finally:
-                messages.info(request, "********************************************")
+                messages.info(request, "**********************************************")
 
     else:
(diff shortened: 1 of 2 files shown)


concordia/admin/views.py
```diff
@@ -225,13 +225,15 @@ def admin_bulk_import_review(request):
                 all_urls.append(urls)
                 for i, val in enumerate(all_urls):
                     return_result = fetch_all_urls(val)
-                    messages.info(request, f"Asset Count:{return_result[0]}")
+                    for res in return_result[0]:
+                        messages.info(request, f"{res}")
```
acdha, line 229: This will create one message per character in the returned string. I think this block should be `message, count = fetch_all_urls(val)`.

concordia/admin/views.py
```diff
@@ -225,13 +225,15 @@ def admin_bulk_import_review(request):
...
+
                     sum_count = sum_count + return_result[1]
                     time.sleep(10)
 
                 messages.info(request, f"Total Asset Count:{sum_count}")
             finally:
-                messages.info(request, "All processes Complete")
+                messages.info(request, "**********************************************")
```
acdha, line 236: This doesn't have any semantic meaning. I would stick with the old message but since the previous message has the same effect on a successful run I would vote for removing this entirely and making sure that the error path will emit a meaningful message.

importer/tasks.py
```diff
@@ -263,7 +263,7 @@ def import_item_count_from_url(import_url):
         resp.raise_for_status()
         item_data = resp.json()
         output = len(item_data["resources"][0]["files"])
-        return f"{import_url} {output}", output
+        return f"{import_url} - Asset Count: {output}", output
 
     except Exception as exc:
         return f"{exc}", 0
```
acdha, line 269: In addition to logging it, I would add something to prefix the exception since it's not always immediately obvious from the string value. Something like `f"Unhandled exception importing {import_url}: {exc}"`

acdha (changes requested):

concordia/admin/views.py
rabiloc: I want to display one message at a time from the thread workers when they return, so it shows the asset count per url, and the asset count and the total at the end.

concordia/admin/views.py
rabiloc: Yeah will remove that line, was doing a little formatting, but doesn't add much value.

importer/tasks.py
rabiloc: This is the message that the user sees - formatted.

acdha: Yes – the exception path should match that behaviour

Issue #1264, opened 2020-09-28
Filter in admin dashboard: Nesting Projects under Title Campaign


elainekamlley:
**User story/persona**
As an admin when I click on campaign I want to be see only the projects associated with the campaign so that I can filter by just those projects

**Is your feature request related to a problem? Please describe.**
As the program matures there are more campaigns/projects that need to be managed in the application. This has resulted in very long project lists, often having similar project names (ie Diaries or Miscellany) and is becoming harder to filter by the project we want.  

**This feature is applicable to the following pages in the dashboard:**
https://crowd.loc.gov/admin/concordia/asset/
https://crowd.loc.gov/admin/concordia/item/
https://crowd.loc.gov/admin/concordia/transcription/
https://crowd.loc.gov/admin/importer/importitemasset/
https://crowd.loc.gov/admin/importer/importitem/
https://crowd.loc.gov/admin/importer/importjob/

**Additional context**
Import jobs only have a long list of projects. There is no way to filter by campaign first and then projects
Here is the item page: 
![screencapture-crowd-loc-gov-admin-concordia-item-2020-11-24-09_50_24](https://user-images.githubusercontent.com/7362915/100111102-bcc66980-2e3b-11eb-9323-909ea06339d5.png)

**Acceptance Criteria**
GIVEN I am an admin/Community Manager
WHEN I filter by campaign 
THEN I can see a smaller list of all associate project to that campaign

**Acceptance Test:**
Add a list of steps to test for a user to check if functionality satisfies the acceptance criteria.

jkueloc:
@elainekamlley the filters are in dev and ready to be tested.

lalgee:
Looks great in dev @jkueloc! I checked all of the areas where the feature is applied. Thumbs up!

PR #1323, opened 2021-01-27
1264 filter in admin dashboard

jkueloc:
as mentioned in slack:
Good morning Chris - I'm stuck on change to Concordia I have 3 working admin customized SimpleListFilter classes that reduce the number of projects based on chosen champaign and project for item, asset, and transcription.  I'm trying to create a base class and just pass values to it based on the object (item, asset, or transcription).  The sticking point specifically is getting the base class queryset.filter to interpret the passed in foreign key field. I could not get **kwargs to work.  So the filter breaks when filtering on project (vs. campaign)
I'm working in the admin/filters.py for the base class -- had the 3 individuals filters working set up originally in  ..init...py
here's the base class - as it currently is

### DIFF
--- a/concordia/admin/__init__.py
+++ b/concordia/admin/__init__.py
@@ -54,6 +54,71 @@ from .forms import (
 )
 
 
+class CampaignProjectListFilter(admin.SimpleListFilter):
+    """
+    This filter is an example of how to combine two different Filters to work together.
+    """
+
+    # Title displayed on the list filter URL
+    title = "ProjectRedux"
+    # Model field name:
+    parameter_name = "project"
+    # Custom attributes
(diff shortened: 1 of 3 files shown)


jkueloc:
Changes were made locally after conferring with cadams via slack.

acdha:
I'd typically just leave this open and use it until the code is finalized

jkueloc:
Ah. Will do.
Thanks!

jkueloc:
The code for #1264 is now complete and ready for code review review and testing in dev by other team members once approved and merged.  Thank you!

rabiloc (approved):
Changes look good

PR #1372, opened 2021-03-31
1303 Updated styling of all detail pages

klee72:
- Updated all detail pages as described in the ticket
- Removed drop-shadow from cards to make it consistent with item-detail cards and in line with P1 styling
- Updated styling of 'Related links' box and 'View this item on www.loc.gov' button to meet contrast compliance
- Fixed overall spacings of the page, so that all elements/layouts on the page take advantage of the available width 
- Improved responsive design of the gallery
- Added 'zoom in' hover effect (thought it was strange for only 'completed' ones have a hover effect)
- Added spacing between gallery and pagination

### DIFF
--- a/concordia/static/scss/base.scss
+++ b/concordia/static/scss/base.scss
@@ -8,6 +8,7 @@ $orange: #fc4c02;
 $dark: #000;
 $accent: $orange;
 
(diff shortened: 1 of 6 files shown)


jkueloc (commented):
Looks good

PR #1408, opened 2021-04-30
1334 generate letter

rabiloc:


rabiloc:
Will wait until merges are fixed to re-pull

Issue #1478, opened 2021-06-10
Update dump and restore push-container scripts to add --pull 


jkueloc:
When container is rebuilt adding --pull will build the container with the latest base image (latest security updates and patches).

...
"... rebuild with --no-cache, so all layers are rebuilt, and with --pull so that you’re using the latest base image:"
```
$ docker build -t example.com/myorg/myimage --no-cache --pull .
$ docker push example.com/myorg/myimage
```

add update of postgresql9.6 to postgresql12 which is now available in amazon-linux-extras as of 20210707.

Cloudformation still needs updates for postgresql version and AWS Elastic Search (6.7 to 6.8)

jkueloc:
While I'm working on this on - I need to resolve why for the last 2 weeks beginning on 7/27/2021 the dev and test S3 sync/db restores have been receiving the following error - 
`copy failed: s3://crowd-content/early-copyright/copyright-title-pages-oversize-1790-1890/2019713462/583.jpg to s3://crowd-test-content/early-copyright/copyright-title-pages-oversize-1790-1890/2019713462/583.jpg **An error occurred (AccessDenied) when calling the GetObjectTagging operation: Access Denied**`
and then stopping shortly there after. Could be two things might be one.
There was a recent large copyright campaign load going on into Monday 7/27... not sure if that required something new.

PR #1511, opened 2021-08-12
1478 update dump restore containers

jkueloc:
This PR is related to issue #1478.  In addition to the Dockerfiles and push container scripts I updated the cloudformation/infrastructure yaml files impacted by the recent version upgrades of postgresql 9.6 to 12 (12.5) and elastic search 6.7 to 6.8.

One of the linters insisted that the rds.yaml database engine value be a string - '12.5' vs 12.5.
I'm hoping it's correct and I don't have a local linter configuration change to manage. But wanted to call it out just in case.

### DIFF
--- a/db_scripts/dump/Dockerfile
+++ b/db_scripts/dump/Dockerfile
@@ -1,14 +1,5 @@
 FROM git.loc.gov:4567/devops/docker-hub-mirror/amazonlinux:2
-SHELL ["bin/bash", "-c"]
-RUN echo $' \n\
-[pgdg12] \n\
-name=PostgreSQL 12 for RHEL/CentOS 7 - x86_64 \n\
-baseurl=https://download.postgresql.org/pub/repos/yum/12/redhat/rhel-7-x86_64 \n\
-enabled=1 \n\
-gpgcheck=0' >> /etc/yum.repos.d/pgdg.repo
-
-RUN yum makecache && yum -y install postgresql12 \
-    && yum update -y \
+RUN yum update -y && amazon-linux-extras install -y postgresql12 \
     && yum -y install unzip \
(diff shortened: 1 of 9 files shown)


db_scripts/dump/Dockerfile
```diff
@@ -1,14 +1,5 @@
...
-RUN echo $' \n\
-[pgdg12] \n\
-name=PostgreSQL 12 for RHEL/CentOS 7 - x86_64 \n\
-baseurl=https://download.postgresql.org/pub/repos/yum/12/redhat/rhel-7-x86_64 \n\
-enabled=1 \n\
-gpgcheck=0' >> /etc/yum.repos.d/pgdg.repo
-
-RUN yum makecache && yum -y install postgresql12 \
-    && yum update -y \
+RUN yum update -y && amazon-linux-extras install -y postgresql12 \
```
acdha, line 2: Good call here

jkueloc: Thank you!
AWS was slow to add postgres12 to amzon-linux-extras so it was hacked for a bit while waiting for them to catch up. I neglected to check into that before the db upgrade... kinda like not registering the gravity vaccum ahead of time.

acdha: I hadn't seen that myself – left me thinking it'd be nice to have something like a GitHub bot or pre-commit hook which was basically a general-purpose version of the [eslint-plugin-unicorn expiring TODO comments](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/expiring-todo-comments.md) where you could leave a comment like `# TODO [2021-09-01]: switch to extras package` which would start triggering a warning when the expiration date is past.

Issue #1508, opened 2021-07-20
500 error trying to export George Washington BagIt


lalgee:
**What behavior did you observe? Please describe the bug**
When trying to do a BagIt Export for the Ordinary Lives in George Washington's Papers in production the Bag does not export and instead a 500 error loads.

**How can we reproduce the bug?**

1. Go to [George Washington](https://crowd.loc.gov/admin/concordia/campaign/21/change/) campaign in Django admin
2. Click Export Bagit
3. See 500 error

**What is the expected behavior?**
When the BagIt Export button is clicked, the user should remain on the page and a .zip BagIt file should download.

**Additional context**
May be a similar issue to  https://github.com/LibraryOfCongress/concordia/issues/1501

acdha:
This is related to how it's looking up the loc.gov image identifier:

https://github.com/LibraryOfCongress/concordia/blob/1c2d8de56352ce0f5f24a9722ad1f959363ce4a6/exporter/views.py#L49-L71

The pattern probably needs to be updated to handle the new storage-services URLs which were added to loc.gov after that code was written:

```python

download_url | 'http://tile.loc.gov/storage-services/master/mss/mgw/mgw6a/117/1400/1498.jpg'
-- | --
pattern | '/service:([A-Za-z0-9:.\\-\\_]+)/'

```

jkueloc:
pattern = r"/service:([A-Za-z0-9:.\-\_]+)|/master/([A-Za-z0-9:.\-\_\/]+)/"

jkueloc:
@lalgee This is ready for testing.  I was able to export BagIt for both campaigns in a search for "George" and Walt with the new code locally and in crowd-dev post merge request. 
Please give it shot and let me know if you run into the 500 error again.

jkueloc:
@elainekamlley @lalgee the last update that corrects the missing missing subdir/file for the expected bagit zip is currently being held back as a result of errors in the Jenkins build for step 6 that I've ticketed as - Investigate Jenkins Server build errors preventing the concordia image build #1518

laurenseroka:
Tested in dev and everything looks as expected

PR #1512, opened 2021-08-12
1508 export handle new storage services

jkueloc:
This PR is for issue #1508.


jkueloc:
apologies the third commit comment was not useful. Third commit was to return the asset_id.group(2) if the new storage-services URL was matched.

Issue #1630, opened 2022-03-09
Add Asset Id to Admin Campaign Export-CSV 


jkueloc:
Tied to #1611 

Add Asset Id to the Admin Campaign Export-CSV (button in the change panel) to allow for mapping contents back to the Admin Transcriptions Export to CSV - available via the Admin Transcription list Action. 

Currently, Admin Campaign Export-CSV identifies an Asset by Title (relevant outside of Concordia) but not the Asset Id (internal to Concordia). 

The Transcriptions admin action Export to CSV contains the asset Asset Id (internal to Concordia)

The purpose of this enhancement is to be able to perform analytics (post retirement or otherwise) on the number of Transcriptions to an Asset within a Campaign.

lalgee:
Looks great!  No issues testing in dev - was able to look for and map data between the campaign CSV export and exported transcriptions via the asset id

Issue #1611, opened 2022-02-17
Campaign Retiring - Epic


jkueloc:
Campaign Sunsetting - Epic

Supporting document: https://staff.loc.gov/wikis/x/ve_hBQ

- [ ] Define, design and build Report for capturing the sunset campaign's metrics #1623 
- [ ] New model/database representation - to capture metrics of retired campaigns for dashboard representation. At first initial thought, this will be an action within a Campaign Module
       - for reference see existing model graph here #1530
- [ ] New model/database representation - to capture metrics of retired campaigns tied to the user profile contribution page. Tag tied to user-asset-tag-collection (-tag), transcription tied to asset, etc.
- [ ] Modify user profile contribution page for sunset campaign contributions

jkueloc:
This Epic moved to Jira [CONCD-9](https://staff.loc.gov/tasks/browse/CONCD-9)

PR #1631, opened 2022-03-09
add asset id to campaign csv output

jkueloc:
#1630 and #1611


jkueloc:
@lalgee This is ready for your review in development.

PR #1694, opened 2022-06-10
display tag count in Tags area header

rasarkar:


jkueloc (approved):
This PR supports [CONCD-15](https://staff.loc.gov/tasks/browse/CONCD-15)

PR #1713, opened 2022-07-25
Registered users should be able to delete others' tags

rasarkar:
https://staff.loc.gov/tasks/browse/CONCD-16


concordia/views.py
```diff
@@ -1404,7 +1404,9 @@ def submit_tags(request, *, asset_pk):
 
     for tag in existing_user_tags:
         if tag not in all_submitted_tags:
-            user_tags.tags.remove(tag)
+            for collection in asset.userassettagcollection_set.all():
+                if tag in collection.tags.all():
```
acdha, line 1408: Couldn't this be simplified to `collection.tags.remove(tag)` to avoid the extra queries needed to load the entire tag set? If you computed the tags as a set difference (i.e. something like `set(all_submitted_tags).difference(existing_user_tags)`) you could then make a single call `collection.tags.remove(*all_removed_tags)`.

Similarly, rather than iterating over every collection individually you should be able to do this as a single query by hitting `UserAssetTagCollection` directly — something like `UserAssetTagCollection.tags.through.objects.filter(tag__in=removed_tags, userassettagcollection__asset=asset).delete()` but I'd need to test that code.

concordia/tests/test_views.py
```diff
@@ -1088,6 +1088,59 @@ def test_duplicate_tag_submission(self):
...
+    def test_tag_deletion(self):
+        asset = create_asset()
+        self.login_user()
+
+        test_tags = ["foo", "bar"]
+        resp = self.client.post(
+            reverse("submit-tags", kwargs={"asset_pk": asset.pk}),
+            data={"tags": test_tags},
+        )
+        test_tags = [
```
acdha, line 1100: I would use a different variable name to make it clear that you are intentionally removing only one of the values — “test_tags” is a very generic name so I'd probably go with something like “initial_tags” and “updated_tags” to make the subsequent assertions more explanatory.

concordia/tests/test_views.py
```diff
@@ -1088,6 +1088,59 @@ def test_duplicate_tag_submission(self):
...
+        ]
+        resp = self.client.post(
+            reverse("submit-tags", kwargs={"asset_pk": asset.pk}),
+            data={"tags": test_tags},
+        )
+        data = self.assertValidJSON(resp, expected_status=200)
+        self.assertIn("user_tags", data)
+        self.assertIn("all_tags", data)
+
+        self.assertEqual(sorted(test_tags), data["user_tags"])
```
acdha, line 1111: Rather than sorting the list, you can use [`assertCountEqual`](https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertCountEqual):

```suggestion
        self.assertCountEqual(test_tags, data["user_tags"])
```

concordia/tests/test_views.py
```diff
@@ -1088,6 +1088,59 @@ def test_duplicate_tag_submission(self):
...
+        self.assertEqual(sorted(test_tags), data["all_tags"])
```
acdha, line 1112:
```suggestion
        self.assertCountEqual(test_tags, data["all_tags"])
```

concordia/tests/test_views.py
```diff
@@ -1088,6 +1088,59 @@ def test_duplicate_tag_submission(self):
...
+        )
+        data = self.assertValidJSON(resp, expected_status=200)
+        self.assertIn("user_tags", data)
+        self.assertIn("all_tags", data)
+        self.assertEqual(sorted(test_tags), data["user_tags"])
+        self.assertEqual(sorted(test_tags), data["all_tags"])
+
+        self.client.logout()
+
+        self.login_user("tester")
```
acdha, line 1130: Isn't this the default name, meaning that both of these tests are running as `tester`? I would make this clear in the test suite by using different usernames explicitly in both calls — `self.login_user("first_user")`, `self.login_user("second_user")`, etc. — so there's no chance of a test appearing to work by accident. That could be enforced by doing a check like ensuring that `asset.userassettagcollection_set.user_set.count()` is 2.

concordia/views.py
rasarkar: Great suggestions, thank you. I don't think the "tag__in" filter will work (since the field name is "tags", not "tag"), but I should be able to update the branch with the other code changes.

jkueloc (approved):

PR #1721, opened 2022-08-01
fixing issues in the nav menu

rasarkar:
linebreaks were accidentally added to some of the templatetags in base.html, causing the nav menu items to link and/ or display incorrectlly


rasarkar:
All of the django template blocks ({% %}, {{ }}) should work correctly, now

jkueloc (approved):

PR #1723, opened 2022-08-03
CONCD-14 fix help page link contrast post robo format

jkueloc:
CONCD-14
Note: lots of robo formatting changes by DjCSS. 
Made one change for Prettier and combined elements of a duplicated selector.
pre-commit seemed to get stuck after those changes were git added when going to commit. 
Went with --no-verify after three attemtps to commit but the build locally (npx gulp build) looked good for the change and the robo reformatting didn't seem to break anything.

```
stylelint................................................................Failed
- hook id: stylelint
- exit code: 2
concordia/static/scss/base.scss
 436:1  :heavy_multiplication_x:  Unexpected duplicate selector                 no-duplicate-selectors
           ".concordia-object-card .card-img", first
           used at line 401
```

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -13,8 +13,6 @@ repos:
       rev: 'v1.5.1'
       hooks:
           - id: djhtml
-          - id: djcss
-          - id: djjs
     - repo: https://github.com/ambv/black
       rev: 22.3.0
       hooks:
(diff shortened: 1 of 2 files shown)


acdha:
I would fix the underlying problem or disable that rule — using `--no-verify` means it can now no longer work in CI or for anyone else until that's resolved.

PR #1753, opened 2022-09-20
CONCD-168 link to service letter

rasarkar:
Update link text
https://staff.loc.gov/tasks/browse/CONCD-168


joshuastegmaier (changes requested):
Based on the ticket, the link should be to /resources rather than /contact

rasarkar:
> Based on the ticket, the link should be to /resources rather than /contact

Oops. Ty for catching that! I've updated the link.

joshuastegmaier (approved):

PR #1782, opened 2022-11-03
Concd 184 jkue feature branch update

jkueloc:
CONCD-184 - During testing for AWS role standardization ec2-user was unable to install amazon-linux-extras postgresql12 in order to pg_restore prod database dmp to feature branch concordia database. sudo was added to the command.


jkueloc:
GitHub doesn't seem to keep the resolution of conflict so I do it again after review.  I've tried twice already.

joshuastegmaier (approved):

PR #1845, opened 2023-01-18
Added status to campaign admin list view and removed unused fields

joshuastegmaier:
https://staff.loc.gov/tasks/browse/CONCD-244


rasarkar:
Jira ticket mentions "campaign date". Should that be added, or is it no longer necessary?

joshuastegmaier:
Good catch

PR #1849, opened 2023-01-20
CONCD-230 bump uri component

jkueloc:
Bump decode-uri-component.  Includes https://github.com/LibraryOfCongress/concordia/pull/1813

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -13867,9 +13867,9 @@
             }
         },
         "decode-uri-component": {
-            "version": "0.2.0",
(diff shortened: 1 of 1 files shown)


joshuastegmaier (commented):
This PR doesn't seem to actually do anything--there are no changes. Looking at your last commit, it seems like it undid the change the bot made and reverted back to uri 0.2.0. Is that expected?

jkueloc:
Thank you for catching that!

jkueloc:
Going to use the original dependabot PR instead.

PR #1861, opened 2023-02-01
Added ability to overwrite existing resource files

joshuastegmaier:
https://staff.loc.gov/tasks/browse/CONCD-207


joshuastegmaier:
The failing Lint checks are because of a new version of Black and can be ignored.

rasarkar:
The call to super.save() is throwing a ClientError, when I try to test this in my local env. Should I just ignore that?
https://github.com/LibraryOfCongress/concordia/blob/CONCD-207-jstegmaier-resource-file-management/concordia/models.py#L247

joshuastegmaier:
> The call to super.save() is throwing a ClientError, when I try to test this in my local env. Should I just ignore that? https://github.com/LibraryOfCongress/concordia/blob/CONCD-207-jstegmaier-resource-file-management/concordia/models.py#L247

That's caused by your AWS token be expired. If you refresh it (idaptive-aws-cli-login --renew-all-sessions), the error should go away.

rasarkar (approved):

PR #1870, opened 2023-02-08
Removed and changed various help center and welcome guide links

joshuastegmaier:
https://staff.loc.gov/tasks/browse/CONCD-269

These changes won't work locally without adding new Simple Pages for the new URLs (e.g., /welcome-guide/, /welcome-guide/how-to-tag/, etc.) I'm adding those pages in the dev, test and production environments so the changes won't break there.

### DIFF
--- a/concordia/templates/base.html
+++ b/concordia/templates/base.html
@@ -213,7 +213,7 @@ because it's not worth dealing with IE11 compatibility {% endcomment %}
                                 <li class="list-inline-item mb-1"><a href="/contact/">Contact
                                     Us</a></li>
                                 <li class="list-inline-item"><a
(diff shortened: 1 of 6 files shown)


rasarkar:
Did we want to create a new page for /welcome-guide/? The ticket seemed to indicate that we should use the existing /help-center/welcome-guide/ page.

joshuastegmaier:
The way Simple Pages create breadcrumbs requires the URLs to change in order to change the breadcrumbs /help-center/welcome-guide/ will always include the breadcrumb for the 'help-center' page, which is what they don't want.

rasarkar (approved):

PR #1880, opened 2023-02-17
CONCD-223 label changes, etc.

rasarkar:
- Change "User:" label to "User Id:", "Campaign: to "Campaign Id:"
- Use UniqueConstraint with the constraints option instead.
- Change model display name (to "User completed campaign counts")


joshuastegmaier:
Just want to double check--the changes in this request don't fully match the description. The only thing that's changed is the labels, but not the constraints or model display name. Is that expected?

rasarkar:
> Just want to double check--the changes in this request don't fully match the description. The only thing that's ...

Good question. There should be a change to the class Meta, as well as a migration. Let me try to figure out what's going on

rasarkar:
No idea what's going on here, but for some reason git refuses to acknowledge the changes to the model.

joshuastegmaier (approved):

PR #1940, opened 2023-04-05
CONCD-210 Admin ability to change status to Completed

rasarkar:


joshuastegmaier:
The action needs to have the admin.action decorator like the other actions, to define the permission required and give a description (the description isn't technically required, but I think we should do that as a standard.)

The permission for now can be the 'reopen' permission that reopen_asset_action uses, though check with the CMs to see if they want a different permission for this (and the other new actions) instead.

I don't think the exclusion for the user's own transcription is necessary at https://github.com/LibraryOfCongress/concordia/blob/7661db2775695f4b7bf04a8b80b200bbe16c0cbe/concordia/admin/actions.py#L109. This is administration action, so it should be possible for a staff member to, e.g., correct a mistake in a transcription then accept it--but that's also something to double check with the CMs.

rasarkar:
I'll make the change fo the decorator, but the last part in your comment is incorrect - the system will **definitely** throw an error if a user attempts to accept their own transcription, even if it's through the admin. Whether or not that's the desired behavior, I don't know for sure.

joshuastegmaier (approved):

PR #2038, opened 2023-07-20
Replaced part of OCR pop up text

joshuastegmaier:
https://staff.loc.gov/tasks/browse/CONCD-468


rasarkar (approved):

joshuastegmaier:
@jkueloc This is a small text change we're hoping to merge into the release branch before the release goes live, assuming that doesn't cause any problems with the release process.

PR #2227, opened 2024-01-10
Initial implementation of maintenance mode

joshuastegmaier:
Updated pre-commit-config to allow 256k files (needed for the maintenance error page image.)

Rolled back stylelint packages so they actually work. Will make a ticket to figure out what's broken in the newer version.

https://staff.loc.gov/tasks/browse/CONCD-648 for the maintenance mode in general
https://staff.loc.gov/tasks/browse/CONCD-649 for the design of the error page

Maintenance mode is turned on and off in the admin (only accessible by super users):
chrome_pwi6DPgOhh

Staff users will still be able to access the site while in maintenance mode, with all front-end pages displaying a banner:
chrome_T891HD6ct3

All non-staff users will receive the 503 error page while maintenance mode is activated.

The admin is still accessible while maintenance mode is active--if you're not logged in, you'll still get a 503 error on the front-end, but you can go to the admin URL to login, after which the frontend will also be accessible to you.

Maintenance mode uses Redis, so if that's down, maintenance mode can't be activated through the admin (it will default to no maintenance mode if it can't contact Redis.) We can override that to automatically go into maintenance mode when it can't contact Redis (or push a release that does that in the case of emergency), but I thought defaulting to the site not going down whenever there's a hiccough with Redis was the better default.

The updates to stylelint and related packages a while ago caused it to break. This wasn't revealed until this commit because no stylesheets have been updated since then. I rolled the updates back to the last known good versions of all packages.


jkueloc:
@joshuastegmaier image compression, please see...
[When you update any of the files under `concordia/static/img`, please use an](https://github.com/LibraryOfCongress/concordia/blob/86d7d7743e07a1efdccf1b282f0c89a9098d6125/docs/for-developers.md?plain=1#L266)

joshuastegmaier:
> @joshuastegmaier image compression, please see... [When you update any of the files under `concordia/static/img`, please use an](https://github.com/LibraryOfCongress/concordia/blob/86d7d7743e07a1efdccf1b282f0c89a9098d6125/docs/for-developers.md?plain=1#L266)

We should update this to not be a Mac-only solution. In any case, attempting to optimize this image just resulted in the exact same image size. I'm guessing the Library (where the image came from) already does optimizations to shrink the images.

jkueloc:
> > @joshuastegmaier image compression, please see... [When you update any of the files under `concordia/static/img`, please use an](https://github.com/LibraryOfCongress/concordia/blob/86d7d7743e07a1efdccf1b282f0c89a9098d6125/docs/for-developers.md?plain=1#L266)
> 
> We should update this to not be a Mac-only solution. In any case, attempting to optimize this image just resulted in the exact same image size. I'm guessing the Library (where the image came from) already does optimizations to shrink the images.

Thank you for double checking! I'll create a ticket for the image compression/doc update.

jkueloc (approved):

PR #2230, opened 2024-01-11
Revert "Bump follow-redirects from 1.15.3 to 1.15.4"

jkueloc:
Reverts LibraryOfCongress/concordia#2228

app container healthz check and metrics are failing - ECS can't reach a steady state.
Need to troubleshoot that first.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4303,9 +4303,9 @@
             }
         },
         "node_modules/follow-redirects": {
-            "version": "1.15.4",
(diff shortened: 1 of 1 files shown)


jkueloc:
dev was in maintenance mode which returns the 503 to healthz check which has to clear 3 times in a time period before the ECS task will keep running - vs. containly stopping and relaunching because it didn't get a the correct # of healthz checks.

PR #2254, opened 2024-01-30
CONCD-674 display heading should not be larger than the title

rasarkar:
https://staff.loc.gov/tasks/browse/CONCD-674


joshuastegmaier:
Github says there's a conflict. I don't think there's a point to me approving it until that's fixed, since I'd probably need to approve it again afterwards.

joshuastegmaier (approved):

PR #2276, opened 2024-02-13
CONCD-654 Merge Feature branch into Main

rasarkar:


concordia/settings_template.py
joshuastegmaier, line 1: These lines:
MAINTENANCE_MODE_IGNORE_STAFF = True
MAINTENANCE_MODE_IGNORE_SUPERUSER = True

Should be removed. They were originally in the settings file, but that setting is now controlled through the admin, so they should not be set in settings.

package.json
joshuastegmaier, line 1: All of the prettier and stylelint libraries should be removed from package.json (prettier is also in the regular dependencies, where it should remain, since it is used by the site.)

joshuastegmaier (approved):

PR #2307, opened 2024-03-06
CONCD-712 workaround 

rasarkar:
eval queryset before it can be passed to an async context


jkueloc:
@rasarkar @joshuastegmaier 
I'll merge this but I don't think it addresses all that's broken. There is no workaround for export_to_csv_action in __init__.py which is also broken.  export_to_csv_response covered the tags and Campaign - but not the actions for transcriptions, assets, etc..

PR #2334, opened 2024-04-09
CONCD-728 Create admin ability to turn OCR off for specific assets

rasarkar:
https://staff.loc.gov/tasks/browse/CONCD-728


joshuastegmaier:
I think the help text for the fields should be changed. All of them except Asset have the same typo (the first "this" should be "the", I think.) I'd also not use "descendent" or "children" and just say "assets" for clarity ("Turn off OCR for all assets of this campaign", etc.) It's annoying to do another migration just for that, but the text will be displayed in the admin, so I think it's worth the change.

joshuastegmaier (dismissed):

joshuastegmaier (approved):

PR #2347, opened 2024-04-18
CONCD-599 unit test for the Asset.get_ocr_transcript() method

rasarkar:


joshuastegmaier:
This unit test doesn't actually seem to test any of our code (it just tests if tesseract works), so I don't think it should be added. It should be changed to use Asset.get_ocr_transcript.

joshuastegmaier (approved):

PR #2352, opened 2024-04-23
Upgrade to gulp 5

joshuastegmaier:
https://staff.loc.gov/tasks/browse/CONCD-778


rasarkar (approved):

jkueloc:
I don't see where we are using everything the pip-audit fine is reporting on so moving this forward.

PR #2408, opened 2024-06-12
Updated Dockerfile to install non-English tesseract language packs (main)

joshuastegmaier:
https://staff.loc.gov/tasks/browse/CONCD-845

Opening this pull request now for a release hotfix now, but it shouldn't be merged until we've decided how to precede


joshuastegmaier:
We can proceed with merging this into main. Even if we want to do it separately as a hotfix, it would need to be merged in here as well.

PR #2410, opened 2024-06-12
Reverted changes to next transcribable/reviewable

joshuastegmaier:


rasarkar:
The "jump into a transcription" link seems to work fine, but the "jump into a review" link keeps redirecting me to the homepage. I'll try to figure out whether or not that's being caused by something in my environment

rasarkar (approved):

PR #2479, opened 2024-08-02
Moved all javascript scripts on the asset detail page  to their own files.

joshuastegmaier:
http://staff.loc.gov/tasks/browse/CONCD-896

This is the first step in the refactoring. All JavaScript scripts are moved out into their own files. The JS isn't in its final state, since everything is just dumped into the global context, but this gets them out of asset_details.html, reducing the lines in the file by about a third.


rasarkar:
So neither of the links work, haha (they just link back to Teams, not to GitHub or Jira)

rasarkar (approved):

PR #2481, opened 2024-08-05
Split asset_detail.html template into smaller templates. Re-added a JS function that was needed.

joshuastegmaier:
https://staff.loc.gov/tasks/browse/CONCD-896


rasarkar (approved):
Somewhat related: we should probably rename quick_tips_modal to campaign_tips_modal.

joshuastegmaier:
I actually decided against doing that since quick tips is used elsewhere. We can continue using that name in the code for the feature.

jkueloc:
@joshuastegmaier This is still marked as a draft...?

joshuastegmaier:
Just didn't hit the button when the checks all finished. Sorry!

PR #2484, opened 2024-08-09
revert bad merge to release

jkueloc:
https://staff.loc.gov/tasks/browse/CONCD-910


joshuastegmaier (approved):

jkueloc:
I will likely need to reapply the previous two hotfixes for the 8.1.0 "Release" to get the release branch state where it should be before the bad merge. Since both of those fixes were merged separately into the release and mains branches, it appears git is set to reverse it via the merge to main which we are now reversing.

PR #2531, opened 2024-09-19
Added bypass for retrieving the turnstile secrets.

joshuastegmaier:
Also a couple of changes I meant to make earlier: 
 Moved turnstile context processor to its own file. 
 Removed redundant/unused context processor

The try/except/pass block is a possible security issue because it can hide errors caused by an attack, but in this case I just bypassed the Bandit test for it since it is a temporary measure (and I believe with this usage it wouldn't actually be a potential security risk anyway). The fix would be catch the specific exceptions being raised, but those are just normal Exceptions anyway, so that would require rewriting the secrets code to raise different exceptions.

### DIFF
--- a/concordia/context_processors.py
+++ b/concordia/context_processors.py
@@ -32,32 +32,3 @@ def site_navigation(request):
 def maintenance_mode_frontend_available(request):
     value = cache.get("maintenance_mode_frontend_available", False)
     return {"maintenance_mode_frontend_available": value}
-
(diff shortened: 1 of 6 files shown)


rasarkar:
I'll double-check, but log in does not appear to be working.

joshuastegmaier:
We don't need this anymore since the secrets permissions have been updated.

PR #2543, opened 2024-09-25
Added resetting to the turnstile widget after form submission

joshuastegmaier:
Also changed turnstile to only display a widget when interaction is required. Moved the widget to below the form buttons so it has room to appear but doesn't change the spacing on the form when not visible.

Added caching for validation of anonymous users. Users should only be tested against Turnstile once per day (so long as they keep the same session), which is the same interval as used with captcha.

https://staff.loc.gov/tasks/browse/CONCD-952


joshuastegmaier:
I went ahead and added the cached validation for anonymous users to the original pull request.

PR #2549, opened 2024-10-01
Create codeql.yml

jkueloc:
Codeql started running two separate workflows beginning on Sept 23 for reasons unknown.  a second workflow instance of CodeQL showed up in the workflows giving error, " Code Scanning could not process the submitted SARIF file: CodeQL analyses from advanced configurations cannot be processed when the default setup is enabled". Attempts to turn off CodeqL or set to advance forced the creation of this workflow.

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -1,7 +1,6 @@
 name: 'CodeQL Advanced'
 
 on:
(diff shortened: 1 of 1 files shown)


acdha:
This seems to be running without errors. I'd merge if it wasn't a draft.

PR #2585, opened 2024-11-07
Fixes for broken image messages

joshuastegmaier:
Fixed bug with displayHTMLMesage when receiving a Text node.
Changed contact link to be capitalized, use correct URL and open in a new tab.

https://staff.loc.gov/tasks/browse/COND-972


joshuastegmaier:
The CodeQL warning is not relevant because we're not actually pulling text out of the DOM. displayMessage just puts its text (which we create/control) in a Text node, so there's no possibility of XSS.

jkueloc (approved):

PR #2604, opened 2024-11-21
Remaing concordia.views tests

joshuastegmaier:
https://staff.loc.gov/tasks/browse/CONCD-1003


joshuastegmaier:
pip-audit failures are unrelated to any of these changes. I will make a ticket to do those updates, since it looks like we'll need to jump forward at least two major versions for the redis update and at least one for pillow, so I'll want to do more testing than just throwing it into this changeset.

rasarkar (approved):

PR #2608, opened 2024-11-29
Added test for update_task_status decorator

joshuastegmaier:
Also:
 * Modified update_task_status to call the passed object an object rather than a model
 * Modified ImportJob.__str__ to properly handle when created_by is None

https://staff.loc.gov/tasks/browse/CONCD-852


joshuastegmaier:
CodeQL failure is the common firefox snap timeout failure, not related to this change.

jkueloc (approved):

PR #2729, opened 2025-03-10
Modified rate limit message migration to use new pop-up message

joshuastegmaier:
https://staff.loc.gov/tasks/browse/CONCD-1114

This won't change the data in the database if this migration has already been applied (though you can rollback to 0001 configuration migration and then reapply 0002 and 0003), but ensures the correct values are in place after database restores in dev/test and when these changes go to production.


joshuastegmaier:
Incorrect change to banner message

rasarkar (approved):

PR #2902, opened 2025-06-17
Visualization accessibility improvements

joshuastegmaier:
https://staff.loc.gov/tasks/browse/CONCD-1268

Added a non-visible table for each visualization to provide an alternate view for, e.g., screen readers.
Added the chroma-js library to generate color palettes that maintain contrast ratios
Added keyboard interactions to visualizations
Explicitly set the initial aspect ratio of the chart, and disabled chart.js's maintainAspectRatio in order to allow it to better zoom past 200%.
Updated chart.js to the latest version.

### DIFF
--- a/concordia/settings_template.py
+++ b/concordia/settings_template.py
@@ -80,6 +80,7 @@ NPM_FILE_PATTERNS = {
     ],
     "chart.js": ["auto/*", "dist/*"],
     "@kurkle/color": ["dist/*"],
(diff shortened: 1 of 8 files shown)


rasarkar:
I'm seeing a bunch of console errors: it looks like the import map is missing a comma on line 14.

rasarkar (approved):

PR #3381, opened 2026-08-11
First draft to address save and reservation errors post cf change Aug 7.

jkueloc:
CONCD-1619

### DIFF
--- a/concordia/middleware.py
+++ b/concordia/middleware.py
@@ -1,3 +1,5 @@
+from django.core import signing
+from django.utils.deprecation import MiddlewareMixin
 from maintenance_mode.http import get_maintenance_response
 from maintenance_mode.middleware import (
     MaintenanceModeMiddleware as BaseMaintenanceModeMiddleware,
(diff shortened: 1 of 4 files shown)


concordia/middleware.py
```diff
@@ -11,3 +13,58 @@ def process_request(self, request):
...
+
+
+class CloudflareAuthStatusMiddleware(MiddlewareMixin):
+    """
+    Sets an HMAC-signed cookie (_cf_acc_status) for authenticated, non-anonymous users.
+    Cloudflare Edge WAF rules evaluate this cookie to bypass challenges for logged-in
+    users.
+    """
+
+    COOKIE_NAME = "_cf_acc_status"
```
acdha: One challenge: anyone can set a cookie name, including bot operators. If we called `get_signed_cookie` to validate that will block request processing early. Since our goal is to avoid expensive database operations, the lower overhead of checking a signature is probably fine.

We might also want to make the cookie name a configuration item (periodically rotating?) so someone wanting to attack it at least has to create an account and leave more signs of their activity.

concordia/middleware.py
```diff
@@ -11,3 +13,58 @@ def process_request(self, request):
...
+                        salt=self.SIGNING_SALT,
+                        max_age=86400 * 7,
+                    )
+                    if data.get("uid") == user.pk and data.get("auth") is True:
+                        needs_cookie = False
+                except (signing.BadSignature, signing.SignatureExpired):
+                    needs_cookie = True
+
+            if needs_cookie:
+                response.set_cookie(
```
acdha: Why not use `HttpResponse.set_signed_cookie`?

jkueloc: Thank you! Will change that as well.

concordia/middleware.py
jkueloc: Good point on rotating.  I'll add it to the other environment secrets - just need to not forget the syncing in Cloudflare when rotated.

concordia/middleware.py
```diff
@@ -11,3 +14,66 @@ def process_request(self, request):
...
+class CloudflareAuthStatusMiddleware(MiddlewareMixin):
+    """
+    Sets and validates an HMAC-signed cookie for authenticated, non-anonymous users.
+    Allows Cloudflare Edge rules to bypass challenges for logged-in users while
+    validating HMAC signatures early in Django before expensive DB operations occur.
+    """
+
+    SIGNING_SALT = "concordia.cloudflare.auth_status"
+
+    @property
```
acdha, line 36: Tempted to wrap these with `@functools.cached_property` even though it's probably not too much faster than the settings object — this is going to run on every request so that might matter more than normal.

acdha (commented):
In general the main question I have left is why we're not using `get_signed_cookie` and `set_signed_cookie`.

concordia/middleware.py
jkueloc: I understand the inclination. It would maybe only add ~0.1 microseconds per request...?  
We would need to force new deployment on rotation if we switched @functools.cached_property.

concordia/middleware.py
jkueloc: My bad - The why not get_signed_cookie and set_signed_cookie was initially for payload format consistency and edge compatibility but has now been incorporated.

concordia/middleware.py
acdha: Don't we need to do that anyway? `settings_ecs.py` appears to only load it once at startup.

concordia/middleware.py
```diff
@@ -1,13 +1,111 @@
...
+        current_cookie_name = self.cookie_name
+        cookie_present = current_cookie_name in request.COOKIES
+
+        if is_authenticated:
+            valid_signature = False
+
+            if cookie_present:
+                try:
+                    # Low-overhead HMAC signature check using get_signed_cookie
+                    raw_val = request.get_signed_cookie(
```
acdha, line 82: Set `default=None` so you can gracefully handle cases where it's not present.

concordia/middleware.py
```diff
@@ -1,13 +1,111 @@
...
+                        event_code="cloudflare_auth_cookie_corrupt_payload",
+                        reason="Failed to parse cookie payload.",
+                        reason_code="invalid_json",
+                        user=user,
+                    )
+                    valid_signature = False
+
+            # Re-issue or set fresh signed cookie if invalid, missing, or expired
+            if not valid_signature:
+                payload = json.dumps({"uid": user.pk, "auth": True})
```
acdha: Are we using `auth` anywhere? It doesn't appear so.

concordia/middleware.py
```diff
@@ -1,13 +1,111 @@
...
+                response.set_signed_cookie(
+                    current_cookie_name,
+                    value=payload,
+                    salt=self.SIGNING_SALT,
+                    max_age=self.cookie_age,
+                    httponly=True,
+                    secure=not request.META.get("DEVELOPMENT", False),
+                    samesite="Lax",
+                )
+        else:
```
acdha: `elif cookie_present`:

acdha (commented):
I think this is basically ready to test

